PR #241, opened 2015-02-03
Exceptions in events are swallowed

tbee:


tbee:
This commit only catches what is required by the signature, and wraps that in a new exception. All other exceptions are simply passed through.

PR #245, opened 2015-02-07
Adds support to multiple methods as event listeners

### LINKED ISSUE #244: Support more than one method annotated as event listeners
Commit a49cd42a0c03be304387d6fc4b7910d090d9a1c8 adds support to `@PrePersist`, `@PreUpdate` and other event annotations. But looks like it supports only a single method per event type, while the docs suggests that more than one method could be annotated:

>  Is used to specify callback methods for the corresponding lifecycle event. This annotation may be applied to methods of an entity class, a mapped superclass, or a callback listener class. - http://docs.oracle.com/javaee/6/api/javax/persistence/PrePersist.html

megazord:
Basically, it supports more than one method annotated with `@PrePersist`, `@PreUpdate` and so on.

Fixes #244

### DIFF
--- a/src/main/java/com/avaje/ebeaninternal/server/deploy/BeanLifecycleAdapterFactory.java
+++ b/src/main/java/com/avaje/ebeaninternal/server/deploy/BeanLifecycleAdapterFactory.java
@@ -2,6 +2,8 @@ package com.avaje.ebeaninternal.server.deploy;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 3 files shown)


megazord:
Thank you very much, @rbygrave. Do you have an idea about when version 4.5.5 will be released?

Again, thank you for accepting the PR.

rbygrave:
I'll look to release 4.5.5 tonight.

Thanks for the pull request.


Cheers, Rob.

megazord:
Nice! Thank you.

Issue #259, opened 2015-03-19
Problem with RawSql: query does not contain any named bind parameters?


ForNeVeR:
Is there some problem with RawSql caching? I'm using the recommended `4.5.4` version of ebean (I had the same problem with older `2.8.1` though), and the following code is throwing an exception:

```java
	public static void main(String[] args) {
		System.out.println(call());
		System.out.println(call()); // <- exception from this invocation
	}

	private static int call() {
		String sql = "select * from my_table where id = :id and name = :name";
		RawSql rawSql = RawSqlBuilder.unparsed(sql)
				.columnMapping("id", "id")
				.columnMapping("name", "name")
				.create();
		Query<MyTable> query = Ebean.find(MyTable.class);
		query.setParameter("id", "1");
		query.setParameter("name", "name");
		query.setRawSql(rawSql);
		return query.findList().size();
	}
```

It throws the following exception:

```
Exception in thread "main" javax.persistence.PersistenceException: query does not contain any named bind parameters?
    at com.avaje.ebeaninternal.server.util.BindParamsParser.parseNamedParams(BindParamsParser.java:119)
    at com.avaje.ebeaninternal.server.util.BindParamsParser.parseNamedParams(BindParamsParser.java:113)
    at com.avaje.ebeaninternal.server.util.BindParamsParser.parseSql(BindParamsParser.java:99)
    at com.avaje.ebeaninternal.server.util.BindParamsParser.parse(BindParamsParser.java:47)
    at com.avaje.ebeaninternal.server.util.BindParamsParser.parse(BindParamsParser.java:43)
    at com.avaje.ebeaninternal.server.query.CQueryPredicates.buildBindWhereRawSql(CQueryPredicates.java:239)
    at com.avaje.ebeaninternal.server.query.CQueryPredicates.prepare(CQueryPredicates.java:276)
    at com.avaje.ebeaninternal.server.query.CQueryPredicates.prepare(CQueryPredicates.java:264)
    at com.avaje.ebeaninternal.server.query.CQueryBuilder.buildQuery(CQueryBuilder.java:193)
    at com.avaje.ebeaninternal.server.query.CQueryEngine.findMany(CQueryEngine.java:162)
    at com.avaje.ebeaninternal.server.query.DefaultOrmQueryEngine.findMany(DefaultOrmQueryEngine.java:80)
    at com.avaje.ebeaninternal.server.core.OrmQueryRequest.findList(OrmQueryRequest.java:284)
[...]

PR #262, opened 2015-03-26
Fix issue #259

ForNeVeR:
New code doesn't use `getPreWhere` in case of unparsed SQL (because `preWhere` is never defined for an unparsed SQL).

~~If you want me to contribute also a test case then please could you shortly introduce me to a testing system - there's a lot of test code, I don't know how it is organized.~~ _(**UPD:** There is a test case now.)_

Please see a description and comments in #259 for some explanation. There's a separate test project also (it can be easily converted to a proper test case I hope): https://github.com/ForNeVeR/ebean-issue-259


ForNeVeR:
Aha, I got it and added a test case.

rbygrave:
Great thanks

rbygrave:
Fixed via associated pull request.

PR #294, opened 2015-05-13
add distinct to db2

andersha:
We use db2 a lot but have used play 2.3.x branch, now with play 2.4.0 near we will start using a newer version of ebean, and then we want this small patch into the mainline. It just adds distinct to the generated SQL when necessary.


rbygrave:
Great. I am looking to release the next version - 4.6.3 in a week ... yet me know if you want a release earlier than that.

Thanks, Rob.

andersha:
That will be fine. We will need this after the release of Play 2.4. Until then we use 2.3 ;)

Sendt fra min iPhone

PR #316, opened 2015-06-30
Resolve NullPointerException on JtaTransacionManager constructor

castovoid:
JtaTransactionManager alway throw a PersistenceException with a NullPointerException. This pull request resolve that.


rbygrave:
Thanks.

Issue #425, opened 2015-09-30
Extra instance of root level bean with findIterate() and fetch of OneToMany relationship - Ebean parses the first bean correctly and the rest not 


archie-sh:
If there are 2 customers ie. customer 1 and customer 2 and .fetch("orders") is used then the output would be the following 

Customer id 1
--> Order id 1
--> Order id 3
--> Order id 4
Customer id 2
--> Order id 2
Customer id 2
--> Order id 5

where it should be 

Customer id 1
--> Order id 1
--> Order id 3
--> Order id 4
Customer id 2
--> Order id 2
--> Order id 5

michaelsiebers:
+1

bond211:
+1

anteoc:
+1

naemar:
+1

rbygrave:
@naemar why +1 a closed issue?

naemar:
ah sorry, left the tab open without pressing comment - no worries

rbygrave:
Cool. I was wondering if that +1 meant something like ... "can I have this released soon".  I'll release in the next day or so BTW.

PR #426, opened 2015-09-30
#425 Ebean parses the first bean correctly and the rest not

archie-sh:
The solution is not optimal but gives a good idea on how to solve the problem


rbygrave:
#425 is fixed and I think this means this pull request should be closed (in preference to the fix included in #425)

Cheers, Rob.

rbygrave:
I'll close this as the fix has come via the commits in  #425

PR #510, opened 2015-12-29
Added support for custom database functions that have multiple arguments

apuckey:
seperated by commas and spaces inside brackets.

this allows things like:

```
select col1, MONTH(MAKEDATE(2015, 241)) as month, col3 from table
```

previously it would split it in the middle of the custom function creating a bad mapping.

Also added a test case for this database function


rbygrave:
Excellent work - thanks.

PR #512, opened 2016-01-03
Develop v3

AlbertAY:


rbygrave:
This looks like a mistake to me in that there should be no merging of commits from develop-v3 into master?

Should we not close this merge request?

rbygrave:
I'll close this pull request - we should not be merging from develop-v3 into master.  develop-v3 is effectively the master for the old version 3.x code base and there if someone is stuck on version 3.x and needs a patch.

I'm sure this is just a mistake of some kind. 

Cheers, Rob.

PR #608, opened 2016-03-18
SQLite Ddl : autoincrement attribute

btimo:
https://www.sqlite.org/autoinc.html

Apparently SQLite syntax doesn't accept the keyword `autoincrement` alone in the datatype.


rbygrave:
Thanks. Yes, I think it is reasonable to leave off autoincrement explicitly as suggested by the SQLite documentation etc. 

Thanks.

PR #645, opened 2016-04-08
review duplicate var

icode:


rbygrave:
Thanks, nice spotting.

PR #735, opened 2016-06-14
Fixing squid:S1210 - "equals(Object obj)" should be overridden along with the "compareTo(T obj)" method

apxeolog-df:
This pull request is focused on resolving occurrences of Sonar rule squid:S1210 - “"equals(Object obj)" should be overridden along with the "compareTo(T obj)" method”. 
You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1210
Please let me know if you have any questions.
Artyom Melnikov.


rbygrave:
I'm pretty sure I'm going to reject this pull request.  There is no intention for MigrationResource or MigrationVersion to support equals/hashCode and doing so would be misleading.

PR #737, opened 2016-06-14
Fixing squid:S1148 - Throwable.printStackTrace(...) should not be called

apxeolog-df:
This pull request is focused on resolving occurrences of Sonar rule 
 squid:S1148 - “Throwable.printStackTrace(...) should not be called”. 
 You can find more information about the issue here: 
 https://dev.eclipse.org/sonar/rules/show/squid:S1148
 Please let me know if you have any questions.
 Artyom Melnikov.

### DIFF
--- a/src/main/java/com/avaje/ebeaninternal/server/lib/DaemonScheduleThreadPool.java
+++ b/src/main/java/com/avaje/ebeaninternal/server/lib/DaemonScheduleThreadPool.java
@@ -62,7 +62,6 @@ public final class DaemonScheduleThreadPool extends ScheduledThreadPoolExecutor 
 
       } catch (Exception e) {
         logger.error("Error during shutdown of " + namePrefix, e);
-        e.printStackTrace();
       }
     }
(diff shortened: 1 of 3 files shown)


dastultz:
On Tue, Jun 14, 2016 at 7:41 AM, Artyom Melnikov <notifications@github.com>
wrote:

> This pull request is focused on resolving occurrences of Sonar rule
> squid:S1148 - “Throwable.printStackTrace(...) should not be called”.
>
I would argue that the 3 calls to print the stack trace are appropriate.
The error is sent to a logger first, but these are all shutdown conditions.
It's possible the loggers could be closed out before the messages can be
written. Sending the stack trace to standard out is a good "just in case"
measure.

/Daryl

rbygrave:
Daryl is exactly correct.  The reasons these also go to standard out is that on the JVM there is effectively a race condition at shutdown and the loggers can be shutdown too early.

I'll reject this pull request.

PR #748, opened 2016-06-22
fix DeployBeanDescriptor.java of the method setView

yuqiyi:
the method setView in DeployBeanDescriptor has wrong of
this.dependentTables = this.dependentTables;


rbygrave:
Thanks, funnily enough I literally just saw that 10 minutes ago as well.

yuqiyi:
That's so funny.Otherwise, I have a problem now.
 I have one table named task, and other table named like items_taskId, the taskId is a dynamic value from table task.
  How can i map the Entity Item to the dynamic table?
  The simple thing want to do is change the final sql "from item" to "from item_xxx".
I just want to select, update, insert of the Item Entity,but not for Multi table operate.
How can I do?

PR #814, opened 2016-08-19
ENHANCEMENT: multiple indices & meta-annotation support similar to spring-framework

rPraml:
This change adds support for multiple indices with @Indices(indices = {@Index(..), @Index(..) }) at class level and at field level. (An index over multiple fields that is specified at field level MUST contain that field. You can use ${fa} as field alias)

Also added "meta-annotation" support similar to spring-framework. You can specify  a bunch of annotations or aliases (for all fields with size=100) e.g.
```
@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Size(max=100)
public @interface SizeMedium {
} 
```


rPraml:
Hmm. I need help :)
As fas as I see does PostGres no size checking (or something else is going wrong)
My test creates a VARCHAR(100) and tries to insert 101 bytes ... which is correctly revoked with a PersistenceException in h2 but not in pg

PR #818, opened 2016-08-30
Some fixes

rPraml:
Hello Rob,

some cherry picked fixes from my other branch.
as you said in the forum, i will undo the most changes next time.
I have a concept how to implement my use cases without too much changes in Ebean.

cheers
Roland


src/main/java/com/avaje/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -798,7 +798,7 @@ public void pathSet(Object bean, Object value) {
 
     if (bean != null) {
       Object logicalVal = convertToLogicalType(value);
-      setValue((EntityBean) bean, logicalVal);
+      setValueIntercept((EntityBean) bean, logicalVal);
```
rbygrave: This looks like the correct change but I need to check this.

PR #828, opened 2016-10-05
Feature/postconstruct

rPraml:
Hello Rob,

here is a pull request for a feature we would need.

Background:
We use spring and our entity-beans should be auto-wire capable. We also need a mechanism to set default vaules for new beans. So we use the enhancement in this pull request together with this listener
```java
config.setPostConstructListeners(Arrays.asList(new BeanPostConstructListener() {
	public void autowire(final Object bean) {
		applicationContext.getAutowireCapableBeanFactory().autowireBean(bean);
	}
	public void postConstruct(final Object bean) {}

	public boolean isRegisterFor(final Class<?> cls) {
		return BaseModel.class.isAssignableFrom(cls);
	}

	public void postCreate(final Object bean) {
		((BaseModel) bean).setDefaults();
	}
}));
```

1.  autowire is called immediately after "_ebean_newInstance()". We pass the generated bean to Spring, so that it gets all autowired properties.
2. in the next step, the listener's postConstruct and the methods of the bean that are annotated with `javax.annotation.PostConstruct` are invoked.
3. if this is a new bean (= created with EbeanServer.createEntityBean) the listener's postCreate method is invoked, which can set default values in the bean.
(If there's a better way to set default values for new beans i'm willing to undo this change)

I tried to separate this pull request into compact commits.

cheers
Roland

### DIFF
--- a/src/main/java/com/avaje/ebean/EbeanServer.java
+++ b/src/main/java/com/avaje/ebean/EbeanServer.java
@@ -167,8 +167,9 @@ public interface EbeanServer {
   /**
    * Create a new instance of T that is an EntityBean.
    * <p>
-   * Generally not expected to be useful (now dynamic subclassing support was removed in
-   * favour of always using enhancement).
+   * Useful if you use BeanPostConstructListeners or &#64;PostConstruct Annotations.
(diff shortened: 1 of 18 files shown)


rbygrave:
Very nice work !!!

PR #841, opened 2016-10-26
Bugfix/softdelete calls wrong lifecycle method

rPraml:
the lifecycle calls when soft-deleting an entity were "preUpdate" / "postDelete" - which is definitiefly wrong.
it should either call "preUpdate" / "postUpdate" or "preDelete" / "postDelete" (or pre/postSoftDelete?)

As soft delete/undelete is technically just an update, I think it is ok to call pre/postUpdate now.

I also have fixed some trailing commas in the debug outputs.

cheers
Roland


rbygrave:
Awesome, thank you.

0xbaadf00d:
Is it possible to add two other callback `postSoftDelete` and `preSoftDelete` ? Because, when soft delete is used, developer maybe don't want to perform same action when `.delete()` and when `.save()` are called.

PR #851, opened 2016-11-04
FIX some java warnings / rawtypes

rPraml:
Hello Rob, I hope I do not annoy you with my pull requests ;)

but here's a pull request where I've fixed some java warnings - there should be no effective code change.

cheers
Roland


rbygrave:
Awesome, that's great thanks.  Yes, looks to be no effective change to me so merging - nice tidy up.  With Java 8 (+ Java 7) it is good to get into this type of cleanup work so thanks for this.

> I do not annoy you with my pull requests ;)

No that's great.  You've added some great features etc so keep up the great work!!

Cheers, Rob.

Issue #850, opened 2016-11-04
Migrate to Java 8 as a requirement (for Ebean 9.x) 


rbygrave:
As per the discussion at: https://groups.google.com/forum/#!topic/ebean/l9EzvdlsbSg

Ebean has been open source for 10 years now (Nov 2006) so this seems poetic for this anniversary.

rbygrave:
A general note to people watching the project.

There is an opportunity to do some pull requests to migrate to the diamond operator `<>`.  If you have not done a github merge request before and are keen to give it a try then I'd encourage you to give it a try.  

I'd encourage you to take a couple of classes, change the explicit collection types to `<>` and submit the merge request. These small merge requests are easy for me to review (so quick to merge) and might be a good way to "dip your toes in" for contributing to other OSS projects as well.

Cheers, Rob.

PR #860, opened 2016-11-06
Replace anonymous classes with lambda expressions

0xbaadf00d:
According to #850 (Java 8 migration), this pull request replace all anonymous classes with lambda expressions.


rbygrave:
Thanks, good tidy up work !!

PR #864, opened 2016-11-07
Remove redundant local variables

0xbaadf00d:
This pull request remove redundant local variables which are immediately returned. These variables add nothing to the comprehensibility of a method.


rbygrave:
Thanks!!

PR #873, opened 2016-11-08
Batching exception safety

ksperling:
This adds some tests for exception safety around batchOnCascade handling, plus a second commit with my attempt at fixing the issue.

The `testBatchScenarioWithSavepoint` is the scenario motivating this for me: I've got code in a few places where I use a save point to make an insert or update that may fail due to a constraint violation. If it does fail, I want to be able to roll  back to the save point and take a different course of action (e.g. update an existing record instead of creating a new one etc). This kind of code broke after upgrading from ebean 7.5 to 8, because the batch and batch mode do not get correctly reset when an operation fails.

Additional tests and changes might be required when working with batching enabled at the top level, I haven't tested.


rbygrave:
Looks good.  Thanks Karsten.

PR #884, opened 2016-11-21
Wildcard embedded columns

rPraml:
I've added a featue to support wildcard mapping for @Embedded
e.g. `@EmbeddedColumns(columns = "*=bill_*")`will prefix all properties of the @Embeddable with "bill_"


rbygrave:
So this is a nice idea / change ... and well merge it in.

However, it makes me wonder if in fact the original idea / design of EmbeddedColumns is not very good.  That is, if instead we had:

@EmbeddedColumns(prefix="bill_")

... if could be that this approach actually supports 90% of the time we want to override the column mapping for @Embedded beans.  Perhaps we are better off:

- Adding the more explicit "prefix" style support via @EmbeddedColumns(prefix="bill_")
- Deprecating the existing  @EmbeddedColumns(columns = "...")

PR #887, opened 2016-11-22
Add NamingConvention Support for JoinColumn (OneToMany, ManyToOne)

lvffilho:


rbygrave:
What is the use case (example) where the existing functionality is not working?  Can you give an example of the issue this fixes (before and after).

Thanks

rbygrave:
What's going on here?

lvffilho:
If a have a class with JoinColumn define, and configure a CamelCaseNamingConvention. Ebean do not call convert method and name still "area_type_id". The correct would be areaTypeId using my namingConvention. The same problem occur whem my @JoinColumn dont have a "name" property the framework only concat property + _ + fk and dont call naming convention.

```
@Entity
public class Area extends BaseEntity implements Serializable {

    private static final long serialVersionUID = 1L;

    @ManyToOne
    @JoinColumn(name = "area_type_id")
    @NotNull
    private AreaType areaType;
```

rbygrave:
Fixed as https://github.com/ebean-orm/ebean/issues/906

I'll close this one.

PR #898, opened 2016-11-24
Basic mssql 2012 support

rPraml:
Hello Rob,

it would be great, if you can review the changes for sqlserver 2012.
BTW: there is a divergence between Platform and DatabasePlatform.
Especially in AnnotationBase you are trying to convert: `this.platform = Platform.valueOf(databasePlatform.getName().toUpperCase());`
So, why not using the Platform property directly?

cheers
Roland

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,34 @@
     <tag>HEAD</tag>
   </scm>
 
+  <profiles>
+  	<profile>
+		<!-- Note: SQLServer 2012 support is not yet stable! --> 
+  		<id>sqlserver-tests</id>
+  		<properties>
+  			<datasource.default>ms</datasource.default>
+  		</properties>
+  		<dependencies>
+			<dependency>
+				<groupId>com.microsoft.sqlserver</groupId>
+				<artifactId>sqljdbc42</artifactId>
+				<version>4.2</version>
+				<scope>test</scope>
+			</dependency>
+  		</dependencies>
+  		<repositories>
+			<repository>
+				<id>foconis</id>
+				<url>https://repo.foconis.de/maven-release/</url>
+			</repository>
+			<repository>
+				<id>foconis-snapshot</id>
+				<url>https://repo.foconis.de/maven-snapshot/</url>
+			</repository>
+		</repositories>
+  	</profile>
+  </profiles>
+  
   <dependencies>
 
     <dependency>
--- a/src/main/java/com/avaje/ebean/config/dbplatform/BasicMsSqlLimiter.java
+++ b/src/main/java/com/avaje/ebean/config/dbplatform/BasicMsSqlLimiter.java
@@ -1,0 +1,25 @@
+package com.avaje.ebean.config.dbplatform;
+
+/**
+ * Adds MS-SQL Limiter support.
+ * Introduced in MS SQL Server 2012 enhancing the ORDER BY clause in with OFFSET/FETCH support.
+ */
+public class BasicMsSqlLimiter implements BasicSqlLimiter {
+
+  @Override
+  public String limit(String dbSql, int firstRow, int maxRows) {
+
+    StringBuilder sb = new StringBuilder(50 + dbSql.length());
+    sb.append(dbSql);
+    if (!dbSql.toLowerCase().contains("order by")) {
+      sb.append(" ORDER BY 1 ");
+    }
+    sb.append(" ").append("offset");
+    sb.append(" ").append(firstRow).append(" rows");
+    if (maxRows > 0) {
+      sb.append(" fetch next ").append(maxRows).append(" rows only");
+    }
+    return sb.toString();
+  }
+
+}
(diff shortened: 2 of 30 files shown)


src/main/java/com/avaje/ebean/config/dbplatform/MsSqlServer2005Platform.java
```diff
@@ -17,7 +17,7 @@
 
   public MsSqlServer2005Platform() {
     super();
-    this.name = "mssqlserver2005";
+    this.name = "mssqlserver";
```
rPraml: ups. should be named "sqlserver"

pom.xml
```diff
@@ -24,6 +24,34 @@
...
+				<groupId>com.microsoft.sqlserver</groupId>
+				<artifactId>sqljdbc42</artifactId>
+				<version>4.2</version>
+				<scope>test</scope>
+			</dependency>
+  		</dependencies>
+  		<repositories>
+			<repository>
+				<id>foconis</id>
+				<url>https://repo.foconis.de/maven-release/</url>
```
rPraml: I need the sqljdbc42 driver from microsoft which is not on mvncentral, (ok, there is one http://mvnrepository.com/artifact/org.formcept - but no official one)
so I put it to our local repository (not reachable from the internet)

src/main/java/com/avaje/ebean/config/dbplatform/BasicMsSqlLimiter.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class BasicMsSqlLimiter implements BasicSqlLimiter {
+
+  @Override
+  public String limit(String dbSql, int firstRow, int maxRows) {
+
+    StringBuilder sb = new StringBuilder(50 + dbSql.length());
+    sb.append(dbSql);
+    if (!dbSql.toLowerCase().contains("order by")) {
+      sb.append(" ORDER BY 1 ");
```
rPraml: SQL-Server always needs an order by clause to support "pagination"

0xbaadf00d:
Is these modifications for MS-SQL 2012 only or for all MS-SQL server version >= 2012 ?

src/main/java/com/avaje/ebeaninternal/server/persist/dml/InsertMeta.java
```diff
@@ -18,6 +19,9 @@
  */
 public final class InsertMeta {
 
+  private enum IdGeneration {
+    BY_SERVER, BY_CLIENT, BY_CLIENT_WITH_IDENTIY_INSERT;
```
rPraml: This is the most complex part, basically there are two ways to generate an id
1. by server - when the id property is null (e.g. auto-increment/sequence)
2. by client - when the id property is set (e.g. uuid)

Unfortunately, sqlserver fails if the client tries to set/enforce an id on an auto-increment column. If you want to do this, you must prepend a "SET IDENTITY_INSERT tablename ON" before your insert statement.

So what I do here:
1. Determine if the ID of the bean is generated by the server
2. if yes - Determine, if the platform needs identity_insert
3. if yes - surround the SQL insert with "set identity_insert on/off"

rPraml:
@0xbaadf00d it should work for versions >= 2012.  
To be clear: I don't know if it will work for 2012, I run the tests against "Microsoft SQL Server 2014 - 12.0.2269.0"
I named the platform "MsSqlServer2012" because with 2012 there came the "OFFSET/FETCH" feature to support "limit by" queries. (Before 2012 you have to write rather complex statements to get the enties "limit by 10,20" )

rbygrave:
*> by server - when the id property is null (e.g. auto-increment/sequence)*

SQL Server does not support getting the Id value back when using JDBC batch
with auto-increment (Identity).  I suspect this is not going to change
anytime soon.

With that in mind I believe the best approach (well, the one I'd recommend)
is to use sequences and have that set via the client (via Ebean). If the
sequences are created with a decent step (say increment by 50) then this
can be very efficient with the only downside being the possibility of gaps
in the dev/testing case (where the EbeanServer is started/stopped a lot and
it grabs the next sequence and moves the value 50+ leaving gaps - it
doesn't really matter of course but this upsets some people).

This approach then matches the Oracle approach.

The only issue with adopting this approach is a decent number of existing
Ebean/SQLServer users will be using identity.


Cheers, Rob.

src/main/java/com/avaje/ebeaninternal/server/persist/dml/InsertMeta.java
rbygrave: I'm unconvinced by this part/change.  We should just adopt sequences (with a good increment by) and get full JDBC batch support.  If SqlServer ever supports Identity with getGeneratedKeys using JDBC batch then we revisit (but it hasn't happened and at this point (some years later) I suspect it won't ... so instead adopt sequences and operate like Oracle for JDBC batching inserts).

PR #923, opened 2016-12-14
Pass dataType to DataBind for Types.OTHER

maksym-kit:
This change allows to use ScalarType to allow to use Java enums in where() Expressions.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/persist/Binder.java
+++ b/src/main/java/io/ebeaninternal/server/persist/Binder.java
@@ -324,7 +324,7 @@ public class Binder {
           break;
 
         case java.sql.Types.OTHER:
-          b.setObject(data);
+          b.setObject(data, dataType);
           break;
 
         case java.sql.Types.JAVA_OBJECT:


0xbaadf00d:
Hmm not sure about your patch... we can already use Enumeration in where predicates.

Ex:

```java
ContractModel.find
            .where()
            .eq("status", ContractStatus.CLOSED)
            .findEach(...);
```

rbygrave:
As per the comment from @0xbaadf00d ... Ebean already allows Java enums in where clause.  Unless there is a failing test you wish to add this PR should be closed.

Thanks, Rob.

maksym-kit:
EBean maps enums to varchar or integer and does not work when column is of custom enum type. For such case EBean supports ScalarType extensions to handle such types and ScalarType correctly works for INSERT and UPDATE statements. However, when such column was used in WHERE clause, EBean does not pass Types.OTHER to PreparedStatement.setObject and PostgreSQL throws an SQL error that custom enum type does not match varchar.

Test implementation requires access to running PostgreSQL. Should test case be included in your build or one-shot test case that shows misbehavior will suffice?

rbygrave:
Yes a one-shot test will suffice.  Perhaps a fork of https://github.com/ebean-orm-examples/example-minimal ... but I don't really mind. I just want to see the full context for this merge request. 

... only now have I noticed the name of your branch is "postgresql-enum-support". That is, it was not previously clear that this branch was regarding the use of PG enums.

0xbaadf00d:
What the impact on existing query using PostreSQL when we migrate to the next Ebean version ?

rbygrave:
This should only effect custom scalartype implementations that map to jdbc type java.sql.Types.OTHER

0xbaadf00d:
Ok, thanks

vnnv:
Is that mean that now I can define PostgreSQL enum and use it in select and insert/update statements?

rbygrave:
> Is that mean that now I can define PostgreSQL enum a

Yes, if you manage your own DDL and define your own ScalarType - otherwise no.

PR #929, opened 2016-12-22
SqlServer 2016 history support.

vilmosnagy:
Hi Rob!

I've implemented the basic SQL2011 History support for SQL Server 2016.

I'm not totally sure about the DbMigration part - but the queries works pretty fine, if the `system_versioning`  is turned on the tables.

I'm not sure, how to distinguish SQL Server 2016 from previous SQL Servers, so currently I've to set this by hand in the `ServerConfig` instance.


rbygrave:
That is great work - thanks, I'll merge it in.

> distinguish SQL Server 2016 

For me I feel we actually don't have to distinguish.  That is, people have to "opt in" by adding @History ... and they should know that they can only do that on an appropriate version of SqlServer.

What that means is that I feel we can move this to SqlServerPlatform ... (so I might look to do that after the merge).

Cheers, Rob.

PR #937, opened 2017-01-02
MigrationXmlWriter adds DOCTYPE to xml

focbenz:
This avoids warning of missing external DTD in eclipse IDE and maybe other xml validators.


rbygrave:
Thanks!

PR #942, opened 2017-01-09
WhoCreated, WhoModified on OneToMany connections.

vilmosnagy:
Hi Rob,

as discussed in the Google Groups thread, [here](https://groups.google.com/forum/#!topic/ebean/Te1qct1u2m4), I'd like to make a PR.

Any feedback is welcomed, and I'll fix.

Thanks,

### DIFF
--- a/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationFields.java
+++ b/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationFields.java
@@ -140,6 +140,21 @@ public class AnnotationFields extends AnnotationParser {
     if (formula != null) {
       prop.setSqlFormula(formula.select(), formula.join());
     }
+
+    initWhoCreated(prop);
+    initWhoModified(prop);
+  }
+
+  private void initWhoModified(DeployBeanProperty prop) {
+    if (get(prop, WhoModified.class) != null) {
+      generatedPropFactory.setWhoModified(prop);
+    }
+  }
+
+  private void initWhoCreated(DeployBeanProperty prop) {
+    if (get(prop, WhoCreated.class) != null) {
+      generatedPropFactory.setWhoCreated(prop);
+    }
   }
 
   private void readField(DeployBeanProperty prop) {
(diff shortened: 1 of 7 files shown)


rbygrave:
Well done, looks good!!

I'll look to merge it in tonight ... I'll probably look to change BindableAssocOne to have inheritance (add new BindableAssocOneGeneratedInsert, BindableAssocOneGeneratedUpdate classes) which would mean we can remove all the if statements from executing at runtime and give us a small runtime performance improvement. I'll see how that looks anyway.

Cheers, Rob.

src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationFields.java
```diff
@@ -282,12 +297,8 @@ private void readField(DeployBeanProperty prop) {
...
     }
 
-    if (get(prop, WhoCreated.class) != null) {
-      generatedPropFactory.setWhoCreated(prop);
-    }
-    if (get(prop, WhoModified.class) != null) {
-      generatedPropFactory.setWhoModified(prop);
-    }
+    initWhoCreated(prop);
+    initWhoModified(prop);
```
0xbaadf00d: What about performance with this extra call ? Not sure about creating a method just for one instruction is a good id here.

rbygrave: True, although these methods are only called during initialisation of EbeanServer (bootup time) and not what I'd call runtime (for every insert or update times every BindableAssocOne).  So you might be right but at this stage I haven't looked at performance optimisation of "bootup time" much and I'm happy with initWhoCreated() as it is called from 2 places etc.  

So for me I think this is all good.

vilmosnagy: I could merge these two methods into one (something called `initCommonAnnotationsOnFields`). That would mean one less method call with the same code quality (in my opinion).

rbygrave: I'd say just leave it.  I'll merge it in shortly and square it away - all good.  
Cheers, Rob.

PR #949, opened 2017-01-17
FIX: replaced some hard coded id columns

rPraml:
I've found some hard-coded "id" strings. This will fail, if PK is not named "id"

@rbygrave Maybe you have a look at these corrections.

Cheers,
Roland

### DIFF
--- a/src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
+++ b/src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
@@ -288,7 +288,7 @@ public class DefaultBeanLoader {
     }
 
     if (embeddedOwnerIndex > -1) {
-      query.select("id," + ebi.getProperty(embeddedOwnerIndex));
+      query.select(desc.getIdProperty().getName() + "," + ebi.getProperty(embeddedOwnerIndex));
     }
 
     // don't collect AutoTune usage profiling information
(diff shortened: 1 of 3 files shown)


rbygrave:
2 of these 3 changes are not right.  I'll look at this again in a bit.

src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
```diff
@@ -435,7 +435,7 @@ private void addProperty(SqlTreeProperties selectProps, BeanDescriptor<?> desc,
       BeanProperty p = desc.findBeanProperty(propName);
       if (p == null) {
         logger.error("property [" + propName + "] not found on " + desc + " for query - excluding it.");
-        p = desc.findBeanProperty("id");
```
rPraml: If I remember correctly, I ran into an NPE when using query.select("...wrong_spelled_prop...") on an entity which primary key was not named "id" but "key". Should I provide a test case for this?

src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
```diff
@@ -288,7 +288,7 @@ private void refreshBeanInternal(EntityBean bean, SpiQuery.Mode mode, int embedd
     }
 
     if (embeddedOwnerIndex > -1) {
-      query.select("id," + ebi.getProperty(embeddedOwnerIndex));
```
rPraml: Maybe this was the select that triggers the NPE mentioned below... I really should write a test that addresses this issue...

rbygrave: Test case first ... yes.  This change is wrong though ... it should just not include the id property at all.

rbygrave:
This can be closed.  I've pulled the change we want from this into: https://github.com/ebean-orm/ebean/issues/973

rPraml:
thx

PR #950, opened 2017-01-17
Fix/notnull classes

rPraml:
Hello Rob,

here is a pull request that solves my NotNull & EbeanDDL problem.
We can drop the EbeanDDL support from now (I hope I was the only one using that feature)

I also have done some improvements in tests:
- fixed packages in dynamic enhancer
- Deleted Junit-3 style EbeanTestCase and using Junit-4 style BaseTestCase instead, because this loads the dynamic enhancer.

Cheers,
Roland


rbygrave:
I'm not sure about the design of this (relative to the not null issue per say).  Leaving this on backlog but I'm not keen on this change currently.

rPraml:
Shoud I create a PR that converts the old junit 3.8 tests to 4?

rbygrave:
You mean effectively remove EbeanTestCase I think ... yes, good.

rbygrave:
We should close this PR.

PR #951, opened 2017-01-17
ENH: Support to generate and run DbMigration per schema

rPraml:
Maybe you can take a look at these changes and give some comments.


rPraml:
I am still investigating in this and I will hopefully provide a better solution

rPraml:
Will close this pull request as it is obsolete

PR #952, opened 2017-01-18
ENH: Add maven-invoker-plugin to perform IT-tests

rPraml:
Hello Rob,

I've added the maven-invoker plugin to write IT tests that are decoupled from the src/test class paths.
I also provide an example how to use it.

Cheers
Roland

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 target/
 logs/
 log/
+it-repo/
 ebean-autotune.xml
 ebean-profiling*.xml
 /db
--- a/pom.xml
+++ b/pom.xml
@@ -313,6 +313,45 @@
         </executions>
       </plugin>
 
+      <!-- plugin for IT tests. Taken from https://github.com/reficio/p2-maven-plugin -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-invoker-plugin</artifactId>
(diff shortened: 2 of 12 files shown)


rbygrave:
I'm not sure what this is trying to achieve per say.

rPraml:
I can simulate how ebean behaves in a real application with it own set of models, own migration scripts, own configuration and so on. Yes, most of them is already covered by the testing framework in ebean, but I had no idea how I can set up test cases for ddl generation/migration. I did this in my own project before.

In other words, to write an integration test, you have to create a complete maven project in "src/test/integration/..." use the just packaged ebean as maven dependency and run your test. After the test you can run a validate.groovy script to check generated files (you can also write common junit tests that will fail if groovy is not your favourite language :)

This test e.g. https://github.com/FOCONIS/ebean/tree/master/src/test/integration/test-ddl-migration discovered the caching issue in datasource and also in H2: https://groups.google.com/forum/#!topic/h2-database/eJGE-CD_qhk

I know this PR adds some dependencies and writing IT-tests is not so easy as in ebean/src/test, because you need a pom.xml + ebean.properties/logback/... , but on the other hand, IT tests give you a better control about used dependencies and so on. And maybe the one or other IT-test can considered also as example how to use a certain feature.

rPraml:
closing this PR

PR #962, opened 2017-02-01
Fix warnings

rPraml:
Hello Rob,

a very big pull request that fixes a lot of (eclipse) warnings. There is one real code change in PlatformDdlWriter, all other changes should not affect existing code.

### DIFF
--- a/src/main/java/io/ebean/dbmigration/DbMigration.java
+++ b/src/main/java/io/ebean/dbmigration/DbMigration.java
@@ -374,7 +374,7 @@ public class DbMigration {
         // writer needs the current model to provide table/column details for
         // history ddl generation (triggers, history tables etc)
         DdlWrite write = new DdlWrite(new MConfiguration(), request.current);
-        PlatformDdlWriter writer = createDdlWriter(databasePlatform, "");
+        PlatformDdlWriter writer = createDdlWriter(databasePlatform);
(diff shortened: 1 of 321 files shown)


rPraml:
Hello Rob, should I clean up/split this PR?

src/main/java/io/ebean/dbmigration/DbMigration.java
```diff
@@ -431,20 +431,18 @@ protected void writeExtraPlatformDdl(String fullVersion, CurrentModel currentMod
 
     for (Pair pair : platforms) {
       DdlWrite platformBuffer = new DdlWrite(new MConfiguration(), currentModel.read());
-      PlatformDdlWriter platformWriter = createDdlWriter(pair);
+      PlatformDdlWriter platformWriter = createDdlWriter(pair.platform);
```
rbygrave: I'm not that happy when a PR contains multiple changes and is so massive.  This one contains a whole bunch of @Override ... but then also includes a design change for DDL writing.  

We don't what this type of change buried in a PR that changes 300 files.  If a PR is going to add @Override everywhere that is OK as long as that is all the PR does.  It is not ok to also include other subtle changes like this.

This needs to be 2 PRs.

rbygrave:
Yes, you need to split this PR or close it.

rPraml:
closed - created new PR that adds only @overrides

PR #971, opened 2017-02-18
Staticanalysis cleanups

KoenDG:
As discussed here: https://groups.google.com/forum/#!topic/ebean/ZXedzZhaP3U

This pull request is the result of executing several static analysis tools on the code, looking at the results and introducing some small refactors.

There is one functionality change in here, as suggested by @rPraml which employs nanoseconds in the defaultservercache, as opposed to System.currentTimeMillis(), which, quote, "as currentTimeMillis depends on the system clock which can give you unpredictable results on DST switch or leap seconds.", end quote.

There's other points in the code where I'm unsure of something, in which case I left a question asking about it.

Besides that, it's mainly cleaning up imports, forcing the editorconfig on files that weren't formatted properly and some small code cleanups.

There's also:

- initializing lists with an appropriate size if it's possible to predict it at the point in the code.
- making inner classes static to prevent implicit references to the enclosing classes in the compiled code. These references can prevent the enclosing class from being GC'd, at which point you have yourself a memory leak.
- The `EnumMap` class was applicable in one instance. This is a specialised Map where the key is an Enum, extremely compact and efficient: https://docs.oracle.com/javase/8/docs/api/java/util/EnumMap.html
- try-with-resources implemented where possible
- Moving declarations/initializations to later points in the code. This is only done when between the creation of the variable and its first use there is a return or an exception. There's no point in creating these variables before a piece of code returns/throws if the variable only used after that.
- Removing null checks before instanceof checks. Those are redundant since "null instanceof object" always returns false.
- Using "" + someInt to get a string of that Integer is inefficient. String.valueOf() is preferred.

That should be it. I'll be making comments on the pull request and removing those same comments from the code so they're not in the code anymore but only on the PR.

EDIT: Once Github is under less load, which is currently preventing me from doing so.

EDIT2: Seems like removing the comment from the code also makes the comment on github disappear. Leaving them in for now.

### DIFF
--- a/src/main/java/io/ebean/Model.java
+++ b/src/main/java/io/ebean/Model.java
@@ -175,6 +175,7 @@ public abstract class Model {
    * @param propertyName the name of the property on the bean to be marked as 'unset'
    */
   public void markPropertyUnset(String propertyName) {
+    //  Intellij claims the cast to EntityBean is incompatible
     ((EntityBean) this)._ebean_getIntercept().setPropertyLoaded(propertyName, false);
   }
 
--- a/src/main/java/io/ebean/Model.java
+++ b/src/main/java/io/ebean/Model.java
@@ -175,7 +175,6 @@ public abstract class Model {
    * @param propertyName the name of the property on the bean to be marked as 'unset'
    */
   public void markPropertyUnset(String propertyName) {
-    //  Intellij claims the cast to EntityBean is incompatible
     ((EntityBean) this)._ebean_getIntercept().setPropertyLoaded(propertyName, false);
   }
 
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -279,7 +279,7 @@ public final class EntityBeanIntercept implements Serializable {
    * Return true if the entity bean is new or dirty (and should be saved).
    */
   public boolean isNewOrDirty() {
-    return isNew() || isDirty();
+      return isNew() || dirty;
   }
 
   /**
@@ -534,7 +534,7 @@ public final class EntityBeanIntercept implements Serializable {
    */
   public void markPropertyAsChanged(int propertyIndex) {
     setChangedProperty(propertyIndex);
-    setDirty(true);
+      this.dirty = true;
   }
 
   public void setChangedProperty(int propertyIndex) {
@@ -577,7 +577,7 @@ public final class EntityBeanIntercept implements Serializable {
         changedProps[i] = true;
       }
     }
-    setDirty(true);
+      this.dirty = true;
   }
 
   /**
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -858,7 +858,7 @@ public final class EntityBeanIntercept implements Serializable {
     loadedProps[propertyIndex] = true;
   }
 
-  private final void preGetterCallback() {
+  private void preGetterCallback() {
     if (preGetterCallback != null) {
       preGetterCallback.preGetterTrigger();
     }
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -279,7 +279,7 @@ public final class EntityBeanIntercept implements Serializable {
    * Return true if the entity bean is new or dirty (and should be saved).
    */
   public boolean isNewOrDirty() {
-      return isNew() || dirty;
+    return isNew() || isDirty();
   }
 
   /**
@@ -534,7 +534,7 @@ public final class EntityBeanIntercept implements Serializable {
    */
   public void markPropertyAsChanged(int propertyIndex) {
     setChangedProperty(propertyIndex);
-      this.dirty = true;
+    setDirty(true);
   }
 
   public void setChangedProperty(int propertyIndex) {
@@ -567,7 +567,6 @@ public final class EntityBeanIntercept implements Serializable {
    * For forced update on a 'New' bean set all the loaded properties to changed.
    */
   public void setNewBeanForUpdate() {
-
     if (changedProps == null) {
       changedProps = new boolean[owner._ebean_getPropertyNames().length];
     }
@@ -577,7 +576,7 @@ public final class EntityBeanIntercept implements Serializable {
         changedProps[i] = true;
       }
     }
-      this.dirty = true;
+    setDirty(true);
   }
 
   /**
--- a/src/main/java/io/ebean/config/ClassLoadConfig.java
+++ b/src/main/java/io/ebean/config/ClassLoadConfig.java
@@ -118,7 +118,9 @@ public class ClassLoadConfig {
   /**
    * Wraps the preferred, caller and context class loaders.
    */
-  protected class ClassLoaderContext {
+  protected static class ClassLoaderContext {
+    // A static inner class does not keep an implicit reference to its enclosing class. References to the enclosing class prevents it from being gc'd
+    // Can be static since it does not reference its enclosing class.
 
     /**
      * Optional - if set only use this classLoader (no fallback).
--- a/src/main/java/io/ebean/config/JsonConfig.java
+++ b/src/main/java/io/ebean/config/JsonConfig.java
@@ -5,6 +5,9 @@ package io.ebean.config;
  */
 public abstract class JsonConfig {
 
+  // According to Intellij, this class has no concrete subclass.
+  // Refactor by extracting the 2 enums?
+
   /**
    * Defined the format used for DateTime types.
    */
--- a/src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
+++ b/src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
@@ -5,43 +5,43 @@ import io.ebean.config.dbplatform.DbEncryptFunction;
 
 /**
  * Oracle encryption support.
- *
+ * <p>
  * <p>
  * You will typically need to create your own encryption and decryption
  * functions similar to the example ones below.
  * </p>
- *
+ * <p>
  * <pre class="code">
- *
- *  // Remember your DB user needs execute privilege on DBMS_CRYPTO
- *  // as well as your encryption and decryption functions
- *
- *
- *  // This is an Example Encryption function only - please create your own.
- *
+ * <p>
+ * // Remember your DB user needs execute privilege on DBMS_CRYPTO
+ * // as well as your encryption and decryption functions
+ * <p>
+ * <p>
+ * // This is an Example Encryption function only - please create your own.
+ * <p>
  * CREATE OR REPLACE FUNCTION eb_encrypt(data IN VARCHAR, key in VARCHAR) RETURN RAW IS
- *
- *     encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
- *
- *     BEGIN
- *          RETURN DBMS_CRYPTO.ENCRYPT(UTL_I18N.STRING_TO_RAW (data, 'AL32UTF8'),
- *            encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8') );
- *     END;
- *     /
- *
- *
- *
- *  // This is an Example Decryption function only - please create your own.
- *
+ * <p>
+ * encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
+ * <p>
+ * BEGIN
+ * RETURN DBMS_CRYPTO.ENCRYPT(UTL_I18N.STRING_TO_RAW (data, 'AL32UTF8'),
+ * encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8') );
+ * END;
+ * /
+ * <p>
+ * <p>
+ * <p>
+ * // This is an Example Decryption function only - please create your own.
+ * <p>
  * CREATE OR REPLACE FUNCTION eb_decrypt(data IN RAW, key IN VARCHAR) RETURN VARCHAR IS
- *
- *     encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
- *
- *     BEGIN
- *          RETURN UTL_RAW.CAST_TO_VARCHAR2(DBMS_CRYPTO.DECRYPT
- *            (data, encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8')));
- *     END;
- *     /
+ * <p>
+ * encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
+ * <p>
+ * BEGIN
+ * RETURN UTL_RAW.CAST_TO_VARCHAR2(DBMS_CRYPTO.DECRYPT
+ * (data, encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8')));
+ * END;
+ * /
  * </pre>
  */
 public class OracleDbEncrypt extends AbstractDbEncrypt {
--- a/src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
+++ b/src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
@@ -5,20 +5,20 @@ import io.ebean.config.dbplatform.DbEncryptFunction;
 
 /**
  * Oracle encryption support.
- * <p>
+ *
  * <p>
  * You will typically need to create your own encryption and decryption
  * functions similar to the example ones below.
  * </p>
- * <p>
+ *
  * <pre class="code">
- * <p>
+ *
  * // Remember your DB user needs execute privilege on DBMS_CRYPTO
  * // as well as your encryption and decryption functions
- * <p>
- * <p>
+ *
+ *
  * // This is an Example Encryption function only - please create your own.
- * <p>
+ *
  * CREATE OR REPLACE FUNCTION eb_encrypt(data IN VARCHAR, key in VARCHAR) RETURN RAW IS
  * <p>
  * encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
@@ -28,15 +28,15 @@ import io.ebean.config.dbplatform.DbEncryptFunction;
  * encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8') );
  * END;
  * /
- * <p>
- * <p>
- * <p>
+ *
+ *
+ *
  * // This is an Example Decryption function only - please create your own.
- * <p>
+ *
  * CREATE OR REPLACE FUNCTION eb_decrypt(data IN RAW, key IN VARCHAR) RETURN VARCHAR IS
- * <p>
+ *
  * encryption_mode NUMBER := DBMS_CRYPTO.ENCRYPT_AES128 + DBMS_CRYPTO.CHAIN_CBC  + DBMS_CRYPTO.PAD_PKCS5;
- * <p>
+ *
  * BEGIN
  * RETURN UTL_RAW.CAST_TO_VARCHAR2(DBMS_CRYPTO.DECRYPT
  * (data, encryption_mode, UTL_I18N.STRING_TO_RAW(key, 'AL32UTF8')));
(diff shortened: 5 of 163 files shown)


src/main/java/io/ebean/Model.java
```diff
@@ -175,6 +175,7 @@ public void markAsDirty() {
    * @param propertyName the name of the property on the bean to be marked as 'unset'
    */
   public void markPropertyUnset(String propertyName) {
+    //  Intellij claims the cast to EntityBean is incompatible
     ((EntityBean) this)._ebean_getIntercept().setPropertyLoaded(propertyName, false);
```
KoenDG: Intellij claims the cast to EntityBean is incompatible

src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -859,7 +858,7 @@ public void initialisedMany(int propertyIndex) {
     loadedProps[propertyIndex] = true;
   }
 
-  private final void preGetterCallback() {
+  private void preGetterCallback() {
```
KoenDG: Redundant. Making a method final only serves to prevent if from being override. Which making it private also does.

Explained here: https://stackoverflow.com/questions/23161362/why-can-private-method-be-final

EDIT: Dammit this wasn't supposed to be a review comment why is that the green option?

src/main/java/io/ebean/config/ClassLoadConfig.java
```diff
@@ -118,7 +118,9 @@ public ClassLoader getClassLoader() {
   /**
    * Wraps the preferred, caller and context class loaders.
    */
-  protected class ClassLoaderContext {
+  protected static class ClassLoaderContext {
```
KoenDG: As stated by the comment: this class is an inner class. Making it static results in it not keeping an implicit reference to its enclosing class. Such a reference could prevent the enclosing class from being GC'd. So that situation is a potential memory leak.

src/main/java/io/ebean/config/JsonConfig.java
```diff
@@ -5,6 +5,9 @@
  */
 public abstract class JsonConfig {
 
+  // According to Intellij, this class has no concrete subclass.
+  // Refactor by extracting the 2 enums?
+
   /**
```
KoenDG: As stated by the comment: this class contains 2 Enums that have no extra functionality? Split into 2 separate files?

src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
```diff
@@ -13,12 +13,11 @@
  *
  * <pre class="code">
  *
- *  // Remember your DB user needs execute privilege on DBMS_CRYPTO
```
KoenDG: This is the result of having Intellij accept the editorconfig file as default for the project and then autoformatting the entire project.

src/main/java/io/ebean/dbmigration/model/PlatformDdlWriter.java
```diff
@@ -29,7 +29,7 @@
 
   private final DatabasePlatform platform;
 
-  private final String platformPrefix;
+  private final String platformPrefix; // Not used anywhere
```
KoenDG: Leave it in?

src/main/java/io/ebean/util/StringHelper.java
```diff
@@ -68,14 +68,15 @@
...
-      String value = tag.substring(equalsPos + 2, endQuotePos);
       // dp("name="+name+"; value="+value+";");
 
       // trim off any whitespace from the front of name...
       name = trimFront(name, " ");
       if ((name.indexOf(SINGLE_QUOTE) > -1) || (name.indexOf(DOUBLE_QUOTE) > -1)) {
         throw new RuntimeException("attribute name contains a quote [" + name + "]");
       }
+
+      String value = tag.substring(equalsPos + 2, endQuotePos);
```
KoenDG: Moved this here, past the throw statement. No point in creating the variable in memory and then potentially exiting the context without the variable ever having been used.

src/main/java/io/ebean/util/StringHelper.java
```diff
@@ -447,24 +444,19 @@ public static String replaceStringMulti(String source, String[] match, String re
...
+    int len = source.length();
 
     for (int i = startPos; i < len; i++) {
-      sourceChar = source.charAt(i);
+      char sourceChar = source.charAt(i);
       if (i > lastMatch) {
         sb.append(sourceChar);
       } else {
-        matchCount = 0;
+        int matchCount = 0;
```
KoenDG: I'm suddenly wondering if this move isn't actually bad for performance.

src/main/java/io/ebeaninternal/server/cache/DefaultServerCache.java
```diff
@@ -276,19 +274,24 @@ public void runEviction() {
...
 
-    long idleExpire = System.currentTimeMillis() - (maxIdleSecs * 1000);
-    long ttlExpire = System.currentTimeMillis() - (maxSecsToLive * 1000);
+    // Changing to nanotime as mentioned by Roland Praml on the mailinglist/google group
+    
+    // Using startNanos because: it seems that in this situation, the important thing is when the computation started.
+    // Looking at this code I feel like that should be the point of reference.
+    // Calling nanoTime 2 times, even so close, is going to give different results.
+    long idleExpireNano =  startNanos- TimeUnit.SECONDS.toNanos(maxIdleSecs);
+    long ttlExpireNano = startNanos - TimeUnit.SECONDS.toNanos(maxSecsToLive);
```
KoenDG: This is an actual change to the functionality of the product. @rPraml and @rbygrave plz have a look when you have the time.

src/main/java/io/ebeaninternal/server/core/DatabasePlatformFactory.java
```diff
@@ -134,6 +134,7 @@ private DatabasePlatform byDatabaseMeta(DatabaseMetaData metaData) throws SQLExc
     String dbProductName = metaData.getDatabaseProductName();
     dbProductName = dbProductName.toLowerCase();
 
+    // Isn't being used
     int majorVersion = metaData.getDatabaseMajorVersion();
```
KoenDG: Both Eclipse and Intellij report this field is not being used from any point in the code. Leave it?

src/main/java/io/ebeaninternal/server/core/MultiTenantDbSchemaSupplier.java
```diff
@@ -34,7 +34,7 @@
 
   @Override
   public DataSource getDataSource() {
-    return schemaDataSource;
+    return schemaDataSource; // Potential memory leak: return of inner class prevents outer class from being gc'd
```
KoenDG: As stated by the comment. Of concern or not?

src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
```diff
@@ -736,7 +736,7 @@ public void setBoundId(Object idValue) {
   @Override
   public final void checkRowCount(int rowCount) {
     if (ConcurrencyMode.VERSION == concurrencyMode && rowCount != 1) {
-      String m = Message.msg("persist.conc2", "" + rowCount);
+      String m = Message.msg("persist.conc2", String.valueOf(rowCount));
```
KoenDG: ("" + primitive) is not a performant way to get a string representation of said primitive.

src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
```diff
@@ -1993,6 +1992,10 @@ public void sort(List<T> list, String sortByClause) {
   public ElComparator<T> getElComparator(String propNameOrSortBy) {
     ElComparator<T> c = comparatorCache.get(propNameOrSortBy);
     if (c == null) {
+      // This is the same race condition as in DefaultServer and CQueryPlanStats
+      // If avoiding locking is a priority, be sure to investigate Map.computeIfAbsent source code if this is ever refactored to that.
+      // It's a method introduced in java 8 for this exact "if null -> create -> put" scenario.
+      // The conccurrentHashMap implementation has a few synchronization statements in it.
       c = createComparator(propNameOrSortBy);
```
KoenDG: DefaultServer and CQueryPlanStats both have similar constructs as this and state that while it is a race condition, it should not be worried about. Is that the same case here?

src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
```diff
@@ -2153,7 +2156,7 @@ protected ElPropertyValue buildElGetValue(String propName, ElPropertyChainBuilde
 
     if (propertyDeploy && chain != null) {
       ElPropertyDeploy fk = elDeployCache.get(propName);
-      if (fk != null && fk instanceof BeanFkeyProperty) {
+      if (fk instanceof BeanFkeyProperty) { // instanceof will return false if the variable is null, so a not-null check is redundant.
```
KoenDG: As the comment says: instanceof a null object will always return false. So an extra null check is not necessary.

src/main/java/io/ebeaninternal/server/deploy/DeployDocPropertyOptions.java
```diff
@@ -53,7 +54,7 @@ public void setDocCode(DocCode doc) {
   }
 
   private void setNullValue(String value) {
-    if (!value.equals("")) {
+    if (!value.isEmpty()) {
```
KoenDG: More performant.

src/main/java/io/ebeaninternal/server/deploy/generatedproperty/GeneratedCounterLong.java
```diff
@@ -17,7 +17,7 @@ public GeneratedCounterLong() {
    */
   @Override
   public Object getInsertValue(BeanProperty prop, EntityBean bean, long now) {
-    return (long) 1;
+    return 1L;
```
KoenDG: This seemed unnecessary.

src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationBase.java
```diff
@@ -321,7 +319,9 @@ private static final Method getNullMethod() {
     for (T ann : anns) {
       try {
         if (getPlatformsMethod == null) {
-          getPlatformsMethod = ann.getClass().getMethod("platforms");
+          // Google error prone says: Calling getClass() on an annotation may return a proxy class, use annotationType() instead.
+          // http://errorprone.info/bugpattern/GetClassOnAnnotation
+          getPlatformsMethod = ann.annotationType().getMethod("platforms");
```
KoenDG: Google's "Error Prone" compiler pointed out only 1 error prone thing about ebean's code and this was it.

Details here: http://errorprone.info/bugpattern/GetClassOnAnnotation

src/main/java/io/ebeaninternal/server/deploy/parse/DeployBeanInfo.java
```diff
@@ -13,80 +13,80 @@
...
+  /**
+   * Create with a DeployUtil and BeanDescriptor.
+   */
+  public DeployBeanInfo(DeployUtil util, DeployBeanDescriptor<T> descriptor) {
+    this.util = util;
+    this.descriptor = descriptor;
+  }
+
+  public String toString() {
+    return String.valueOf(descriptor);
```
KoenDG: This is an actual change on top of the formatting, just pointing it out.

src/main/java/io/ebeaninternal/server/expression/SimpleExpression.java
```diff
@@ -36,7 +37,8 @@ public void writeDocQuery(DocQueryContext context) throws IOException {
       String idName = prop.getAssocIdExpression(propName, "");
       Object[] ids = prop.getAssocIdValues((EntityBean) value());
       if (ids == null || ids.length != 1) {
-        throw new IllegalArgumentException("Expecting 1 Id value for " + idName + " but got " + ids);
+        // Regular toString() of a classic array would output something like [Ljava.lang.Object;@1540e19d
+        throw new IllegalArgumentException("Expecting 1 Id value for " + idName + " but got " + Arrays.toString(ids));
```
KoenDG: As the comment states. The Arrays class has its own toString method for this.

src/main/java/io/ebeaninternal/server/loadcontext/DLoadBeanContext.java
```diff
@@ -181,14 +180,8 @@ public void loadBean(EntityBeanIntercept ebi) {
...
-        Iterator<EntityBeanIntercept> iterator = list.iterator();
-        while (iterator.hasNext()) {
-          EntityBeanIntercept batchEbi = iterator.next();
-          if (batchEbi != ebi && context.desc.cacheBeanLoad(batchEbi, persistenceContext)) {
-            // bean successfully loaded from L2 cache so remove from batch load
-            iterator.remove();
-          }
-        }
+        // bean successfully loaded from L2 cache so remove from batch load
+        list.removeIf(batchEbi -> batchEbi != ebi && context.desc.cacheBeanLoad(batchEbi, persistenceContext)); // Intellij java8 suggestion.
```
KoenDG: This is a refactor by Intellij. Java8 has "removeIf" for exactly the given scenario.

src/main/java/io/ebeaninternal/server/persist/dml/DmlBeanPersister.java
```diff
@@ -85,7 +85,7 @@ private int execute(PersistRequestBean<?> request, PersistHandler handler) {
       throw new PersistenceException(msg, e);
 
     } finally {
-      if (!batched && handler != null) {
+      if (!batched) {
```
KoenDG: The handler variable is accessed several times before this point. If the variable was ever null, it would have thrown before this point. Therefore, a nullcheck is pointless. Feel free to check the rest of the code for this, it's not used anywhere else but in this class.

src/main/java/io/ebeaninternal/server/query/CQuery.java
```diff
@@ -317,6 +317,9 @@ private boolean prepareBindExecuteQueryWithOption(boolean forwardOnlyHint) throw
         return false;
       }
 
+      // Call to native method nanoTime() while locked causes an expensive context switch.
+      // startNano is used only in updateExectuionStatistics, which itself is called in various places.
+      // Is this a problem?
       startNano = System.nanoTime();
```
KoenDG: Intellij presented me with this: doing a call on a System method forces a switch of context, which can be expensive. Is this a problem?

The call happen inside a synchronize statement, feel free to look at the file.

src/main/java/io/ebeaninternal/server/query/CQuery.java
```diff
@@ -544,7 +547,7 @@ protected boolean hasNext() throws SQLException {
     return versionList;
   }
 
-  private Version readNextVersion() throws SQLException {
+  private Version<T> readNextVersion() throws SQLException {
```
KoenDG: These 2 were apparently missing. All other methods involved in this change have them and the list that's being filled expects it. So I added it in.

src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
```diff
@@ -387,13 +387,17 @@ private void addPropertyToSubQuery(SqlTreeProperties selectProps, BeanDescriptor
...
         String name = propName.substring(pos + 1);
         p = ((BeanPropertyAssoc<?>) p).getTargetDescriptor().findBeanProperty(name);
       }
     }
 
+    // Above this, there is a check for p == null, but that only logs an error and then doesn't break of return or throw. The code keeps going.
+    // So there's a potential case where this add gets a null value as parameter.
+    // In SqlTreeProperties, this goes to a point where prop.getName() is performed, which will throw a nullpointer if p == null.
+    // Is this intended?
     selectProps.add(p);
```
KoenDG: At the start of this method, p is checked for null. If so, logs an error and doesn't stop. Instead, jumps here. So there's potential for a null variable being passed along here.


What should be done?

src/main/java/io/ebeaninternal/server/text/json/DJsonContext.java
```diff
@@ -45,7 +45,7 @@
 
   private final JsonFactory jsonFactory;
 
-  private final TypeManager typeManager;
+  private final TypeManager typeManager; // Not used anywhere
```
KoenDG: What to do with it?

src/main/java/io/ebeaninternal/server/type/ScalarTypeBigDecimal.java
```diff
@@ -62,7 +62,12 @@ public BigDecimal readData(DataInput dataInput) throws IOException {
     if (!dataInput.readBoolean()) {
       return null;
     } else {
-      return new BigDecimal(dataInput.readDouble());
+      /* From intellij:
+      Reports on calls to BigDecimal constructors which accept a double value.
+      These constructors can have somewhat unpredictable results because many numbers cannot be represented exactly in a double.
+      It is recommended to use the BigDecimal.valueOf() method or the constructors which accept a String instead.
+       */
+      return BigDecimal.valueOf(dataInput.readDouble());
```
KoenDG: See comment in code. Should something be done about it, or is it actually okay?

src/test/java/org/tests/cache/TestCacheCollectionIds.java
```diff
@@ -79,7 +79,7 @@ private int fetchCustomer(Integer id) {
     Customer customer2 = Ebean.find(Customer.class, id);
 
     List<Contact> contacts2 = customer2.getContacts();
-    contacts2.size();
+
```
KoenDG: Seems pointless.

src/test/java/org/tests/model/softdelete/ESoftDelDown.java
```diff
@@ -8,7 +8,7 @@
   String down;
 
   public ESoftDelDown(String up) {
-    this.down = down;
+    this.down = up;
```
KoenDG: This and the one below this didn't seem right.

src/test/java/org/tests/timezone/TimezoneTests.java
```diff
@@ -70,6 +70,7 @@ private void setZone(String zone) {
     TimeZone.setDefault(TimeZone.getTimeZone(zone));
   }
 
+  // Private method that is never used anywhere.
   private void insert(String zone) throws SQLException {
```
KoenDG: As the comment states. Drop it?

src/main/java/io/ebean/Model.java
rbygrave: Correct.  Nothing to worry about.

src/main/java/io/ebean/config/JsonConfig.java
rbygrave: That is a public API change.  So has to be treated in separate PR.

src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
rbygrave: Best to NOT autoformat the entire project - leave that to me (as it makes your PR bigger than it needs to be to get 'effective changes' in).

src/main/java/io/ebeaninternal/server/cache/DefaultServerCache.java
rbygrave: Yup, effectively this means the whole PR can't go in.

src/main/java/io/ebeaninternal/server/core/MultiTenantDbSchemaSupplier.java
rbygrave: Not.

src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
rbygrave: It doesn't matter here ... but there are places we can use computeIfAbsent().  

What I would say is doing multiple smaller PR's is **_much_** better for me. More focus, better turn around, less chance of missing something.

src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorManager.java
```diff
@@ -567,7 +567,7 @@ private void initialiseAll() {
 
   private void checkForValidEmbeddedId(BeanDescriptor<?> d) {
     IdBinder idBinder = d.getIdBinder();
-    if (idBinder != null && idBinder instanceof IdBinderEmbedded) {
+    if (idBinder instanceof IdBinderEmbedded) { // If variable is null, will result in false
```
rbygrave: Actually adding in the comment // If variable ...   means I can't review this once.  You should leave out all comments like that from the code and put them in the PR instead if you like.

rbygrave:
Right. I got about a 1/3 way through.  What I'd like to see you do is do a minimal PR with just the changes you are 100% confident about and is really to merge (so no comments in the code that need to be removed etc, ready to merge).  Lets get that submitted and I'll review.  Don't put anything in you are not 100% confident on ... those can go into another PR later.

That is, lets aim for 10 PR's that are much smaller and focused (one at a time). 1 massive PR will be too time consuming and never actually get merged.  Focus on the things we are 100% confident on and get those reviewed and merged and go from there.

src/main/java/io/ebeaninternal/server/query/CQuery.java
0xbaadf00d: The usage of `System.currentTimeMillis()` is it possible ? Yes you always have the context switching, but `currentTimeMillis` is faster than `nanoTime`.

Basically, `System.currentTimeMillis()` just reads a global variable maintained by Windows (which is what it has low granularity), whereas `System.nanoTime()` actually has to do IO operations.

src/main/java/io/ebean/config/dbplatform/oracle/OracleDbEncrypt.java
KoenDG: Alright, will start reverting that.

src/main/java/io/ebeaninternal/server/cache/DefaultServerCache.java
KoenDG: I'll get a diff file out for it and make a new branch with it.

KoenDG:
Closing now as suggested by Rob. Will be making separate branches/PRs in the future.

PR #974, opened 2017-02-22
FIX: do not hit database if inherited entity has no children

rPraml:
I had the problem when creating reference bean that every time the DB was hit. In my opinion it is not neccesary to do that if the bean class has no children.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -589,7 +589,7 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
     }
 
     InheritInfo inheritInfo = desc.getInheritInfo();
-    if (inheritInfo == null) {
+    if (inheritInfo == null || inheritInfo.getChildren().isEmpty()) {
       return (T) desc.contextRef(pc, null, false, id);
     }
 
--- a/src/main/java/io/ebeaninternal/server/deploy/InheritInfo.java
+++ b/src/main/java/io/ebeaninternal/server/deploy/InheritInfo.java
@@ -165,7 +165,14 @@ public class InheritInfo {
   public BeanProperty[] localProperties() {
     return descriptor.propertiesLocal();
   }
-
+  
+  /**
+   * Return the children.
+   */
+  public ArrayList<InheritInfo> getChildren() {
+    return children;
+  }
+  
   /**
    * Get the bean property additionally looking in the sub types.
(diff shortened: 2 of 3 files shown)


rbygrave:
What is the scenario / testcase that means a BeanDescriptor has a inheritInfo not null but has empty children ?

The point is we want to really fix the underlying issue (and this proposed change does not really do that).

rPraml:
Hello Rob, I have the following entities


```
// Common super class:
@Entity
@MappedSuperclass
@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
@DiscriminatorColumn(name = "kind")
public abstract class AclEntry { }

// subclass 1
@Entity
@DiscriminatorValue("U")
public class User extends AclEntry { }

// subclass 2
@Entity
@DiscriminatorValue("U")
public class User extends AclEntry { }
```

So, `Ebean.getReference(Group.class, id)` hits the DB which is not neccessary, because the only valid result is to return a "Group" with the given id.
It would be OK that `Ebean.getReference(AclEntry.class, id)` hits the DB (as it is also not possible to return an AclEntry as this class is abstract in this example)

Still open question (but not yet part of my use case):
I don't now what should happen if I have an additional inheritance layer:
```
@Entity
@DiscriminatorValue("S")
public class SuperUser extends User { }
```

It is clear that `Ebean.getReference(SuperUser .class, id)` does not need to hit the DB.
In my scenario it would be OK that `Ebean.getReference(User .class, id)` returns an instance of User and does not hit the DB (although the stored bean in the DB is of type SuperUser)

Concrete: I have to import a lot of data (millions of rows) that has a ManyToOne-reference to various `AclEntry`-entities.
I know the ID of the user - the import process is speed up if Ebean.getReference does not hit the DB

rbygrave:
> It is clear that Ebean.getReference(SuperUser .class, id) does not need to hit the DB.

Right. I can create a test case for that (asserting that no sql is executed) so I'll do that.

> I know the ID of the user - the import process is speed up if Ebean.getReference does not hit the DB

As a workaround for that scenario ... for Ebean a "new entity bean with only the Id property populated" is treated as a reference bean when related beans are saved.  That is, instead of using Ebean.getReference ... create a new entity bean (User) and only set the id property and use that instead of Ebean.getReference (and Ebean will treat it like a reference bean).  In the above we can new up the concrete User bean and get the desired behavior.

rPraml:
FYI: I already have added a small test case

rbygrave:
Thanks, merging.

PR #976, opened 2017-02-22
Simple fixes

KoenDG:
New PR based on the old one. This one is only the really simple stuff. I made a list of the things that I need more info on or that have to be their own PR.

I'll wait out this one before I start on the others.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
+++ b/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
@@ -377,11 +377,11 @@ class CQueryBuilder {
       PreparedStatement statement = connection.prepareStatement(sql);
       predicates.bind(statement, connection);
 
-      List<String> propertyNames = new ArrayList<>();
-
       ResultSet resultSet = statement.executeQuery();
       ResultSetMetaData metaData = resultSet.getMetaData();
+
       int cols = 1 + metaData.getColumnCount();
+      List<String> propertyNames = new ArrayList<>(cols - 1);
       for (int i = 1; i < cols; i++) {
         String tableName = metaData.getTableName(i).toLowerCase();
         String columnName = metaData.getColumnName(i).toLowerCase();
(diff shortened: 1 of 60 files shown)


src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
```diff
@@ -377,11 +377,11 @@ private SqlTree createNativeSqlTree(OrmQueryRequest<?> request, CQueryPredicates
...
       predicates.bind(statement, connection);
 
-      List<String> propertyNames = new ArrayList<>();
-
       ResultSet resultSet = statement.executeQuery();
       ResultSetMetaData metaData = resultSet.getMetaData();
+
       int cols = 1 + metaData.getColumnCount();
+      List<String> propertyNames = new ArrayList<>(cols - 1);
       for (int i = 1; i < cols; i++) {
```
rPraml: @rbygrave I also noticed this method when debugging and was wondering why you add +1 to cols. I would suggest to change the comparison in the for loop from "<" to "<=" and remove all the +1's and -1's.

src/test/java/org/tests/cache/TestCacheCollectionIds.java
```diff
@@ -79,7 +79,7 @@ private int fetchCustomer(Integer id) {
     Customer customer2 = Ebean.find(Customer.class, id);
 
     List<Contact> contacts2 = customer2.getContacts();
-    contacts2.size();
```
rPraml: This could have side effects (but shouldn't) as "BeanList.size()" will trigger a lazy load. But BeanList.iterator() does the same.

rPraml:
looks OK for me

src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
rbygrave: JDBC API is 1 based (and not 0 based) hence the +1

src/test/java/org/tests/cache/TestCacheCollectionIds.java
rbygrave: Yes, this will invoke/exercise lazy loading so I might put it back in.

rbygrave:
Good work - thank you !!

PR #980, opened 2017-02-23
fix several bugs in findSingleAttibute

rPraml:
This pull request fixes several bugs in "findSingleAttribute" - mostly in conjunction with distinct
- on inherited entities
- on id column
- does not work with fetch properties

### DIFF
--- a/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
+++ b/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
@@ -438,7 +438,7 @@ public class SqlTreeBuilder {
         p = desc.findBeanProperty("id");
         selectProps.add(p);
 
-      } else if (p.isId()) {
+      } else if (p.isId() && (query == null || !query.isDistinct())) {
         // do not bother to include id for normal queries as the
         // id is always added (except for subQueries)
 
--- a/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
+++ b/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
@@ -322,7 +322,9 @@ public class SqlTreeBuilder {
       return new SqlTreeNodeManyRoot(prefix, (BeanPropertyAssocMany<?>) prop, props, myList, temporalMode, disableLazyLoad);
 
     } else {
-      return new SqlTreeNodeBean(prefix, prop, props, myList, temporalMode, disableLazyLoad);
+      // do not read Id on child beans (e.g. when used with fetch())
+      boolean withId = (query == null || !query.isDistinct()); 
+      return new SqlTreeNodeBean(prefix, prop, props, myList, withId, temporalMode, disableLazyLoad);
     }
   }
 
--- a/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
+++ b/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
@@ -323,7 +323,7 @@ public class SqlTreeBuilder {
 
     } else {
       // do not read Id on child beans (e.g. when used with fetch())
-      boolean withId = (query == null || !query.isDistinct()); 
+      boolean withId = (query == null || !query.isSingleAttribute()); 
       return new SqlTreeNodeBean(prefix, prop, props, myList, withId, temporalMode, disableLazyLoad);
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
```diff
@@ -438,7 +440,7 @@ private void addProperty(SqlTreeProperties selectProps, BeanDescriptor<?> desc,
         p = desc.findBeanProperty("id");
         selectProps.add(p);
 
-      } else if (p.isId()) {
+      } else if (p.isId() && (query == null || !query.isSingleAttribute())) {
         // do not bother to include id for normal queries as the
         // id is always added (except for subQueries)
```
rPraml: we must not skip id properties in singleAttribute queries - because it is the only one.

src/main/java/io/ebeaninternal/server/query/SqlTreeNodeBean.java
```diff
@@ -135,6 +135,11 @@ private SqlTreeNodeBean(String prefix, BeanPropertyAssoc<?> beanProp, BeanDescri
 
   @Override
   public BeanProperty getSingleProperty() {
+    if (properties == null || properties.length == 0) {
+      // if we have no property ask first children (in a distinct select with join)
+      // if we have also no children, NPE happens anyway.
+      return children[0].getSingleProperty();
+    }
     return properties[0];
```
rPraml: if we want a single property, we use the first one we can find - while writing this comment, it could make sense to iterate over all children until we find one.

src/main/java/io/ebeaninternal/server/query/SqlTreeNodeBean.java
```diff
@@ -434,11 +439,11 @@ public void appendSelect(DbSqlContext ctx, boolean subQuery) {
...
 
-    if (!subQuery && inheritInfo != null) {
-      ctx.appendColumn(inheritInfo.getDiscriminatorColumn());
-    }
-
     if (readId) {
+      if (!subQuery && inheritInfo != null) {
+        ctx.appendColumn(inheritInfo.getDiscriminatorColumn());
+      }
+
```
rPraml: only read discriminator when also read id... (see * )

src/main/java/io/ebeaninternal/server/query/SqlTreeNodeBean.java
```diff
@@ -162,9 +167,9 @@ private String getPath(String prefix, String propertyName) {
   @Override
   public void buildRawSqlSelectChain(List<String> selectChain) {
     if (readId) {
-      if (desc.hasInheritance()) {
+      if (inheritInfo != null) {
         // discriminator column always proceeds id column
-        selectChain.add(getPath(prefix, desc.getInheritInfo().getDiscriminatorColumn()));
+        selectChain.add(getPath(prefix, inheritInfo.getDiscriminatorColumn()));
       }
```
rPraml: ... *) as it is also done here

src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
```diff
@@ -1220,6 +1221,10 @@ public void findEach(Consumer<T> consumer) {
   @Override
   @SuppressWarnings("unchecked")
   public <A> List<A> findSingleAttributeList() {
+    if (!detail.hasSelectClause()) {
+      // (no explicit select set - clear all properties)
+      detail.setBase(new OrmQueryProperties(null, new LinkedHashSet<>()));
+    }
     return (List<A>) server.findSingleAttributeList(this, null);
```
rPraml: if no explicit select clause is set, ebean selects all properties somewhere later - which messes up the single-attribute-query when you want explicitly **fetch**  a property deeper in the object graph. This is not nice, but this is the minimal invasive solution.

rPraml:
Hello Rob, hopefully my last PR for this week ;)

src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
rbygrave: This does not look right the correct place for this if ... setBase() code .... in that people can call server.findSingleAttributeList(Query, Transaction) and this check should also support that alternative way to execute the query. 

I'll look to move this check after merge.

rbygrave:
Thanks, merging - I'll make a little tweak after merge.

PR #981, opened 2017-02-24
use composed key in multi tenant environment instead a cache per tenant

rPraml:
Hello Rob,

please review the changes in this PR
 
**old behavour**
- there is a cache per table and tenant
- use case: clustered environment: Caches will not invalidate properly, becacuse clusterpartner does not know the tenant id and does not know which of the table caches to invalidate
- Ebean.externalModification() works for currentTenant only

**new behaviour**
- there is a cache per table - they are using a compound key (tenantId, entitiyId)
- a TableIUD-event will therefore clear the cache for all tenants
- Ebean.externalModification() will do also.

### DIFF
--- a/src/main/java/io/ebean/cache/ServerCacheFactory.java
+++ b/src/main/java/io/ebean/cache/ServerCacheFactory.java
@@ -1,5 +1,7 @@
 package io.ebean.cache;
 
+import io.ebean.config.CurrentTenantProvider;
+
 /**
  * Defines method for constructing caches for beans and queries.
  */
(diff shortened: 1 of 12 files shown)


rbygrave:
Had a first review and this looks like a good change.  I'll review again later.

rbygrave:
Note that this is really just waiting for me to have a window of time ... (to additionally update the hazelcast and ignite l2 cache adapters)

rPraml:
If you are continuing with this issue, please take a look at this commit:
https://github.com/FOCONIS/ebean/commit/ddf0859fca79b6eff396685140c73791e32c2ddc
If you switch tenant during a transaction you might get the wrong bean from PC

**Use Case:**
- We have to import a CSV-file, e.g. first column contains tenant id
- The whole import is surrounded by a Transaction (import all or nothing)
- for each CSV-line we may have to switch the tenant
- we perform also updates where we discovered the bug.

rbygrave:
Yes, I am looking to merge this in.  Just looking for a window to do it as I'll need to update the L2 implementations (hazelcast and ignite).

> for each CSV-line we may have to switch the tenant

Interesting use case. Hmmm.

PR #982, opened 2017-02-24
Fix multi tenant bean change

rPraml:
Hello Rob,

added the tenant id to BeanChange - as BeanChanges are logged async - and currentTenantProvider is most likely a ThreadLocal, it is neccessary to store the tenantId of the originating thread.


rbygrave:
Yes, good - thanks !

PR #986, opened 2017-03-04
Warning suppressions.

KoenDG:
Went over compilation warnings with Xlint. Only imports for DateMidnight are left, as they cannot receive a suppression annotation.

DateMidnight has been deprecated. The source code states:

```
 * @deprecated The time of midnight does not exist in some time zones
 * where the daylight saving time forward shift skips the midnight hour.
 * Use {@link LocalDate} to represent a date without a time zone.
 * Or use {@link DateTime} to represent a full date and time, perhaps
 * using {@link DateTime#withTimeAtStartOfDay()} to get an instant at the
 * start of a day.
```

I doubt it can be migrated, since user projects depending on it, though deprecated, will continue to need the support for it. That's the kind of change that can break projects if not announced.

So I just added suppression statements.

### DIFF
--- a/src/main/java/io/ebean/text/json/EJson.java
+++ b/src/main/java/io/ebean/text/json/EJson.java
@@ -123,6 +123,7 @@ public class EJson {
   /**
    * Parse the json returning as a List taking into account the current token.
    */
+  @SuppressWarnings("unchecked")
   public static List<Object> parseList(JsonParser parser, JsonToken currentToken) throws IOException {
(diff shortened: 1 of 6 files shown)


rPraml:
You can remove the Import and use the full qualified class name in the @deprecated methods. Depending on how often DateMidnight is used, this may make sense.

KoenDG:
Gah, of course. Pushed the change.

rbygrave:
Great, thanks !

PR #987, opened 2017-03-05
Added multi tenant catalog support.

alexanderwink:
Basically this is the same as the schema provider but instead of changing the schema it changes the catalog. This works well for db engines that doesn't support setting the schema but support setting the catalog (like MySQL).


rbygrave:
Good work, thanks ... merging.

Issue #909, opened 2016-12-02
@Lob with @Size ... produces incorrect DDL for Postgres


rbygrave:
### Steps to reproduce

```java
  @Size(max = 2000)
  @Lob
  String note;
```

PR #988, opened 2017-03-06
Wrong alter table / alter column DDL was generated

rPraml:
Hello Rob,

this PR fixes wrong DDL generation for postgres:

*before*
```SQL
alter table my_table alter column my_column type clob;
```
=> clob not supported by postgres (related to #909)
*now*
```SQL
alter table my_table alter column my_column type text;
```

I think it was also wrong for sqlserver & mysql - the unit test for oracle seems to be wrong also.

cheers
Roland

### DIFF
--- a/src/test/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl_AlterColumnTest.java
+++ b/src/test/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl_AlterColumnTest.java
@@ -68,7 +68,7 @@ public class PlatformDdl_AlterColumnTest {
     assertEquals("alter table mytab alter column acol type varchar(20)", sql);
 
     sql = oraDdl.alterColumnType("mytab", "acol", "varchar(20)");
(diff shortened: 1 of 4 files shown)


src/test/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl_AlterColumnTest.java
```diff
@@ -68,7 +68,7 @@ public void testAlterColumnType() throws Exception {
     assertEquals("alter table mytab alter column acol type varchar(20)", sql);
 
     sql = oraDdl.alterColumnType("mytab", "acol", "varchar(20)");
-    assertEquals("alter table mytab modify acol varchar(20)", sql);
+    assertEquals("alter table mytab modify acol varchar2(20)", sql);
```
rPraml: rob, please take a look at this change - my change did break this test - although I think varchar2 is correct here.

rbygrave:
Looks all good - thank you !!

PR #990, opened 2017-03-10
Wrong cast to query and childid propblem fix

PeterCsubak:
Hi Rob,

In our project we use the `@SoftDelete` feature a lot, and I found some bugs. This PR contains fixes for two easy ones. 

Please merge this, so we don't need to use the forked one.

Thanks in adwance!

Best regards,
Peter


rbygrave:
Thanks.  Ideally we create test cases for any fixes (to ensure we have coverage going forward).  If we get a chance to describe the scenarios or even provide a test case that would be really great.

I'll merge this - thanks.

PR #998, opened 2017-03-18
Moving defaultservercache implementation to use nanoseconds

KoenDG:
As suggested previously by @rPraml the use of milliseconds to manage cache evictions could potentially be problematic. For instance, during DST jumps.

The PR changes the implementation to use nanoseconds instead.


rbygrave:
Great thanks.

PR #1006, opened 2017-04-15
Small changes for ArrayStack.

KoenDG:
Alright, second attempt after cleaning up my git config.

Going over some other stuff, I noticed this suppressed inspection and also the method signature of the `contains` method. Since the class is parameterized, I figured the parameter should also be used in the contains method.

Also that method that only gets used by the class could be private instead of protected.

It's used by WriteJson.java and a unit test.


rbygrave:
Thanks.

PR #1007, opened 2017-04-15
Some changes that are possible in Java language level 7 and 8.

KoenDG:
With Java7, it is not required to detail the class on the right hand side of a generic statement.

And with Java8, sorting can now also be done on the object itself, rather than using the static method Collections.sort


rbygrave:
Thanks again :)

Issue #1012, opened 2017-04-24
Filtering by embedded value in a joined table does not create sql join


OndrejPacay:
## Expected behavior
If there are 2 models related models A and B while B is containing some embedded value and we are selecting on A and filtering on B.embedded = X then a sql join should be created.
## Actual behavior
No SQL join is created and the query fails with javax.persistence.PersistenceException
### Steps to reproduce
See the test case in the pull request.
```java
ExpressionList<EEmbInner> el = Ebean.find(EEmbInner.class).where()
                .eq("outer.datePeriod.date1", new Date(11111));
        List<EEmbInner> list = el.findList();
```

```console
javax.persistence.PersistenceException: Query threw SQLException:Column "T1.DATE1" not found; SQL statement:
select t0.id, t0.nome_inner, t0.update_count, t0.outer_id from eemb_inner t0 where t1.date1 = ? 
```

OndrejPacay:
Hi @rbygrave , 
I have a fix for this issue, is it possible that I can push it to a feature branch and create a pull request?

Thanks, Ondrej

rbygrave:
Yes, by all means prepare a pull request.

Thanks, Rob.

PR #1013, opened 2017-04-24
#1012 - Filtering on embedded values in a joined table does not create a sql join

OndrejPacay:
Please see issue #1012


rbygrave:
Good stuff, thanks.

PR #1015, opened 2017-04-27
Fix/select distinct cascaded fetch with findSingleAttributeList() on associated property path

rPraml:
Hello Rob,

today we discovered an issue with non working "fetch single attribute" when a child attribute was queried.

It was not trivial to fix this, because "markQueryJoins" adds all missing IDs, so I added a flag if only empty OrmQueryProperties should be added.
Maybe you take a look at https://github.com/ebean-orm/ebean/pull/980 again and find a more common solution for this.

cheers
Roland


rbygrave:
I think this is reasonable enough for now.  I'll bring this PR in.

PR #1016, opened 2017-04-28
FIX BUG：and blank space after 1=1 (Noop expression for equals null with @SoftDelete)

yuanxuegui:
If add ebean.expressionEqualsWithNullAsNoop=true in ebean.properties，and use findCount、findList，it will generate sql like:
 where t0.username = ?  and 1=1and t0.is_deleted = 0 ,
so query will threw SQLException.


rbygrave:
This breaks the test NoopExpressionTest.test_withPreAndPost(). I'll fix that.  Generally it is good if PR's ensure they also fix associated tests.

yuanxuegui:
OK

PR #1017, opened 2017-04-29
Once again, Static code analysis

KoenDG:
Hello again @rbygrave @rPraml and all others.

Here's a few static code analysis recommendations, mostly focussed on performance and language level options.

This PR is mostly focussed on lists which get instantiated with default size and then `addAll` is performed on them. Most Collection/subclass of Collection constructors have a variant that takes an existing Collection. That's one step less.

Some `for` constructs that only copy an existing list, replaced by `addAll` where possible.

Some unused imports, removed.

A single `computeIfAbsent` option was found.

2 inner classes made static, which causes them to not get internally linked to the outer class, which is sort of pointless if the inner class holds no actual reference to a field/method of the outer class. But the JVM does it anyway unless you make it static, or so I've been lead to believe.

Also, I found a manually written comparison of primitive arrays, one per type. `Arrays.equals` performs exactly this functionality. I had a look at the source code and it's even 2 steps shorter than the code manually written here. Clever use of code constructs in the JDK implementation.

It's a bit of a mixed bag, so let me know if you want me to separate this up a bit.

Next up: Intellij detected 18 cases of `tail recursion` that can be replaced. But that's for after this PR is resolved.

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -3,6 +3,7 @@ package io.ebean.bean;
 import io.ebean.Ebean;
 import io.ebean.ValuePair;
 
+import java.util.Arrays;
 import javax.persistence.EntityNotFoundException;
 import javax.persistence.PersistenceException;
 import java.beans.PropertyChangeEvent;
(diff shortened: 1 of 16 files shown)


rbygrave:
Thanks.

PR #1019, opened 2017-04-30
Replaced Tail Recursion while while/true loops.

KoenDG:
Hello everyone.

So as I stated in my previous PR, this one is all about replacing tail recursion with while/true loops.

The point being avoiding doing more method calls and stepping deeper into a list of returns.

The behavior of calling the method again as the last step of a branch is exactly the same idea as doing a while/true loop: as long as the condition you want is not met, you keep going. In the old case, that means calling the method again with an updated value, in this new case, it means doing the loop again with an updated value.

Take some time to carefully analyze this. They all retain their exit conditions in the exact same way as the old code.


rbygrave:
Good work, thanks.

PR #1022, opened 2017-05-10
Fix: Bug when calculating order by clause with primary keys and distinct queries

AlexWagner:
When using `select distinct col_name ... order by col_name` col_name in select was t1.id and col_name in order by was t0.parent_id


rbygrave:
This is good work but I think we want the best sql in this case which means we want to allow the foreign key's to ... avoid the extra join and use the foreign key columns.

I'll log another ticket taking the test cases you have, adjust them a little and fix with what I consider the better sql (avoiding the extra join).

See https://github.com/ebean-orm/ebean/issues/1023

Cheers, Rob.

rbygrave:
I think we should look to close this now ... As fixed by #1023

rbygrave:
Hi Alex, are you happy to close (given we have fixed as per #1023 and extra tests as per https://github.com/ebean-orm/ebean/pull/1024)

AlexWagner:
Hi Rob, yes of course

Issue #1023, opened 2017-05-11
Bug with findSingleAttributeList() when used against @ManyToOne / foreign key


rbygrave:
### Steps to reproduce

```java
List<Integer> customerIds = Ebean.find(Contact.class)
        .setDistinct(true)
        .select("customer")
        .orderBy().desc("customer")
        .findSingleAttributeList();
```
### expected sql
```sql
select distinct t0.customer_id from contact t0 order by t0.customer_id desc
```

## Example including a join

```java
List<Integer> customerIds = Ebean.find(Contact.class)
        .setDistinct(true)
        .fetch("customer","billingAddress")
        .orderBy().desc("customer.billingAddress");
        .findSingleAttributeList();
```
```sql
select distinct t1.billing_address_id 
  from contact t0 
  join o_customer t1 on t1.id = t0.customer_id  
  order by t1.billing_address_id desc
```

PR #1024, opened 2017-05-12
ADD: Testcases for "fetchDistinct"

rPraml:
Hello Rob, please take a second look at issue #1023 

I've added 5 test cases that will still fail. I'll try to provide a fix in different branch

### DIFF
--- a/src/test/java/org/tests/query/other/TestQuerySingleAttribute.java
+++ b/src/test/java/org/tests/query/other/TestQuerySingleAttribute.java
@@ -329,45 +329,7 @@ public class TestQuerySingleAttribute extends BaseTestCase {
 
     assertThat(sqlOf(query)).contains("select distinct t1.billing_address_id from contact t0 join o_customer t1 on t1.id = t0.customer_id  order by t1.billing_address_id desc");
   }
-
-  @Test 
-  public void distinctWithOrderByPk2() { 
- 
-    ResetBasicData.reset(); 
- 
-    Query<Contact> query = Ebean.find(Contact.class) 
-        .setDistinct(true) 
-        .fetch("customer","id") 
-        .orderBy().desc("customer.id"); 
- 
-    List<Integer> ids = query.findSingleAttributeList(); 
-    
-    // Note: this is not yet the best optimized query.
-    // The join is not neccessary if it is a foreign key join
-    // (if it is a left join, result would be different!)
-    assertThat(sqlOf(query)).contains("select distinct t1.id "
-        + "from contact t0 join o_customer t1 on t1.id = t0.customer_id  "
-        + "order by t1.id");
-  } 
    
-  @Test 
-  public void distinctWithCascadedFetchOrderByPk2() { 
- 
-    ResetBasicData.reset(); 
- 
-    Query<Contact> query = Ebean.find(Contact.class) 
-        .setDistinct(true) 
-        .fetch("customer.billingAddress","id") 
-        .orderBy().desc("customer.billingAddress.id"); 
- 
-    List<Short> ids = query.findSingleAttributeList(); 
- 
-    // Note: This is also not the best optimized query
-    assertThat(sqlOf(query)).contains("select distinct t2.id from contact t0 "
-        + "join o_customer t1 on t1.id = t0.customer_id  " 
-        + "left join o_address t2 on t2.id = t1.billing_address_id  order by t2.id"); 
-  }
-  
   @Test 
   public void distinctWithOrderByPkAndQuery() { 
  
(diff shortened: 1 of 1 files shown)


rPraml:
We have built upon ebean an application that shows complex queries in a grid. We give that component the labels & el-paths that we want to show.

For a subset of your "Contact" entity the input data could be:

Header label  |  EL Path 
---------| -------
First Name | firstName
Last Name | lastName
Customer | customer.name
Billing City | customer.billingAddress.city
Billing Line1 | customer.billingAddress.line1
Shipping | customer.shippingAddress.city

Now we build programmatically a query like this:
```java
baseQuery = Ebean.find(Contact.class).where() ...  .query(); // this subset of data should be shown
query = baseQuery.copy()
        .select("firstName lastName")
        .fetch("customer", "name")
        .fetch("customer.billingAddress", "city name1")
        .fetch("customer.shippingAddress", "city"); // these columns are required
```
and pass it to our component.
Now we get a nice 'Excel-like' sheet similar to this

First Name | LastName | Customer | Billing City | Billing Line 1| Shipping 
------------|-----------|-----------|------------|--------------|---------------
Roland | P. | RP | Ti.... | B.... | V....
...|...|...|...|...|...
...|...|...|...|...|...
...|...|...|...|...|...

(Note: It was a bit cumbersome to distribute the el path to 'fetch()' & 'select()', so we added a 'fetchProperties', see here: https://github.com/FOCONIS/ebean/blob/master/src/main/java/io/ebeaninternal/server/querydefn/OrmQueryDetail.java#L569)

Now there is an **extra-featue**, a column filter as you might know from excel. A column can have a small drop down/ComboBox where it shows all distinct values of this column:

First Name [v] |  ...
---------------| -----------
... | ...

To get the distinct values, we reuse the `baseQuery` from above and perform a fetchSingleAttribute() for the EL-path of the according column.

Till now, everything works fine, if we do not have ID-columns in the grid.

> I consider all these tests as effectively invalid (for what we are trying to achieve SQL wise).

In my opinion, all tests do a valid ebean query (the produced SQL is not the best, yes)

I understand your point of view, that I also can use "customer.billingAddress" instead of "customer.billingAddress.id", but this means, that I get a `List<Address>`  instead of a `List<Integer>`.
This adds additional complexity to my code (have to detect that edge case & map `List<Address>` -> `List<Integer>` ) and memory footprint/overhead would also be bigger.
And I really need a List of Integers for my combobox ;-)

The patch from Alexander Wagner would at least fix our problem (but breaks your two new tests)

Maybe I've an idea now, how to fix this.

rbygrave:
*> I understand your point of view, that I also can use
"customer.billingAddress" instead of "customer.billingAddress.id
<http://customer.billingAddress.id>", but this means, that I get a
List<Address> instead of a List<Integer>.*

No, they don't, the tests I put up all returned the expected *List<Integer>
*... result.

The above is not accurate enough, you need to be really accurate here.  I
did not put any tests up using "customer.billingAddress" but instead
fetch("customer",
"billingAddress").

The patch from Alexander Wagner was not appropriate in many many cases.



*> Maybe I've an idea now, how to fix this.*

I'd rather you understood the fix I have already pushed first.  It is not
clear you understand you get List<Integer> from those queries (but yes you
do and we also get the best sql).

rbygrave:
So yes these extra tests all look good.  

Ideally Ebean would be smart enough in these findSingleAttributeList() cases to know when to automatically change/convert from the likes of `fetch("customer.billingAddress", "id")` to `fetch("customer", "billingAddress")`.  I realise this is not going to be obvious to people generally.  However, we can get the right SQL and right result so at this stage we just need to write the queries this way but maybe that isn't a bad thing.

rPraml:
> I'd rather you understood the fix I have already pushed first. 
Thanks Rob, I think I got it - query.select/fetch has less common with EL-paths. As told before I've added a fetchProperties method. Handling the special case for ID properties fixes the problem.

Do you think a fetchProperties (or fetchElPaths might be a better name) could be useful in ebean standard? Then I would make the effort to create a pull request.
The idea behind this concept was, to tell ebean all the required el-paths of the object graph that are required to display the data in a grid-table without lazy-load/n+1 problem.

BTW: the code is here https://github.com/FOCONIS/ebean/blob/861b9f8c7d1681d3f3379cadd486ea23467da0da/src/main/java/io/ebeaninternal/server/querydefn/OrmQueryDetail.java#L593
It would be great if you take a look on this piece of code, I know It is rather complex (it iterates over all EL paths and build the apropriate select/fetch statements) Maybe you have a hint how to do this easier.

Roland

rbygrave:
> Do you think a fetchProperties (or fetchElPaths might be a better name)
could be useful in ebean standard?

I think it already exists. Depends on exactly what you mean but there is
already a mechanism to define/parse the "fetch graph".

PR #1028, opened 2017-05-20
For debug and trace levels, either added `isEnabled` check or made arguments parameterized.

KoenDG:
As the title states.

Also, there's this regex that was using alternation when the only purpose was to create a logical `OR` construct, which is what character classes are for.

For more info on character clas vs alternation for logical `OR` constructs: http://stackoverflow.com/a/22133387

And the answer directly below it.


rbygrave:
Thanks

PR #1030, opened 2017-05-20
Structure introduced to indicate that a comparison of objects is indeed between object reference and not object content.

KoenDG:
Before anything else, don't feel like reviewing this should take priority over other work, it's entirely a pet peeve of mine when looking at `==` comparisons.

=========

Because comparing objects with `==` can be confusing, I created a class `JHelper` with 1 static method that has a very specific name, indiciating that the point of the method is to compare the object reference of 2 objects for equality and not the actual contents of the objects.

This PR should initially be used to go over these cases and verify that this was indeed the case for all affected instances.

Readability is the point, basically. When reading that point in the code, people reading it should know "this is what's intended with this check", instead of doing a quick check to see whether or not a value comparison was actually supposed to be done.

Also, not sure on proper location for this class, plz advise.

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -1,6 +1,5 @@
 package io.ebean.bean;
 
-import io.JHelper;
 import io.ebean.Ebean;
 import io.ebean.ValuePair;
 
@@ -833,7 +832,7 @@ public final class EntityBeanIntercept implements Serializable {
     if (obj2 == null) {
       return false;
     }
-    if (JHelper.objectSameReference(obj1, obj2)) {
+    if (obj1 == obj2) {
       return true;
     }
     if (obj1 instanceof BigDecimal) {
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -1,5 +1,6 @@
 package io.ebean.bean;
 
+import io.JHelper;
 import io.ebean.Ebean;
 import io.ebean.ValuePair;
 
@@ -832,7 +833,7 @@ public final class EntityBeanIntercept implements Serializable {
     if (obj2 == null) {
       return false;
     }
-    if (obj1 == obj2) {
+    if (JHelper.objectSameReference(obj1, obj2)) {
       return true;
     }
     if (obj1 instanceof BigDecimal) {
--- a/src/main/java/io/ebean/common/BeanList.java
+++ b/src/main/java/io/ebean/common/BeanList.java
@@ -1,5 +1,6 @@
 package io.ebean.common;
 
+import io.JHelper;
 import io.ebean.bean.BeanCollection;
 import io.ebean.bean.BeanCollectionAdd;
 import io.ebean.bean.BeanCollectionLoader;
@@ -97,7 +98,7 @@ public final class BeanList<E> extends AbstractBeanCollection<E> implements List
    */
   private boolean containsInstance(Object bean) {
     for (Object element : list) {
-      if (element == bean) {
+      if (JHelper.objectSameReference(element, bean)) {
         return true;
       }
     }
--- a/src/main/java/io/ebean/common/BeanList.java
+++ b/src/main/java/io/ebean/common/BeanList.java
@@ -1,6 +1,5 @@
 package io.ebean.common;
 
-import io.JHelper;
 import io.ebean.bean.BeanCollection;
 import io.ebean.bean.BeanCollectionAdd;
 import io.ebean.bean.BeanCollectionLoader;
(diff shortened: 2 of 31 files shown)


0xbaadf00d:
For me is not an enhancement and add a lot of call, so add an overhead.

rPraml:
As far as I understand KoenDG, he tried to point out all places where a '==' comparison is used and someone else should review these places. 

The JIT will probably inline the call so that this call should not affect performance.
On the one hand I like self explaining code, but here I would really suggest to write one or two lines of comment where it is not clear why '==' is used instead of equals

src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -832,7 +833,7 @@ protected boolean areEqual(Object obj1, Object obj2) {
     if (obj2 == null) {
       return false;
     }
-    if (obj1 == obj2) {
+    if (JHelper.objectSameReference(obj1, obj2)) {
```
rPraml: It's a common patteen in equals method. Use == here (.equals) is used later

src/main/java/io/ebean/common/BeanList.java
```diff
@@ -97,7 +98,7 @@ public void internalAddWithCheck(Object bean) {
    */
   private boolean containsInstance(Object bean) {
     for (Object element : list) {
-      if (element == bean) {
+      if (JHelper.objectSameReference(element, bean)) {
```
rPraml: Use == here. Method name makes usage clear

src/main/java/io/ebean/config/ClassLoadConfig.java
```diff
@@ -157,7 +158,7 @@ ClassLoader contextLoader() {
         // try the context loader first
         return classForName(name, contextLoader);
       } catch (ClassNotFoundException e) {
-        if (callerLoader == contextLoader) {
+        if (JHelper.objectSameReference(callerLoader, contextLoader)) {
```
rPraml: It does not matter if you use == or .equals here

src/main/java/io/ebean/config/dbplatform/DbPlatformTypeMapping.java
```diff
@@ -138,14 +139,14 @@ public DbPlatformType lookup(String name, boolean withScale) {
   private DbPlatformType getJsonType(DbType type, boolean withScale) {
 
     DbPlatformType dbType = get(type);
-    if (dbType == JSON_CLOB_PLACEHOLDER) {
+    if (JHelper.objectSameReference(dbType, JSON_CLOB_PLACEHOLDER)) {
```
rPraml: DbType is an enum, so == can be used

src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationBase.java
```diff
@@ -158,7 +159,7 @@ public static boolean isInJavaLangAnnotationPackage(Annotation annotation) {
     if (ann != null) {
       return ann;
     } else {
-      while (clazz != null && clazz != Object.class) {
+      while (clazz != null && !JHelper.objectSameReference(clazz, Object.class)) {
```
rPraml: There is only one Object.class, so it is clear when == is uesd

src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationBase.java
```diff
@@ -191,7 +192,7 @@ public static boolean isInJavaLangAnnotationPackage(Annotation annotation) {
 
     Annotation[] anns = annotatedElement.getAnnotations(); // directly annotatated or inherited
     for (Annotation ann : anns) {
-      if (ann.annotationType() == annotationType) {
+      if (JHelper.objectSameReference(ann.annotationType(), annotationType)) {
```
rPraml: AnnotationTypes are classes, so using == should be safe

KoenDG:
@0xbaadf00d @rPraml Yeah, the entire point is to have people who know the code have look and pose the question "In this instance, is object reference comparison what we intended, or do we in fact want to know if the values are equal?"

These are all instances that Intellij reported as "Comparing object with == instead of .equals()"

I'll have a look at the suggestions later.

KoenDG:
Okay, I removed the suggested instances.

At this point, some of the remaining ones are `Object` or `? extends E` classes.

And since you can do stuff like this: [https://ideone.com/TapVVo](https://ideone.com/TapVVo)

Should I replace them with `equals`? Unless someone says "I know the full breadth of possibilities that will ever come to be in these `Object` instances and it will never be anything where the overriden `equals` has a condition after the `==` check that can return `true`"

KoenDG:
Ah, I forgot about this. Looking back at it, I both find it useful and pedantic at the same time.

I'll leave it up to Rob and how he feels about it.

KoenDG:
Since this hasn't been spoken about in so long and it wasn't really necessary, I'm gonna clean this up.

PR #1031, opened 2017-05-29
FIX: correct behavior of "startsWith, "contains" etc

rPraml:
This PR escapes the like query to perform a true "startsWith", "contains", ... search.

**Use Cases**

1) Ebean.find(Customer.class).where().startsWith("name", "%")
_Expected behavior:_ Get all customers that start with "%"
_Current behavior (before fix):_ All customers were returned (because all match to LIKE '%%')

2) Ebean.find(Customer.class).where().contains("name", "[none]")
_Expected behavior:_ Get all customers where the string "[none]" is in the name.
_Current behavior (before fix):_ Postgres: fine, Sql-Server: Return all customers, that contains one of the characters 'n','o','e'. Sql-Server uses different LIKE syntax: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql

### DIFF
--- a/src/main/java/io/ebean/LikeType.java
+++ b/src/main/java/io/ebean/LikeType.java
@@ -6,22 +6,22 @@ package io.ebean;
 public enum LikeType {
 
   /**
-   * You need to put in your own wildcards.
+   * You need to put in your own wildcards - no escaping is performed.
    */
   RAW,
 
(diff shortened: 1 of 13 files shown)


rbygrave:
Thanks !!

PR #1034, opened 2017-05-29
Pr/id in as collection2

rPraml:
This PR depends on "https://github.com/ebean-orm/ebean/pull/1033"
It adds sanity checks when using "idIn" or "in" queries with an empty collection.
The expected behavior is: Return nothing if collection was empty.


rbygrave:
Thanks, good stuff.

PR #1035, opened 2017-05-29
FIX: deleting and adding same bean in beanCollection leads to deletion in DB

rPraml:
This PR fixes an issue we had when sorting a BeanCollection (for display).
The problem was, that Collections.sort(beanCollection) removes and re-adds the bean to the collection.
Then only deletion was tracked and the bean was deleted in the database.
Now, if a bean is re-added that is marked for deletion, the deletion is undone


rbygrave:
This is a classic for test driven ... ie. adding tests to BeanListTest

rbygrave:
Note that this didn't completely fix the issue.  Sort on REMOVALS fails testing due to the order of adds and removes with the sort.

Pushing more for this.

PR #1036, opened 2017-05-29
Sequence-generators are now tenant aware

rPraml:
Hello Rob,

please take a look at this PR. This makes SequenceGenerators tenant-aware.

FYI: We will use TenantMode DB (instead SCHEMA) now in conjunction with SqlServer due several reasons:
- setSchema is not supported by the driver (required too much hacks in ebean)
- better separation of data (we have to use different db credentials on different servers for each tenant)

=> We dropped the idea to switch the tenant during transaction. A transaction always run for one tenant.

### DIFF
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -2617,6 +2617,28 @@ public class ServerConfig {
     ddlInitSql = p.get("ddl.initSql", ddlInitSql);
     ddlSeedSql = p.get("ddl.seedSql", ddlSeedSql);
 
+    // read tenant-configuration from config:
+    // tenant.mode = NONE | DB | SCHEMA | CATALOG | PARTITION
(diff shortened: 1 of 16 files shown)


rPraml:
@rbygrave what do you think about this change? I think the current implementation can run in problems in multi database tenant mode, because generated (and cached) ids from tenant1 can be used in tenant2 and uniqueness is no longer guaranteed

rbygrave:
So this PR had a number of changes (so not just Tenant aware sequences) ... and I was not comfortable with some of those changes and SequenceGenerator itself had a very large diff ... and there is the option of going back to a more traditional sequence approach by using step 50.  

It is much easier to review and merge a PR that deals with one thing / concept.  i.e. 1 PR for 1 commit.

So yes there is likely a problem with sequences and tenantId but this PR is not quite focused enough to merge currently.

rPraml:
I close this PR and will make a new one soon, because it will conflict anyway with the sqlserver PR

PR #1037, opened 2017-05-30
FIX: JtaTransaction is TenantMode.DB aware

rPraml:
Hello Rob, can you take a look at this. This should make JTA transactions tenant-aware in TenantMode.DB
(otherwise, the DB is used that was selected when construction the TransactionFactory)


rbygrave:
I prefer to do ...  https://github.com/ebean-orm/ebean/issues/1042

https://github.com/ebean-orm/ebean/commit/58c5f01e2b05ec4a51368e4315973a26df393a90

... to achieve the same fix.

rbygrave:
I'll close this, effectively fixed via #1042

PR #1053, opened 2017-07-02
Sqlserver support

rPraml:
Hello Rob,

here comes a bigger PR - it will add/fix the DDL support for Sqlserver - I changed a lot of tests (for the next one or two PRs), but I tried to keep this PR minimal invasive to "src/main" so that it is easier to review

PR that will fix all tests for sqlserver will follow

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,13 @@
   </scm>
 
   <dependencies>
+	<!-- MS SQL Server dependency is now publicly available -->
+	<dependency>
+	  <groupId>com.microsoft.sqlserver</groupId>
+	  <artifactId>mssql-jdbc</artifactId>
+	  <version>6.1.3.jre8-preview</version>
+	  <scope>test</scope>
+	</dependency>
 
     <dependency>
       <groupId>org.jetbrains</groupId>
--- a/src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
+++ b/src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
@@ -54,6 +54,8 @@ public class BaseTableDdl implements TableDdl {
    * Used when unique constraints specifically for OneToOne can't be created normally (MsSqlServer).
    */
   protected List<Column> externalUnique = new ArrayList<>();
+  
+  protected List<UniqueConstraint> externalCompoundUnique = new ArrayList<>();
 
   // counters used when constraint names are truncated due to maximum length
   // and these counters are used to keep the constraint name unique
@@ -85,6 +87,7 @@ public class BaseTableDdl implements TableDdl {
   protected void reset() {
     indexSet.clear();
     externalUnique.clear();
+    externalCompoundUnique.clear();
     countCheck = 0;
     countUnique = 0;
     countForeignKey = 0;
@@ -205,6 +208,18 @@ public class BaseTableDdl implements TableDdl {
     for (Column col : externalUnique) {
       String uqName = col.getUniqueOneToOne();
       String[] columnNames = {col.getName()};
+      write.apply()
+        .append(platformDdl.alterTableAddUniqueConstraint(tableName, uqName, columnNames))
+        .endOfStatement();
+
+      write.dropAllForeignKeys()
+        .append(platformDdl.dropIndex(uqName, tableName))
+        .endOfStatement();
+    }
+    
+    for (UniqueConstraint constraint : externalCompoundUnique) {
+      String uqName = constraint.getName();
+      String[] columnNames = StringHelper.delimitedToArray(constraint.getColumnNames(), ",", false);
       write.apply()
         .append(platformDdl.alterTableAddUniqueConstraint(tableName, uqName, columnNames))
         .endOfStatement();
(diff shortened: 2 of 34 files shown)


pom.xml
```diff
@@ -21,6 +21,13 @@
   </scm>
 
   <dependencies>
+	<!-- MS SQL Server dependency is now publicly available -->
+	<dependency>
```
rPraml: will indent this in next PR correctly

pom.xml
```diff
@@ -21,6 +21,13 @@
...
+	  <groupId>com.microsoft.sqlserver</groupId>
+	  <artifactId>mssql-jdbc</artifactId>
+	  <version>6.1.3.jre8-preview</version>
+	  <scope>test</scope>
+	</dependency>
```
rPraml: Will indent this in next PR correctly

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -213,6 +216,18 @@ protected void writeUniqueOneToOneConstraints(DdlWrite write, CreateTable create
...
+      String uqName = constraint.getName();
+      String[] columnNames = StringHelper.delimitedToArray(constraint.getColumnNames(), ",", false);
+      write.apply()
+        .append(platformDdl.alterTableAddUniqueConstraint(tableName, uqName, columnNames))
+        .endOfStatement();
+
+      write.dropAllForeignKeys()
+        .append(platformDdl.dropIndex(uqName, tableName))
+        .endOfStatement();
+    }
```
rPraml: we must also write compound indices separately

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -832,22 +852,11 @@ protected void alterTableDropColumn(DdlBuffer buffer, String tableName, String c
...
-
-    if (!onHistoryTable) {
-      if (isTrue(column.isNotnull())) {
-        buffer.append(" not null");
-      }
-      if (hasValue(column.getCheckConstraint())) {
-        buffer.append(" constraint ").append(column.getCheckConstraintName());
-        buffer.append(" ").append(column.getCheckConstraint());
-      }
+    String ddl = platformDdl.alterTableAddColumn(tableName, column, onHistoryTable);
```
rPraml: sqlserver has its own alter syntax: 
Mysql/Postgres/H2: "alter table add column <colname>"
Sqlserver: "alter table add <colname>"
so I redirected this to platformDdl

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/SqlServerHistoryDdl.java
```diff
@@ -34,16 +34,25 @@ private void enableSystemVersioning(DdlWrite writer, String baseTable) throws IO
...
+    drop.append("IF OBJECT_ID('").append(baseTable).append("', 'U') IS NOT NULL alter table ").append(baseTable).append(" set (system_versioning = off)").endOfStatement();
+    drop.append("IF OBJECT_ID('").append(baseTable).append("_history', 'U') IS NOT NULL drop table ").append(baseTable).append("_history").endOfStatement();
   }
 
   @Override
   public void dropHistoryTable(DdlWrite writer, DropHistoryTable dropHistoryTable) throws IOException {
     String baseTable = dropHistoryTable.getBaseTable();
     DdlBuffer apply = writer.applyHistory();
-    apply.append("alter table ").append(baseTable).append("set (system_versioning = off)").endOfStatement();
+    apply.append("alter table ").append(baseTable).append(" set (system_versioning = off)").endOfStatement();
```
rPraml: hmm... there were so many errors in the SqlServerHistoryDdl, that let me think that this code was never tested.
@rbygrave if we alter the live table and add a column, we MUST provied a default value, if the column is not null.
(which makes sense)
Can we assume that the default value for numeric columns is always 0, String "", Date 1970-01-01 and add this generally in the DDL script?

src/test/java/io/ebean/BaseTestCase.java
```diff
@@ -15,7 +16,7 @@
 
   static {
     logger.debug("... preStart");
-    if (!AgentLoader.loadAgentFromClasspath("ebean-agent", "debug=1;packages=com.avaje.tests,org.avaje.test")) {
+    if (!AgentLoader.loadAgentFromClasspath("ebean-agent", "debug=1;packages=org.tests,org.avaje.test,io.ebean")) {
```
rPraml: I fixed this, so tests should work if they are launched from eclipse

src/test/java/io/ebean/EbeanServerFactory_MultiTenancy_Test.java
```diff
@@ -8,7 +8,7 @@
 
 import javax.sql.DataSource;
 
-public class EbeanServerFactory_MultiTenancy_Test {
+public class EbeanServerFactory_MultiTenancy_Test extends BaseTestCase {
```
rPraml: Also let them inherit from BaseTestCase, so that they will work from eclipse

src/test/resources/ebean.properties
```diff
@@ -122,3 +122,8 @@ datasource.pg.databaseDriver=org.postgresql.Driver
 datasource.pg.capturestacktrace=true
 datasource.pg.maxStackTraceSize=50
 
+#TODO we need an sqlserver for travis - maybe from azure cloud
+datasource.mssql.username=ebean
+datasource.mssql.password=ebean
+datasource.mssql.databaseUrl=jdbc:sqlserver://10.75.1.40:1433;databaseName=ebean_unittest;selectMethod=cursor
+datasource.mssql.databaseDriver=com.microsoft.sqlserver.jdbc.SQLServerDriver
```
rPraml: ADD datasource to our sqlserver for now. Feel free to change this

pom.xml
```diff
@@ -21,6 +21,13 @@
   </scm>
 
   <dependencies>
+	<!-- MS SQL Server dependency is now publicly available -->
+	<dependency>
+	  <groupId>com.microsoft.sqlserver</groupId>
+	  <artifactId>mssql-jdbc</artifactId>
+	  <version>6.1.3.jre8-preview</version>
```
rbygrave: I think we should look to move this to 6.2.0.jre8 after merge

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
rbygrave: Cool.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/SqlServerHistoryDdl.java
rbygrave: Yeah, I can't remember now ...  assume that the default value for ... yes, we might have a general issue/bug there when adding non null column (must have a default).  I think we must use 0, "" etc at this stage.

rbygrave:
Thanks.  Lots of work here, cheers for that.

pom.xml
rPraml: We are using 6.1.7.jre8 - with 6.2.0 Batch insert is somehow broken

PR #1055, opened 2017-07-02
Sqlserver support 2

rPraml:
This is the next PR that adds a lot of sqlserver fixes, but please merge the first one.

### DIFF
--- a/src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerPlatform.java
+++ b/src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerPlatform.java
@@ -45,7 +45,7 @@ public class SqlServerPlatform extends DatabasePlatform {
 
     this.openQuote = "[";
     this.closeQuote = "]";
-    this.specialLikeCharacters =  new char[] { '%', '_', '[' };
+    this.specialLikeCharacters = new char[]{'%', '_', '['};
     this.likeClause = "like ? COLLATE Latin1_General_BIN";
 
     booleanDbType = Types.INTEGER;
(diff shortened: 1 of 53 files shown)


src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerPlatform.java
```diff
@@ -19,23 +24,22 @@
...
     this.idInExpandedForm = true;
     this.selectCountWithAlias = true;
     this.sqlLimiter = new SqlServerSqlLimiter();
     this.basicSqlLimiter = new SqlServerBasicSqlLimiter();
     this.platformDdl = new SqlServerDdl(this);
     this.historySupport = new SqlServerHistorySupport();
-    this.dbIdentity.setIdType(IdType.IDENTITY);
-    this.dbIdentity.setSupportsGetGeneratedKeys(true);
-    this.dbIdentity.setSupportsIdentity(true);
+    dbIdentity.setIdType(IdType.SEQUENCE);
```
rbygrave: We can't just flip to sequence here without breaking people ... meaning we need a plan (like a ServerConfig option) to allow people to migrate.

Absolutely we want SqlServer folks to move to sequences so we do want this ... just need to be mindful of people already somewhat committed to IDENTITY.

rbygrave:
This PR has overlap with the prior PR.  So I think this can't be merged in it's current state.

src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerPlatform.java
rPraml: I Agree, but before introducing a complex migration path, we should ask who already uses sqlserver productive.
As there were so many bugs and also posts on the google groups I think no one uses it really productive, so I would suggest to do a hard switch from IDENTITY to SEQUENCE. For the few pepole this means either to recreate the database or write some custom migration ddls
If most people do not have a problem

rbygrave: Yes, perhaps the best option is to have a platform that extends SqlServerPlatform and overrides to use IDENTITY etc.

PR #1056, opened 2017-07-02
Misc fixes

rPraml:
Hello @rbygrave 
this is a summary of minor fixes. It would be great if you can review the commits and give me some feedback. 
As this is based on Sqlserver-changes, it would be best if you start here: https://github.com/FOCONIS/ebean/commit/1172dd93fb710df97413115d6cdd128b2ddb5158

Most of them are one-liner so that you should be able to chery pick them if you find it useful. (Or I create new PRs, if you want)

The last one https://github.com/FOCONIS/ebean/commit/b3a85172ad129445ae0156307fe4b210e6ce5918 would optimize the queries. It omits access to the discriminator column whenever it is possible.

We noticed that queries like this 
`select ... from big_table t0 join acl_entry t1 on t1.id = t0.user_id and t1.type='U'`
get horrible slow on postgres (several seconds, depends how big "big_table" is)
As long as no one messes around in the database, the check for `t1.type='U'` is not neccessary here (because it can be only 'U' ) and ommiting this will speed up the queries a lot.

cheers
Roland

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,13 +21,13 @@
   </scm>
 
   <dependencies>
-	<!-- MS SQL Server dependency is now publicly available -->
-	<dependency>
-	  <groupId>com.microsoft.sqlserver</groupId>
-	  <artifactId>mssql-jdbc</artifactId>
-	  <version>6.1.3.jre8-preview</version>
-	  <scope>test</scope>
-	</dependency>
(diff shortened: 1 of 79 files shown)


rPraml:
will close this, most of them will get in separate PRs

PR #1059, opened 2017-07-05
Using Regular expression objects for splitting and replacing in strings.

KoenDG:
This PR replaces all occurrences of a `split()` or `replace()` in the code with the same functionality via a compiled regex Pattern.

This Pattern object can be static and thus not have to be compiled every single time the `.split()` or `.replace()` method is called.

All replacements were exactly as they are implemented in the Oracle JDK source files for the String class.

Also, there was this one class level String variable that wasn't being used anywhere, I removed it.


rbygrave:
Great, thanks !!

PR #1066, opened 2017-07-07
New Mode to control queryCache behavior

rPraml:
Hello Rob, thanks for the cache (it was 2 days too late, becaue on tuesday we implemented a "select count" cache in our application)
Here It was possible to "recache" a value, that means, get the value from DB and store it in the cache.
What do you think about this change in this PR


rPraml:
this PR is obsolete now due PR #1067

PR #1067, opened 2017-07-07
Feature/cache improvements 2

rPraml:
Splitted in a separete PR - this add additional features
- cache findIds
- return unmodifiable collections or CopyOnFirstWrite-lists.


rbygrave:
Hi, So yes this all generally makes sense.  I'm not sure we should deprecate setUseQueryCache(boolean) ... just yet.  We may not ever end up using CacheMode with the bean cache given how  that works.  That means the CacheMode really only adds the RECACHE option.

I'll ponder a bit more tomorrow.

src/main/java/io/ebean/common/BeanList.java
```diff
@@ -541,4 +541,11 @@ public int previousIndex() {
     }
 
   }
+
+  @Override
+  public BeanCollection<E> getShallowCopy() {
+    BeanList<E> copy = new BeanList<>(new CopyOnFirstWriteList<>(list));
+    copy.setFromOriginal(copy);
```
rbygrave: copy.setFromOriginal(this);  -- should use this here ?

rPraml:
... still in progress. I will resolve merge conflicts and so on

rbygrave:
So this is ready to merge in right?

rPraml:
@rbygrave this is ready to merge in.

PR #1068, opened 2017-07-11
Decoupled getServer(...) dependency on having a default server connection

biomechanism:
Hi,

A problem I encountered was a dependency on the default server, whether I wanted to use it or not.
I had setup an in memory hsqldb instance for my tests, and my default server was a hsqldb instance in server mode, the problem is that I had to have the actual server running just to run the tests against the in memory DB. I have removed this coupling by shifting the instantiation code from the constructor into the getDefaultServer method. So now when EBean.getServer(...) is called it can get/create the specified server without trying to make a connection to the default one also.

I also added the <pluginManagement> tag to the pom.xml to prevent Eclipse flagging an error.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,324 +1,324 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-	<modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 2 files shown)


biomechanism:
Hmm, OK, something has gone wrong, I'll look into that over the next day or two when I get a change.

biomechanism:
Seems to be related to enhancement, the current transaction is null. But when (via the Eclipse plugin) I disable enhancement, and enable it again, the tests run fine.

0xbaadf00d:
Maybe the issue is related to the eclipse plug-in, not ebean itself.

biomechanism:
The reason for this change is certainly Ebean, but I seem to have tracked it down. It's actually the <pluginManagement> tag in the pom.xml that Maven doesn't like, and Eclipse doesn't like not having it. I'll kill this current fork and create a new one for a cleaner pull request without the pom.xml modification.

PR #1069, opened 2017-07-12
Removed EBean.getServer(...) dependency on the default server.

biomechanism:
Shifted the default server instantiation from the constructor into the
getDefaultServer method.

The problem is that EBean.getServer(...) cannot be used without EBean trying to make a connection to the default server, as this is performed in the constructor. So in the case that I just want to run tests against the test datasource, doing something like EBean.getServer("testdb"), the default server also has to be up and running, otherwise the tests won't run. If the default server is up, then tests run fine. This coupling has been removed with this change.

### DIFF
--- a/src/main/java/io/ebean/Ebean.java
+++ b/src/main/java/io/ebean/Ebean.java
@@ -141,43 +141,35 @@ public final class Ebean {
      */
     private EbeanServer defaultServer;
 
-    private ServerManager() {
-
-      try {
-        // skipDefaultServer is set by EbeanServerFactory
-        // ... when it is creating the primaryServer
-        if (PrimaryServer.isSkip()) {
-          // primary server being created by EbeanServerFactory
-          // ... so we should not try and create it here
-          logger.debug("PrimaryServer.isSkip()");
-
-        } else {
-          // look to see if there is a default server defined
-          String defaultName = PrimaryServer.getDefaultServerName();
-          logger.debug("defaultName:{}", defaultName);
-          if (defaultName != null && !defaultName.trim().isEmpty()) {
-            defaultServer = getWithCreate(defaultName.trim());
-          }
-        }
-      } catch (Throwable e) {
-        logger.error("Error trying to create the default EbeanServer", e);
-        throw new RuntimeException(e);
-      }
-    }
-
     private EbeanServer getDefaultServer() {
       if (defaultServer == null) {
-        String msg = "The default EbeanServer has not been defined?";
-        msg += " This is normally set via the ebean.datasource.default property.";
-        msg += " Otherwise it should be registered programmatically via registerServer()";
-        throw new PersistenceException(msg);
+        try {
+          // skipDefaultServer is set by EbeanServerFactory
+          // ... when it is creating the primaryServer
+          if (PrimaryServer.isSkip()) {
+            // primary server being created by EbeanServerFactory
+            // ... so we should not try and create it here
+            logger.debug("PrimaryServer.isSkip()");
+
+          } else {
+            // look to see if there is a default server defined
+            String defaultName = PrimaryServer.getDefaultServerName();
+            logger.debug("defaultName:{}", defaultName);
+            if (defaultName != null && !defaultName.trim().isEmpty()) {
+              defaultServer = getWithCreate(defaultName.trim());
+            }
+          }
+        } catch (Throwable e) {
+          logger.error("Error trying to create the default EbeanServer", e);
+          throw new RuntimeException(e);
+        }
       }
       return defaultServer;
     }
 
     private EbeanServer get(String name) {
       if (name == null || name.isEmpty()) {
-        return defaultServer;
+        return getDefaultServer();
       }
       // non-synchronized read
       EbeanServer server = concMap.get(name);
--- a/src/main/java/io/ebean/Ebean.java
+++ b/src/main/java/io/ebean/Ebean.java
@@ -139,29 +139,31 @@ public final class Ebean {
     /**
      * The 'default' EbeanServer.
      */
-    private EbeanServer defaultServer;
+    private volatile EbeanServer defaultServer;
 
     private EbeanServer getDefaultServer() {
-      if (defaultServer == null) {
-        try {
-          // skipDefaultServer is set by EbeanServerFactory
-          // ... when it is creating the primaryServer
-          if (PrimaryServer.isSkip()) {
-            // primary server being created by EbeanServerFactory
-            // ... so we should not try and create it here
-            logger.debug("PrimaryServer.isSkip()");
+      synchronized (monitor) {
+        if (defaultServer == null) {
+          try {
+            // skipDefaultServer is set by EbeanServerFactory
+            // ... when it is creating the primaryServer
+            if (PrimaryServer.isSkip()) {
+              // primary server being created by EbeanServerFactory
+              // ... so we should not try and create it here
+              logger.debug("PrimaryServer.isSkip()");
 
-          } else {
-            // look to see if there is a default server defined
-            String defaultName = PrimaryServer.getDefaultServerName();
-            logger.debug("defaultName:{}", defaultName);
-            if (defaultName != null && !defaultName.trim().isEmpty()) {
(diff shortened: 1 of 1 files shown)


rbygrave:
I'm not convinced by this change.  I don't follow what problem you have wrt testing ... and this change impacts the thread safety which I am not convinced by ... so I don't wish to merge this in.
Cheers, Rob.

biomechanism:
Hi Rob, I'll try get back to you and clarify the problem in a few days, just very busy at the moment. Yeah, I see what you mean about thread safety, that can be addressed. Cheers, Shane.

biomechanism:
Hi Rob,

OK, I've added synchronisation in the getDefaultServer method. 

And to try clarify the problem I've created a very basic example on my GitLab instance. 

https://deta.in/biomechanism/EbeanProblemExample 

This doesn't do anything other than call EBean.getServer() in the test. There are two properties files test-ebean.properties and ebean.properties. run mvn clean test, test runs fine... because the ebean.properties file is using an in memory DB and so there is no problem connecting. However, now either change the default parameters to connect to a DB that isn't up an running, or just remove/rename the ebean.properties file, and rerun mvn clean test. The test will no longer run, this is the problem with the coupling of the instantiation of default server in the constructor. 

I hope that clarifies up the issue further, and why I think it should be instantiated as needed in the getDefaultServer() method.

If I've missed anything, or you want further details on anything, certainly let me know.

Cheers,
Shane.

rbygrave:
So no I'm not happy with this in that ... it adds a `synchronized (monitor)` call for EVERY access to the default server.  This as a single call is small but some apps will be called a lot.  The question becomes, do we really want to pay that cost all the time ... **_vs_** ... look at the testing approach and try and understand the problem.

To me ... it is look at the testing approach.

That is, you need to go back 1 step and tell me what your problem is.  For example, EbeanServer can come up in offline mode (with a DataSource).  Your example doesn't tell me what your problem is.

biomechanism:
OK fair enough if that is not an approach you like, but there is still an issue here to be solved in my view, unless I've completely missed something. So, going back to step 1, the problem is that unit tests won't run without a valid active default server connection even when I do not use it, i.e. in the case when the named server is not "db" (taking the default).

As a question, why does Ebean require a valid active default server connection when it is neither requested nor desired?

The example I provide previously is a demonstration that the unit test will not run if a default server connection cannot be made, and when I am only requesting the "h2" server. OK, you need to change the ebean.properties to something other than an in-memory DB to watch the failure. 

The series of events appears to be.

1. I make a call to Ebean.getServer("h2"); in the test.
2. Ebean class is loaded and initialised, and instantiates a ServerManager;
3. The ServerManager in its constructor tries to create a default server connection, and fails
if it cannot. (This is the problem)
4. If the ServerManager can get a default connection, no problem, continue with the Ebean.getServer("h2") call and all is fine and well.

The issue in point 3 is why I moved the instantiation into the getDefaultServer() call. Now I understand your concerns related to synchronising on the getDefaultServer() call if it is going to be used in some critical loops, but is that really the case? I am actually curious as to how people use it in general. Regardless, there may be better options to ponder. 


Regards,
Shane.

rbygrave:
> OK, you need to change the ebean.properties to something other than an in-memory DB to watch the failure.

The question becomes ... why use a different db name?  Why not call them both "db"  ... with the test DB overriding the 'normal / production' one.  Are you looking to run 2 EbeanServer instances in test?

biomechanism:
No, I have no need for two instances for testing. And I'm aware there are workarounds, but for me the question doesn't become anything else, what I see is fundamentally defective behaviour from a piece of software. A connection a user never requested nor wanted (or is generally not even aware of) should not be happening, and in the case of failure it can be completely confusing if they don't have the time/energy to investigate it. 

If your general advice is to tell users to use workarounds and call them all the same, then I would also ask why bother keeping support for named servers at all, when it is a partially broken implementation?

If it is not something you want to address, then of course that is your choice, and I'll go about using workarounds, as so far I've only played with Ebean a little, but I quite like it. Then, if this is your choice I would suggest you explicitly document the deficiencies of the Ebean.getServer() call to at least try and mitigate some user frustration.

For me, If synchronising in the getDefaultServer() method is the only option (I don't know if it is) I still think that is what should be done. It is a non-breaking change that is transparent to the user, and solves the problem. And as I mentioned previously I understand concerns about synchronisation in general, but again I'm wholly unconvinced this is really an issue here. You would be saying you are not interested in fixing a particular deficiency because it might have a significant impact on some applications... then again, it might not. My guess would be it won't have much of an impact on things, if it does it's likely to be a very small amount of applications, and if that is the case then it should not be difficult for them to get just the server once and pass it around to solve their issues. Changes here could be highlighted in the docs and release info, making users aware of the situation.

Anyway, that's my view on it, the choice is obviously yours.

Regards,
Shane.

biomechanism:
Actually, I'll just be a little more accurate in my description, while I said it is a deficiency in the Ebean.getServer() call, this isn't entirely accurate, but it is where the issue manifests in my usage. However if there are other methods people may be likely to make first use of in a similar context they will also be a problem, as it is a side effect introduced by having the default connection setup in the constructor of the ServerManager. 

Regards,
Shane.

rbygrave:
Ok lots of points here so I'll try to address them.

Firstly, Ebean is 10+ years old.  I've made a lot of mistakes in that time and will certainly make more mistakes going forward. I am trying to make good decisions which will last for the next 15 years and avoid decisions I will regret as bad decisions get harder and harder to undo going forward from here.

What this potential change does is add a volatile read and sync block on what I consider a 'hot method' / path. I am pretty particular around these types of changes ... hence I really really don't want to make this change unless I **really** have to.

What I also believe (rightly or wrongly) is that testing is trending away from in-memory H2 type testing towards.  That in many cases it becomes more sensible to run tests against a full Postgres or MySql (JSON, JSONB, Array types, select for update, PostGIS etc) as we can't really get matching features in H2. This probably isn't as true with Oracle and DB2 etc but imo there are growing reasons (with docker etc) to actually test against a full Postgres or MySql.

The implication here for Ebean is that the Model / Active record style pattern which heavily relies on Ebean.getDefaultServer() could become the dominant way to use Ebean (it might already be, I don't know). 

These things all mean I want to keep Ebean.getDefaultServer() operate like it is (without volatile read and sync) as long as that is possible.


>  it should not be difficult for them to get just the server once and pass it around to solve their issues.

That is an approach but it is not how the Model / Active record style use of Ebean works - it is reliant on Ebean.getDefaultServer().  This makes this change significant.


 > why bother keeping support for named servers at all, when it is a partially broken implementation?

I believe it is just that there is a limitation that the 'default server' needs to be up (when initialising any other named server). I don't think it is broken per say but instead a limitation when accessing a 'non default named server'.  It is a limitation I am happy with at this point.


> the case of failure it can be completely confusing if they don't have the time/energy to investigate it.

Yup, sorry about that.  


I believe there are improvements that need to come to help in testing against docker container DB's where we may need to do things very early in the bootup / initialisation of the default server ... like start a docker container and execute some commands etc.  

Hopefully that answers some of your points above.

Cheers, Rob.

biomechanism:
Testing against 'proper' DB instances doesn't change anything, it is just that having an in-memory DB as the Primary Server disguises the issue, as a connection will be successful, and so the problem won't manifest in the call. It is irrelevant whether the test DBs are in-memory or not.

> The implication here for Ebean is that the Model / Active record style pattern which heavily relies on Ebean.getDefaultServer() could become the dominant way to use Ebean (it might already be, I don't know).

Dominant or not, I don't think that affects anything to any significant degree. 

Anyway, while I might be critical of this side effect in Ebean, please don't think it is any kind of personal criticism, or indeed even Ebean as a whole. Had it been otherwise I wouldn't have bothered trying to contribute a fix to an issue I encountered. So instead I would say thanks for what seems to be a pretty great bit of software. 

And if I can't convince you this is worth making then fair enough, I'll close the pull request in the coming days.

Regards,
Shane.

rbygrave:
Thanks Shane.  Yes, I think we should close this for the moment and see how we go with testing using the current initialisation. Cheers.

PR #1073, opened 2017-07-18
Add SqlServerViewHistoryPlatform for Pre SqlServer 2017 (Non native SQL 2011 History support)

onilton:
Add support for history for SqlServer versions before 2016 (not native)

### DIFF
--- a/src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerViewHistoryPlatform.java
+++ b/src/main/java/io/ebean/config/dbplatform/sqlserver/SqlServerViewHistoryPlatform.java
@@ -1,0 +1,12 @@
+package io.ebean.config.dbplatform.sqlserver;
+
+/**
+ * SQL Server platform using the ViewHistory for
+ * versions older than SQL Server 2016
+ */
+public class SqlServerViewHistoryPlatform extends SqlServerPlatform {
+
+  public SqlServerViewHistoryPlatform() {
+    this.historySupport = new SqlServerViewHistorySupport();
+  }
+}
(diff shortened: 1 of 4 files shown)


rbygrave:
Hi, so this is good etc but the question is should you keep this locally to your project or should this be merged into the Ebean distribution.

That is, as a project we probably don't have the resources/time to complete and maintain a full "SqlServer 2016 history" with DDL generation etc given that with SqlServer 2017 we have the native support etc.

That is, if we don't have many pre-SqlServer2017 users needing @History we are better off leaving that out of "Ebean proper" and sticking to supporting SqlServer 2017 fully etc.

Does that make sense?
Are there people out there who want to implement the DDL aspect required and to maintain that code?

onilton:
I still plan to write the dll generation code.

I understand your point, but looking at the code you still have a plataform specifically target 2005, for example. We still support it. Even if microsoft itself does not.

Actually, I thought about renaming `SqlServerViewHistoryPlatform` to `SqlServerPre2016Platform` to be more explicitly of what version we support (like in the 2005 case).

Sometimes it is good to force people to use newer versions, but sometimes like with the project I am working on, I can't touch or manage the infrastructure so I am stuck with an "older" version of SQLServer.

But back to the 2005 version, ebean supports 2005 to 2016 versions, so we support 2005, 2008, 2008 R2, 2012, 2014 and 2016.

It is 5 versions against one with native history support. I am really in favor of dropping views/triggers table support in the future, but I don't think a version with native history it is the most used right now.

 (I could'nt find newer data about share of version, but anyway I found this outdated one, not much trustworthy one https://web.archive.org/web/20170323231419/https://www.spotlightessentials.com/world/sql-server).

But I trust your decision on this. 

If we don't merge this, it is ok. I will probably find another way to get/use history, but I don't plan to maintain a custom local version/fork of ebean, it will be hard for me :(

Maybe we cand add @rPraml to the discussion, since he has being submitting a lot of fixes/improvements to sqlserver ?

rbygrave:
Ok, this is nice and useful stuff here but we are not really going to merge it and I think it's time to close it.

Thanks @onilton

PR #1075, opened 2017-07-21
Use Double.compare and Float.compare to do comparison of floating point values.

KoenDG:
Something small I noticed while looking at some code, not sure if it's needed/wanted. 

I changed some boolean checks to use the recommended method of comparing floating point values in Java.

Using `==` for this can be unreliable both in terms of actual value and the possibility of `NaN` values.

The Oracle JDK8 source code for Float comparison looks like this:

```
public static int compare(float f1, float f2) {
        if (f1 < f2)
            return -1;           // Neither val is NaN, thisVal is smaller
        if (f1 > f2)
            return 1;            // Neither val is NaN, thisVal is larger

        // Cannot use floatToRawIntBits because of possibility of NaNs.
        int thisBits    = Float.floatToIntBits(f1);
        int anotherBits = Float.floatToIntBits(f2);

        return (thisBits == anotherBits ?  0 : // Values are equal
                (thisBits < anotherBits ? -1 : // (-0.0, 0.0) or (!NaN, NaN)
                 1));                          // (0.0, -0.0) or (NaN, !NaN)
    }
```
And Double comparison is identical.

EDIT: Weren't there usually unit tests that got run by Travis or something?

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -1027,7 +1027,7 @@ public final class EntityBeanIntercept implements Serializable {
 
     if (state == STATE_NEW) {
       setLoadedProperty(propertyIndex);
-    } else if (oldValue != newValue) {
+    } else if (Double.compare(oldValue, newValue) != 0) {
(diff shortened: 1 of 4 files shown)


rbygrave:
Thanks !!

Issue #1076, opened 2017-07-25
Connection obtained from DataSource is not released after JTA transaction completes


hofmanndavid:
## Expected behavior
When using ebean in a CMT it should obtain a connection from the datasource and then close the connection after the transaction is completed

## Actual behavior
The connection is not closed (or returned to the pool) after the transaction is completed.

### Steps to reproduce
1. Make your container (wildly 10 in my case) have this datasource `java:jboss/datasources/ebeanDS` and make it have at most 3 connections so the issue will show quickly.
2. Deploy this war to the container [https://github.com/hofmanndavid/ebean-vaadin-cdi](https://github.com/hofmanndavid/ebean-vaadin-cdi)
3. Go to localhost:8080/demo-ebean/api/persons/create and refresh a couple of times

#### History of this issue is discussed in the group
https://groups.google.com/forum/#!topic/ebean/Jc4mmVp_Iuk

My workaround is to add a listener to the transaction completion and close the connection obtained by ebean
```java
DefaultServer ds = (DefaultServer) Ebean.getDefaultServer();
JtaTransaction cst = (JtaTransaction) ds.getCurrentServerTransaction();
TransactionSynchronizationRegistry tsr = (TransactionSynchronizationRegistry) new InitialContext().lookup("java:comp/TransactionSynchronizationRegistry");
tsr.registerInterposedSynchronization( new Synchronization() {
    public void beforeCompletion() { }
    public void afterCompletion(int status) {
        try { cst.getInternalConnection().close(); } catch (SQLException ignore) { }
    }
});
```

My understandting is that the closing of the connection should be implementd in the `JtaTransactionManager.JtaTransactionListener::afterCompletion()` method.

Please point out if I am missing something otherwise I'd love to make a PR

rbygrave:
We can close this now, will be in next release.

PR #1077, opened 2017-07-25
Fix connection leak in JTA managed transactions

hofmanndavid:
Fix for issue https://github.com/ebean-orm/ebean/issues/1076


rbygrave:
Thanks !!

PR #1078, opened 2017-07-25
Added the Serializable interface to classes that extend/implement a Comparator with an anonymous/typed class.

KoenDG:
Hello Rob and others.

Going over code analysis and I ran into this. I'm not entirely sure it's needed/necessary, but I did it anyway. Up to your judgment on how valuable/disruptive this is. I'm not sure it will actually do anything.

Basically, I did this: https://stackoverflow.com/a/8642040/5629413

And then added a serialversionUID for those classes.

But only for classes where the type passed along is either anonymous or `<T>`.

There are some Comparators left but those are for predefined classes used by ebean internally, so I didn't think it necessary.,


rbygrave:
Thanks !

PR #1082, opened 2017-08-17
Fix: query table entity or sql entity has no id

yuanxuegui:
When query table entity or sql entity，include NamedQuery which modified via code and RawSqlQuery which modified via code, findOne or findUnique will throw NullPointerException.


yuanxuegui:
Only commit c231e0cf4d571bf20a89f18998d7a607dd35931b is for this PR.

yuanxuegui:
Commit 9075c67d7c2b4cbb07378f15d5df236b0f7d9535 and 92e8095f0dc64a45c9554a07fd4e412304c986d6 is new api added for filter query

rbygrave:
Great, thanks !!

PR #1085, opened 2017-08-24
Only throwing warning for autocommit when needed

Rdebokx:
The warning considering autocommit being set to true can be annoying when you choose to use autocommit in your application. This change should make the DefaultContainer only throw this warning when the ServerConfig is not set to use autocommit-friendly TransactionManager.


rbygrave:
Great, thanks.

Rdebokx:
Thank you!

PR #1103, opened 2017-08-31
Pr/add tests

rPraml:
This PR adds/change some test cases that I've made in my branch that are mainly sqlserver related


src/test/java/org/tests/query/other/TestQuerySingleAttribute.java
```diff
@@ -62,11 +62,10 @@ public void withOrderBy() {
         .setDistinct(true)
         .select("name")
         .where().eq("status", Customer.Status.NEW)
-        .orderBy().asc("name")
-        .setMaxRows(100);
+        .orderBy().asc("name");
```
rPraml: sqlserver produces "select top 100..." instead of "select ... limit". So I removed this, so that the test will easier pass the sqlserver platform

PR #1107, opened 2017-08-31
REFACTOR: Created a "splitNames" in StringHelper

rPraml:
and use it in various places. Also wrote some tests for StringHelper.

The method replaces a common usage of this pattern: Pattern.compile("[ ,;]"), followed by a trim and an isEmpty check.

It changes the behaviour a bit
- it splits on every whitespace now (not only on space)
- it splits on "\s" "," and ";" (there are some places where only on "," or " "+"," was split)

### DIFF
--- a/src/test/java/io/ebean/StringHelperTest.java
+++ b/src/test/java/io/ebean/StringHelperTest.java
@@ -1,0 +1,79 @@
+package io.ebean;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.*;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Test;
+
+import io.ebean.util.StringHelper;
+
+public class StringHelperTest {
+
+  
+  @Test
+  public void testSplitNames() {
+    assertThat(StringHelper.splitNames("")).hasSize(0);
+    assertThat(StringHelper.splitNames(" , ;")).hasSize(0);
+    assertThat(StringHelper.splitNames("foo bar")).containsExactly("foo", "bar");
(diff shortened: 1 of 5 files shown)


src/test/java/io/ebean/StringHelperTest.java
```diff
@@ -0,0 +1,79 @@
...
+    assertThat(StringHelper.splitNames("foo bar")).containsExactly("foo", "bar");
+    assertThat(StringHelper.splitNames(" foo \n bar    ")).containsExactly("foo", "bar");
+    assertThat(StringHelper.splitNames(" foo , bar ;")).containsExactly("foo", "bar");
+    assertThat(StringHelper.splitNames("foo, bar")).containsExactly("foo", "bar");
+    assertThat(StringHelper.splitNames("foo, bar baz")).containsExactly("foo", "bar", "baz");
+    assertThat(StringHelper.splitNames("foo, bar\nbaz")).containsExactly("foo", "bar", "baz");
+  }
+  
+  @Test
+  public void testReplaceStringMulti() {
```
rPraml: I added some more tests to stringHelper

src/test/java/io/ebean/StringHelperTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.Test;
+
+import io.ebean.util.StringHelper;
+
+public class StringHelperTest {
+
+  
+  @Test
+  public void testSplitNames() {
+    assertThat(StringHelper.splitNames("")).hasSize(0);
```
rPraml: This test should cover common use cases

rbygrave:
Great stuff.

PR #1118, opened 2017-09-09
Featur/dbmigration 3

rPraml:
Hello Rob, here is the next part of DbMigration improvements


src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -715,7 +729,7 @@ public void generate(DdlWrite writer, AddColumn addColumn) throws IOException {
       alterTableAddColumn(writer.apply(), tableName, column, false);
     }
 
-    if (isTrue(addColumn.isWithHistory())) {
+    if (isTrue(addColumn.isWithHistory()) && !sql2011History) {
```
rPraml: as far as I understand, when the db provides true sql2011 history-support, we must not touch the *_history tables.
(at least SqlSever adds new colums automatically to the history table)

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -741,6 +755,15 @@ public void generate(DdlWrite writer, AddColumn addColumn) throws IOException {
...
 
     dropTable(writer.apply(), dropTable.getName());
+
+    if (hasValue(dropTable.getSequenceCol())
+        && platformDdl.getPlatform().getDbIdentity().isSupportsSequence()) {
+      String sequenceName = dropTable.getSequenceName();
+      if (!hasValue(sequenceName)) {
+        sequenceName = namingConvention.getSequenceName(dropTable.getName(), dropTable.getSequenceCol());
+      }
+      dropSequence(writer.apply(), sequenceName);
```
rPraml: When dropping a table, we should also drop the sequence. For this, I had to extend the DropTable XML-model, so that this will work only for further migrations.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -970,17 +993,20 @@ protected void addUniqueConstraint(DdlWrite writer, AlterColumn alter, String uq
 
 
   protected void alterTableDropColumn(DdlBuffer buffer, String tableName, String columnName) throws IOException {
-
-    buffer.append("alter table ").append(tableName).append(" drop column ").append(columnName)
-      .endOfStatement();
+    platformDdl.alterTableDropColumn(buffer, tableName, columnName);
```
rPraml: dropColumn is now handled by platformDdl, because for SqlServer we must delete all refering objects before we can delete the column

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -970,17 +993,20 @@ protected void addUniqueConstraint(DdlWrite writer, AlterColumn alter, String uq
...
   }
 
   protected void alterTableAddColumn(DdlBuffer buffer, String tableName, Column column, boolean onHistoryTable) throws IOException {
     DdlMigrationHelp help = new DdlMigrationHelp(tableName, column);    
-    help.writeBefore(buffer);
+    if (!onHistoryTable) {
```
rPraml: I think, the DbMigration should be applied only on main table.

src/test/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl_AlterColumnTest.java
```diff
@@ -157,7 +162,7 @@ public void testAlterColumnDefaultValue() throws Exception {
     assertEquals("alter table mytab alter acol set default 'hi'", sql);
 
     sql = sqlServerDdl.alterColumnDefaultValue("mytab", "acol", "'hi'");
-    assertEquals("alter table mytab add constraint df_mytab_acol default 'hi' for acol", sql);
```
rPraml: I had problems with named default constrainsts & temporal tables in SqlServer. (I could not add non null columns) - so I reverted the change for named columns and let the sqlserver generate a (random) name for defaults.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/SqlServerDdl.java
```diff
@@ -102,13 +124,24 @@ public String createSequence(String sequenceName, int initialValue, int allocati
...
   public String alterColumnDefaultValue(String tableName, String columnName, String defaultValue) {
-
+    // Unfortunately, the SqlServer creates default values with a random name.
+    // You can specify a name in DDL, but this does not work in conjunction with
+    // temporal tables in certain cases. So we have to delete the constraint with
+    // a rather complex statement.
+    StringBuilder sb = new StringBuilder();
     if (DdlHelp.isDropDefault(defaultValue)) {
-      return "alter table " + tableName + " drop constraint df_" + tableName + "_" + columnName;
+      sb.append("delimiter $$\n");
```
rPraml: What do you think about this hack?: I need to execute the next statements in one request.
delimiter is MYSQL specific, but works also on MSSQL due special handling in DDL-script runner.
But this means, that I cannot copy & paste the generated script to SqlStudio and execute it there. (maybe I'll change that and wrap it in an EXEC('....') statement)

rbygrave: I think someone in SQL Server land needs a good talking to.  A thought is to create and use migration specific stored procedures for sql server.  That is, ensure those stored procedures are created (prerequisite) ... and use them to do this (and perhaps some other things)

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
rbygrave: Good.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
rbygrave: Yes, null and check constraints etc should not be on a history table per say.  So yes, migrations should only apply to the main table.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/SqlServerDdl.java
rbygrave: This is better than what we have but yes longer term ... I suspect  it might be nicer/cleaner to actually create/ensure we have some migration specific stored procedures [for sql server] that we use to drop default constraints, drop columns etc in sql server.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/SqlServerDdl.java
```diff
@@ -157,4 +190,28 @@ public void addColumnComment(DdlBuffer apply, String table, String column, Strin
 
     // do nothing for MS SQL Server (cause it requires stored procedures etc)
   }
+  
+  /**
+   * It is rather complex to delete a column on SqlServer as there must not exist any references
+   * (constraints, default values, indices and foreign keys). The list is not yet complete, as
+   * indices over multiple columns will not yet deleted.
+   * (This may be changed to delete all refering objects by using the sys.* tables later)
```
rbygrave: Yes, I like the idea of a set of migration stored procedures one of which we could use for 'drop column'

src/test/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl_AlterColumnTest.java
rbygrave: ok

PR #1121, opened 2017-09-11
Feature/dbmigration 4

rPraml:


src/main/java/io/ebean/config/dbplatform/DbDefaultValue.java
```diff
@@ -75,14 +95,28 @@ public String convert(String dbDefaultLiteral) {
     if (dbDefaultLiteral == null) {
       return null;
     }
+    if (dbDefaultLiteral.endsWith("}")) {
```
rPraml: I decided to write the timestamps in ODBC compliant form, e.g: { ts 'yyyy-mm-dd hh:mm:ss' } as most platforms can handle them. The values are then converted back for the current platform

src/main/java/io/ebean/config/dbplatform/postgres/PostgresPlatform.java
```diff
@@ -42,7 +42,11 @@ public PostgresPlatform() {
     this.dbIdentity.setIdType(IdType.IDENTITY);
     this.dbIdentity.setSupportsGetGeneratedKeys(true);
     this.dbIdentity.setSupportsSequence(true);
-
+    
+    this.dbDefaultValue.setToDate("${}");
+    this.dbDefaultValue.setToTime("${}");
+    this.dbDefaultValue.setToTimestamp("${}");
```
rPraml: these are the platform specific functions.
`TIMESTAMP(${})` will convert `{ ts '2001-01-01 12:34:56' }` to `TIMESTAMP('2001-01-01 12:34:56')`

src/main/java/io/ebean/dbmigration/ddlgeneration/BaseDdlHandler.java
```diff
@@ -59,6 +61,10 @@ public void generate(DdlWrite writer, ChangeSet changeSet) throws IOException {
         generate(writer, (AddHistoryTable) change);
       } else if (change instanceof DropHistoryTable) {
         generate(writer, (DropHistoryTable) change);
+      } else if (change instanceof AddUniqueConstraint) {
```
rPraml: Add support to add unique keys or alter foreign keys in DbMigration

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/BaseTableDdl.java
```diff
@@ -335,7 +337,7 @@ protected void writeUniqueOneToOneConstraints(DdlWrite write, CreateTable create
       }
       String[] columnNames = {col.getName()};
       write.apply()
-        .append(platformDdl.alterTableAddUniqueConstraint(tableName, uqName, columnNames, Boolean.TRUE.equals(col.isNotnull())))
+        .append(platformDdl.alterTableAddUniqueConstraint(tableName, uqName, columnNames, Boolean.TRUE.equals(col.isNotnull()) ? null : columnNames))
```
rPraml: unfortunately SQLServer, Oracle & DB2 handle null values different than Postgres, MySql & H2.
They allow only one null value. So I add the nullable columns here to exclude them from the unique index. (not yet implemented for Oracle & DB2)

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl.java
```diff
@@ -219,7 +222,7 @@ public String convertDefaultValue(String dbDefault) {
    * Return the drop foreign key clause.
    */
   public String alterTableDropForeignKey(String tableName, String fkName) {
-    return "alter table " + alterTableIfExists + tableName + " " + dropConstraintIfExists + " " + fkName;
+    return "alter table " + alterTableIfExists + tableName + " " + dropConstraintIfExists + " " + maxConstraintName(fkName);
```
rPraml: I had problems with the constraint name limit on oracle. First I thought, that ebean already shortens the names, but as far as I understand the dbmigration process, table names are shortened during the deploy phase and the BeanDescriptor has already the shortened names. But this happens with the current platfrom, which is H2(offline)
So the names in the model.xml's are too long for oracle & db2. This is easy to fix by specifying a shorter name with a @Table() annotation.

Unfortunately foreignkeys of m2m tables are rather long, as the m2m table name = table1+table2 and the fk is m2mTableName+table1+propName which gets quite huge: 'fk_migtest_mtm_c_migtest_mtm_m_migtest_mtm_m'

So I shorten the names here.

src/main/java/io/ebean/dbmigration/ddlgeneration/platform/PlatformDdl.java
```diff
@@ -390,21 +393,26 @@ public void alterTableAddColumn(DdlBuffer buffer, String tableName, Column colum
...
     if (!onHistoryTable) {
-      if (isTrue(column.isNotnull())) {
-        buffer.append(" not null");
-      }
+
 
       if (defaultValue != null) {
-        if (typeContainsDefault(convertedType)) {
-          logger.error("Cannot set default value for '" + tableName + "." + column.getName() + "'");
+        int pos = convertedType.indexOf(" default");
```
rPraml: when the convertedType contains already a default value (e.g. 'bit default 0') I remove it and set the correct default value.

rPraml:
This PR is not yet finished. There are some other features which I want to merge first:
- fix test cases, so that they pass on other platforms (especially oracle/db2)
- support for arrays/table value parameters in "select * from table where id in (?)" queries (Sqlserver supports max 2100 params, postgres 2^16, oracle 1000)
- tenant-safe sequences
- query to count distinct values: "SELECT name, count(*) form table group by name"
- minor bugfixes

rPraml:
@rbygrave will close this and create a new one

PR #1124, opened 2017-09-12
replaced findUnique by "default" method in interfaces.

rPraml:
Hello Rob, I would suggest to replace "findUnique" by a default implementation.


rbygrave:
Yes, good idea - thanks.

PR #1132, opened 2017-09-14
Feature/contitional test runner

rPraml:
Hello Rob,

this PR adds a "ConditionalTestRunner", so that you can exclude/include tests for certain platforms.
I would prefer
```java
@ForPlatform(Platform.H2)
@Test
public void myTest() {
   ...
}
```
instead of
```java
@Test
public void myTest() {
  if (!isH2()) {
    return;
  }
   ...
}
```
as this gives you a better overview about test statistics.
The first test would be counted as "skipped" if platform does not match, the second test as "passed" although it did not test anything.


src/test/java/io/ebean/PrimaryServerTest.java
```diff
@@ -19,8 +20,8 @@ public void testIsSkipPrimaryServer() throws Exception {
     assertFalse(PrimaryServer.isSkip());
   }
 
-  @Ignore
   @Test
+  @ForPlatform(Platform.H2)
```
rPraml: enabled this test for H2

src/test/java/io/ebean/UpdateQueryTest.java
```diff
@@ -30,6 +33,7 @@ public void basic() {
   }
 
   @Test
+  @IgnorePlatform(Platform.SQLSERVER)
```
rPraml: TableAlias does not work on sqlserver, only when the (slower) select mode=cursor is used. (but then other tests will fail)

src/test/java/io/ebean/annotation/ForPlatform.java
```diff
@@ -0,0 +1,17 @@
+package io.ebean.annotation;
```
rPraml: Maybe you can take a look at this. I see that you change lots of package-names to be java9 compatible. I don't know if duplicate packages causes issues in unittests (there's no need to move this annotation to the ebean-annotations artifact IMHO)

rbygrave: > no need to move this annotation to the ebean-annotations

I agree. LGTM

PR #1133, opened 2017-09-14
Cherry picks

rPraml:
Hello Rob,

I open this PR, primary to give me some feedback (it would be best if you review at commit level, as it depends on other open PRs)

It prepares ebean to the next bigger feature: Support TableValueParameters/Arrays for "in" queries.

*What is this*
this code should explain:
```java
    if (isSqlServer()) {
      assertThat(query.getGeneratedSql()).contains("where t0.name in (select * from ?)");
    } else if (isH2()) {
      assertThat(query.getGeneratedSql()).contains("where t0.name in (select * from table(x varchar = ?))");      
    } else if (isPostgres()) {
      assertThat(query.getGeneratedSql()).contains("where t0.name in (select(unnest(?)))");
    } else {
      assertThat(query.getGeneratedSql()).contains("where t0.name in (?, ? )");
    }
```

I got it working for H2/PG/mssql/ora, because we hit the limit of max 2100 parameters in a prepared statemet.


rPraml:
most of that already merged / no longer relevant - so closing this

PR #1135, opened 2017-09-14
Use TableValues / array for select prop in (?) queries.

rPraml:
(as this PR depends on others, they should be reviewed first)
This PR adds support for efficent(er) `select prop in (?) `
It uses an sql array or TableValueParameter (depends on which is supported by platform), so that you do not run in a query limit. It is currently implemented only for simple types. (no composite keys!)

If the platform does not support efficient binding, it falls back to the current behaviour by repeating the `?` as many as query values you have.

known limits:
- SqlServer: max 2100 parameters per query
- Oracle: max 1000 parameters per query
- Postgres: max 2^16 parameters per query


rPraml:
hmm - travis is not happy with the test that tries to delete 66000 IDs. I think I've to disable this.

rPraml:
will close this - implemented by #1186

PR #1146, opened 2017-10-03
Refactor dependency

rPraml:
Hello Rob,

here's a "proof of concept PR" how I would refactor the source, so that there are no ebeaninternal imports in "io.ebean"
this means io.ebeaninternal could become own maven module now, because there are no cyclic dependencies between io.ebean <-> io.ebeaninternal
Unfortunately I'd to change some important api classes, especially the docstore classes - so ebean-elastic needs probably also refactoring and must use the interfaces instead of ebeaninternal (which is cleaner anyway) 

What do you think about such a change?


rbygrave:
The initial thoughts are what we are looking to achieve and why with a view to Java9 modules.  In Java 9 all of io.ebeaninternal will not be published/public outside of the java9 module.  That is, nothing in io.ebeaninternal is exposed to anyone (which is nice, we want to hide as much as we can).

The goal is to expose a nice limited API and a nice SPI for extensions/plugins.  The API and SPI should be as abstract as possible (interfaces) and as small as possible -  everything else is hidden as implementation detail.  With Java 9 we don't **need** to split into 2 artifacts.

My thought is that this change exposes far too much into API / SPI. 

For example, this change moves types from io.ebeaninternal.server.type into io.ebean.type ... and so to me in a Java9 world that is not what we trying to achieve - we want to keep those internal.

The current SPI (io.ebean.plugin) is pretty limited and needs attention, the current dbmigration does not have a nice clean split between API and implementation ... so these do need to be addressed but this PR exposes too much as public API in my view.

Q: What benefits do we get from the splitting into 2 artifacts? 
Rob: It forces cleaner dependency lines in Java 8 between io.ebean and io.ebeaninternal + ability to have other alternative implementations.

Q: Will there ever be an alternative implementation of io.ebean ?
Rob: Probably not.

Q: Does this PR expose / grow the "public API" too much?
Rob: Yup .. too much.

My thought is the next thing to do is extract an interface(s) from DbMigration ... have it only depend on things that are configuration ( in io.ebean.config ) and move all the dbmigration implementation detail into io.ebeaninternal.   DbMigration (interface) should provided by service loader so add a DbMigrationFactory to bootstrap it.  That ought to mean that a lot more stays in io.ebeaninternal (and stays as hidden implementation detail).

rbygrave:
So a second quick look and extra thoughts are:

- Yes, we should extract an interface from DbMigration and move basically all of io.ebean.dbmigration into io.ebeaninternal
- That leaves what looks like 6 odd issues with dependencies from io.ebean into io.ebeaninternal which need addressing
- We can use service loader pattern to break some of those
- ExtraTypeFactory should move into io.ebeaninternal
- There is some thinking / options to consider for some (like the hidden DRawSqlParser - could use service loader for that though)

rbygrave:
Right, I'm going to get into a refactor of DbMigration ... hopefully you haven't started that. If you have please let me know - ta.

rPraml:
No, I have not.

I think, moving dbmigration to ebeainternal is a good step, as it was responsible for most of the API changes in BeanType/Property.

Q: if we go to Java 9, ebeaninternal would not be accessible at all from my application, right?
I still use some ebeaninternal classes (e. g. Some type converters) So we (at least I) will need an API for this, too.

rbygrave:
> Java 9, ebeaninternal would not be accessible at all from my application, right?

In strict Java 9 module mode yes'ish ... in that (last time I looked) modules can be "friends" and have special visibility. So more to look into there. 

I expect to work on the plugin API (io.ebean.plugin) and expose more there as needed.

rbygrave:
Ok, that is all complete now - now dependencies from io.ebean to io.ebeaninternal

rPraml:
@rbygrave Really good work, I close this PR now

PR #1156, opened 2017-10-04
commit transaction only if active (hides root case exception when findIterate query throws)

rPraml:
Hello Rob, when transaction is not active, it throws a "Transaction is inactive" exception

happened here
```
public <T> QueryIterator<T> findIterate(Query<T> query, Transaction t) {

    SpiOrmQueryRequest<T> request = createQueryRequest(Type.ITERATE, query, t);
    try {
      request.initTransIfRequired();
      return request.findIterate();

    } catch (RuntimeException ex) {
      request.endTransIfRequired(); // <---- // throws again an exeption and hides the runtime-exception
      throw ex;
    }
  }
```


rbygrave:
Yup, good - thanks.

PR #1157, opened 2017-10-05
Misc platform fixes

rPraml:
- Oracle does not return row count - so we probably cannot detect Optim…
- DB2 wants a varchar null value (other platfroms are happy with this, …
- FIX Test for DB2 platform


rbygrave:
Good stuff.

PR #1158, opened 2017-10-05
Can handle one to many & many to one relations in docstore only entities

rPraml:
Hello Rob,

We use the ebean-json stuff in conjunction with docstore-only entitites to convert complex object graphs to JSON (and we refer to real entities, that will then lazy load)

Here we had the problem that @OneToMany relations did not work, as ebean thinks it requires a mappedBy field on the many-side. This is not true for docstore-only entities.

I added some unit-tests to provide a simple use case.

cheers
Roland


rbygrave:
Nice catch - I like it.

PR #1164, opened 2017-10-06
FIX: connection leak in query-iterator

rPraml:
When the iterator is at the end, it auto-closes the connection


rbygrave:
Yup, I'd like to encourage the use of try with resources or finally blocks (as per the javadoc for QueryIterator) so in theory people shouldn't rely on this 'fix' per say.

PR #1166, opened 2017-10-10
FIX: Ensure that Ebean.currentTransaction is available in lifecycle handlers

rPraml:
_Expected behaviour_
During a CRUD operation, the current transaction should be availabe in the lifecycle handlers

_Current behaviour_
When saving a bean, an implicit transaction 1 is created, but Ebean.currentTransaction is null.
If you try to do a Ebean.find() e.g. in PostPersist it starts a new transaction 2.
If the find result depends on current bean that is saved, you may run in a dead lock, as SQLServer waits until transaction 1 is commited or rolled back.

_Suggested fix_
createServerTransaction and createQueryTransaction set the created transaction in the transactionScopeManager

### DIFF
--- a/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -2130,12 +2130,26 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
 
   @Override
   public SpiTransaction createServerTransaction(boolean isExplicit, int isolationLevel) {
-    return transactionManager.createTransaction(isExplicit, isolationLevel);
+    SpiTransaction t = transactionManager.createTransaction(isExplicit, isolationLevel);
+    try {
+      transactionScopeManager.set(t);
+    } catch (PersistenceException existingTransactionError) {
+      t.end();
+      throw existingTransactionError;
+    }
+    return t;
   }
 
   @Override
   public SpiTransaction createQueryTransaction(Object tenantId) {
-    return transactionManager.createQueryTransaction(tenantId);
+    SpiTransaction t = transactionManager.createQueryTransaction(tenantId);
+    try {
+      transactionScopeManager.set(t);
+    } catch (PersistenceException existingTransactionError) {
+      t.end();
+      throw existingTransactionError;
+    }
+    return t;
   }
 
   /**
--- a/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -810,14 +810,7 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
   @Override
   public Transaction beginTransaction(TxIsolation isolation) {
     // start an explicit transaction
-    SpiTransaction t = transactionManager.createTransaction(true, isolation.getLevel());
-    try {
-      transactionScopeManager.set(t);
-    } catch (PersistenceException existingTransactionError) {
-      t.end();
-      throw existingTransactionError;
-    }
-    return t;
+    return createServerTransaction(true, isolation.getLevel());
   }
 
   /**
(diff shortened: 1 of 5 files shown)


rbygrave:
> an implicit transaction 1 is created

I note the code change doesn't check the explicit boolean.  Even if it did I don't think that would be sufficient for all use cases.

That is, currently we can call ebeanServer.createTransaction() ... and create a transaction outside of transactionScopeManager (No ThreadLocal). We need to do that in some cases (MySql) and it is the intended behaviour of ebeanServer.createTransaction() that it is outside of ThreadLocal / transactionScopeManager management.

So I do think we need another solution here.

rbygrave:
DefaultServer initTransIfRequired() ... might be an option - covers all the persist cases I believe.  DefaultBeanLoader line 224 is one thing we need to watch.

rbygrave:
My gut says it might be good to change SpiEbeanServer API from:
```java
  SpiTransaction createServerTransaction(boolean isExplicit, int isolationLevel);
```

To:
```java
   // explicit
  SpiTransaction createServerTransaction(int isolationLevel);

  // implicit
  SpiTransaction createServerTransactionImplicit();
```

src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -2130,12 +2130,26 @@ TransWrapper initTransIfRequired(Transaction t) {
 
   @Override
   public SpiTransaction createServerTransaction(boolean isExplicit, int isolationLevel) {
-    return transactionManager.createTransaction(isExplicit, isolationLevel);
+    SpiTransaction t = transactionManager.createTransaction(isExplicit, isolationLevel);
+    try {
```
rbygrave: if (!isExplicit) {  // ... around the try

src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -2130,12 +2130,26 @@ TransWrapper initTransIfRequired(Transaction t) {
...
+    }
+    return t;
   }
 
   @Override
   public SpiTransaction createQueryTransaction(Object tenantId) {
-    return transactionManager.createQueryTransaction(tenantId);
+    SpiTransaction t = transactionManager.createQueryTransaction(tenantId);
+    try {
+      transactionScopeManager.set(t);
```
rbygrave: This breaks MySql in theory ... DefaultBeanLoader line 224

rPraml:
Hello @rbygrave ,

here is a small update of the PR - which unfortunately breaks one test. I have to say that I do not know exactly how the whole transaction stuff should work, so I need some help here 

So I try to summarize what I've understood so far
- threre is a ThreadLocal, that holds the current (Scoped)Transaction
- there can only be one transaction in this ThreadLocal
- to support nested transactions, the ScopeTrans has a `suspended` property that replaces the threadLocal when the transaction ends
- When there is no current transaction, all persist and lazy-load functions create an implicit transaction (implicit means, that it is not registered in the threadLocal?)
- If this happens, I've no access to the current transaction in the lifecycle methods.

My exact use case is that I want to update dependent beans (like a trigger) in an @PostPersist/Update method. This must happen in the same transaction, otherwise I run in a dead lock, as the SQL-Server waits until the first transaction is closed.

rbygrave:
FYI: I have some thoughts which I will try to work through tomorrow night.

rbygrave:
Getting close via - https://github.com/ebean-orm/ebean/tree/feature/begin-server-transaction-1166

rPraml:
Closing this. I tested your changes in #1175 today - it covers my use case. Current transaction is avaliable in all persistence lifecycle handlers (but not in @PostLoad)

rbygrave:
Yeah cool.

> @postload

So the pain / issue here is MySql specific behavior (MySql not allowing the
use of a single Connection when executing "query join" queries on a
findEach/findIterator query).  So I need to check the MySql behavior to see
if they fixed that ... but yeah, that is the issue we need to get around in
order to get it nice and correct for @postload

PR #1168, opened 2017-10-13
Code cleanup

KoenDG:
Some things the analyzer found.

* imports

1 new regex that can be statically compiled

some lists that the length can be determined beforehand


rbygrave:
Thanks.

PR #1177, opened 2017-10-18
Improved the performance of findAnnotation

rPraml:
Hello Rob,
We have > 100 models in our application. The boot time of the server took ~8-9 sec. I profiled the application and noticed, that a lot of time (~25%) is consumed in the findAnnotation function. This patch optimizes the performance and the boot up time is now ~7-8 sec.


rbygrave:
Good stuff.

PR #1178, opened 2017-10-18
Fix / support JSON marshalling of BeanChange + add changeLogAsync option to ServerConfig

rPraml:
This PR fixes deserialization of the BeanChange and adds the ability to write changelog synchronious - and the test can detect failures


rbygrave:
changeLogAsync defaults to false ... so I'll ponder if that is the correct / desired default.  The plugin I want / plan to have pushes the change logs into elasticsearch.

rPraml:
I think you're right, changeLogAsync should be initialized with true as default value.

PR #1201, opened 2017-11-02
MultiValueBind improvements

rPraml:
This PR 
- adds MultiValueBind for H2 (although there is no parameter limit in H2)
- determines if multi-id-value is supported differently
- changes the DocQueryContext::writeIn method (breaking-api)
- adds support for 'null' in "in-values" (this may break your code, especially `notIn(1,2,3)` will return everything, except 1,2,3 - also `null`)
- duplicate "in-values" are eliminated. `in(1,1,2,3,3)` will produce the same query & queryplan as `in(1,2,3)`

### DIFF
--- a/src/main/java/io/ebeaninternal/server/expression/DocQueryContext.java
+++ b/src/main/java/io/ebeaninternal/server/expression/DocQueryContext.java
@@ -61,7 +61,7 @@ public interface DocQueryContext {
   /**
    * Write an In expression.
    */
-  void writeIn(String propertyName, Object[] values, boolean not) throws IOException;
+  void writeIn(String propertyName, Object[] values, boolean containsNull, boolean not) throws IOException;
 
   /**
    * Write an Id in expression.
--- a/src/main/java/io/ebeaninternal/server/expression/InExpression.java
+++ b/src/main/java/io/ebeaninternal/server/expression/InExpression.java
@@ -11,8 +11,10 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 class InExpression extends AbstractExpression {
 
@@ -22,6 +24,7 @@ class InExpression extends AbstractExpression {
 
   private Object[] bindValues;
 
+  private boolean containsNull;
   private boolean multiValueSupported;
 
   InExpression(String propertyName, Collection<?> sourceValues, boolean not) {
@@ -36,17 +39,18 @@ class InExpression extends AbstractExpression {
     this.not = not;
   }
 
-  private Object[] values() {
-    List<Object> vals = new ArrayList<>(sourceValues.size());
+  private void prepareBindValues() {
+    Set<Object> vals = new HashSet<>(sourceValues.size());
     for (Object sourceValue : sourceValues) {
       NamedParamHelp.valueAdd(vals, sourceValue);
     }
-    return vals.toArray();
+    containsNull = vals.remove(null);
+    bindValues = vals.toArray();
   }
 
   @Override
   public void prepareExpression(BeanQueryRequest<?> request) {
-    bindValues = values();
+    prepareBindValues();
     if (bindValues.length > 0) {
       multiValueSupported = request.isMultiValueSupported((bindValues[0]).getClass());
     }
@@ -54,7 +58,8 @@ class InExpression extends AbstractExpression {
 
   @Override
   public void writeDocQuery(DocQueryContext context) throws IOException {
-    context.writeIn(propName, values(), not);
+    prepareBindValues();
+    context.writeIn(propName, bindValues, containsNull, not);
   }
 
   @Override
@@ -90,8 +95,13 @@ class InExpression extends AbstractExpression {
   public void addSql(SpiExpressionRequest request) {
 
     if (bindValues.length == 0) {
-      String expr = not ? "1=1" : "1=0";
-      request.append(expr);
+      if (containsNull) {
+        String expr = not ? " is not null" : " is null";
+        request.append(propName).append(expr);
+      } else {
+        String expr = not ? "1=1" : "1=0";
+        request.append(expr);
+      }
       return;
     }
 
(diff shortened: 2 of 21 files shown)


src/main/java/io/ebeaninternal/server/expression/DocQueryContext.java
```diff
@@ -61,7 +61,7 @@
   /**
    * Write an In expression.
    */
-  void writeIn(String propertyName, Object[] values, boolean not) throws IOException;
+  void writeIn(String propertyName, Object[] values, boolean containsNull, boolean not) throws IOException;
```
rPraml: Here I've changed the DocStrore API!

src/main/java/io/ebeaninternal/server/expression/InExpression.java
```diff
@@ -100,15 +110,32 @@ public void addSql(SpiExpressionRequest request) {
...
     }
+    // if we perform an "in" query (not = false) and we want "null's",
+    // we must append an additional OR statement.
+    // if we perform a "not in" query (not = true) and we have not nulls
+    // in our values list, we must include nulls.
+    // Consider, the db has the values 1,2,3,4,null:
+    // in(1,2)          => 1,2
+    // in(1,2,null)     => 1,2,null  (normal SQL would return 1,2 only)
+    // not in(1,2)      => 3,4,null  (normal SQL would return 3,4 only)
+    // not in(1,2,null) => 3,4       (normal SQL would return 3,4 only)
```
rPraml: I hope this will not break too much...
my argument is, that `.in(null)` / `notIn(null)` should be equivalent to `.eq(null)` / `.ne(null)`
For eq/ne there is already code, that handles the null case

dastultz:
On Thu, Nov 2, 2017 at 6:46 PM, Roland Praml <notifications@github.com>
wrote:

>
>
>    - duplicate "in-values" are eliminated. in(1,1,2,3,3) will produce the
>    same query & queryplan as in(1,2,3)
>
>
One of the things Ebean does, I believe, is "pad" a list to some boundary,
I don't know what it is, say 5 items. So a list of integers, (1, 2, 3, 4,
5) would produce a query with (?, ?, ?, ?, ?). If you have 8 items in the
list it pad to 10 (or whatever the next boundary is). So a list of (1, 2,
3) would also produce (?, ?, ?, ?, ?) with bind values of (1, 2, 3, 3, 3).
This results in fewer query plans as there are fewer variations in the
query text.

Is this related to the issue you are addressing? I would not want this
"padding" feature to go away as it helps us manage memory.

/Daryl

rPraml:
I wonder if I can ever get a successful build on the first commt ;)
but @rbygrave I think there is the plugin missing in the maven repo (or caching issue?):
`[ERROR] Plugin io.ebean:ebean-maven-plugin:11.4.1 or one of its dependencies could not be resolved:
Failure to find io.ebean:ebean-maven-plugin:jar:11.4.1 in 
https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]`

rPraml:
@dastultz I did not see any code, that does padding (but it is a good idea). What MultiValueBinding does, is to produce queries like this
- postgres: `SELECT * from table where id = any(?)`
- h2:  `SELECT * from table where id in (SELECT * FROM TABLE(X =?)`
- mssql: `SELECT * from table where id in (SELECT * FROM ?)`
- oracle: `SELECT * from table where id in (SELECT * FROM TABLE (SELECT ? FROM DUAL))`

so these queries accept an array (or platform specifific datatype) for the placeholder.
This solves two issues:
- we can cache the same query for any number of parameters
- we do not run in parameter limit of prepared statement (mssql: max 2100, oracle: max 1000, pg: max 2^16)

Back to your idea. I find the idea with padding very interesting. @rbygrave what do you think, if we allow only in queries with a multiple of 4 (or any other) if we cannot use the MultiValueBind (for MySql e.g.)
So, ebean would use the following prepared statement `SELECT * from table where id in (?, ?, ?, ?)` for `.in` queries with 1..4 parameters (and 8 placeholders for 5..8 parameters)

dastultz:
On Fri, Nov 3, 2017 at 6:27 AM, Roland Praml <notifications@github.com>
wrote:

>
>
>
> so these queries accept an array (or platform specifific datatype) for the
> placeholder.
>
Oh, that's better! I have a thing I just did that imitates the padding,
I'll see if I can figure out how to bind an array.


/Daryl

rbygrave:
For Postgres binding using arrays with ALL() and ANY() ... Is in master and
will be in the 11.4.1 release.

I'll not merge in any others until we check the query plans first (and it
makes sense to keep H2 using the standard binding for a while to keep
coverage on both approaches).

rbygrave:
Also note that I am not sure about null checking part for IN.  That seems
very unusual to me.

That means I am not keen to merge these changes.

Cheers, Rob.

On 4 Nov 2017 8:34 p.m., "Rob Bygrave" <robin.bygrave@gmail.com> wrote:

For Postgres binding using arrays with ALL() and ANY() ... Is in master and
will be in the 11.4.1 release.

I'll not merge in any others until we check the query plans first (and it
makes sense to keep H2 using the standard binding for a while to keep
coverage on both approaches).

rPraml:
> That means I am not keen to merge these changes.

I share your concerns. I don't see that adding "null" support to "in" will be a problem. But in/notIn should behave symmetric, so we must implement "null" support in both functions - which means, that the result that "notIn" produces, will change.

I suggest
- either to change in/notIn to behave equals to eq/ne. So "in(null)" returns the same as "eq(null)" and "notIn(null)" to "ne(null)" (this is what this PR does) and add a big warning in changelog, that you have to check all "notIn" queries. 
- or specify that "null" must not passed to in/notIn and must handled in your query separately - i.e. throw an IllegalArgumentException when one of the arguments is null and add a small warning in the changeLog that all "in/notIn" queries should be checked, if one argument could be null, because it will fail with an exception now (before: result was not like expected)

In my application, I had the query "value.in(1,2,null)" and was wondering, why I did only get 1 and 2, but no "null" - because the error that null is not working was supressed silently. (this was not what I've expected - an exception would at least point me to the error)

rbygrave:
> throw an IllegalArgumentException when one of the arguments

I think that is the right thing to do at this stage.  I'm happy to re-consider this down the track ... it is just that null in a IN clause is quite different thinking for me given how special null is in SQL.

rPraml:
Most of this PR is merged, so closing...

PR #1209, opened 2017-11-14
Suggestions from intellij code analysis.

KoenDG:
Hello Rob and other readers.

I had the intellij analyzer run over the code base once more. This PR is mostly a bit of cleanup, as I found a bunch of Enums being compared with `equals` instead of just `==` like you do with enums.

Since enums can't override `equals()`, there should not be a danger in doing that.

Added a comment in 2 places where I thought it could use some clarification, I'm assuming the comment I made accurately reflects the decision to write the code that way.

And finally, found a place in the code that was using tail recursion, the act of a method calling itself as the last step of execution. It is more performant to use a while-true loop with a modification of the parameters at the place where the recursive call usually happens. This way you're not stepping deeper into a return-chain.

Hope it helps, let me know if something should be reverted.

### DIFF
--- a/src/main/java/io/ebean/config/EncryptKeyManager.java
+++ b/src/main/java/io/ebean/config/EncryptKeyManager.java
@@ -12,7 +12,7 @@ public interface EncryptKeyManager {
    * This gives the EncryptKeyManager the opportunity to get keys etc.
    * </p>
    */
-  default void initialise() {};
+  default void initialise() {}
 
   /**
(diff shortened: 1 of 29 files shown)


rbygrave:
That is all good, comments are accurate etc.  Thanks again, good stuff.

KoenDG:
Thanks and thanks for the ticket rename, feel like a good thing, I'll try and remember to use that anywhere I do a similar thing.

PR #1244, opened 2018-01-30
DiscriminatorColumn length should be 31

ajcamilo:
The JPA specification, states that the DiscriminatorColumn default length should be 31, and not 10.

This is used to specify the length of the column for the `InheritanceType.SINGLE_TABLE` strategy.


rbygrave:
Looks good, thanks.

PR #1264, opened 2018-02-17
[utf8mb4-support] Limit max key len to 767 bytes

rPraml:
Hello Rob,
i'm working on proper utf8-mb4 & collation-bin support - for this, it is neccessary to limit max key size for MariaDB/MySql to 767 bytes (which are <= 191 chars) - otherwise I cannot run the tests.

This PR will only change the size of string properties in test classes to a value lower than 767 bytes and it should be no problem to merge.

Cheers
Roland


rbygrave:
Fair enough.  Note that we don't really need to put `@Size` on a property of a `@View` entity ... that won't have any effect but that was the only thing I noted.
Cheers

PR #1265, opened 2018-02-17
[collation-bin] fixed table and column name spelling to match exact case

rPraml:
Hello Rob,

this PR fixes the exact spelling in @Column and @JoinColumn annotations, so that it matches the spelling in DDL.

**Background**
I'm running tests against a mssql database with *Latin1_General_100_BIN2* collation, which means case sensitive sorting and comparing, also for table and column names :-(

so `SELECT myid FROM ...` will work, `SELECT myId FROM ...` not

I have to say that I am not very happy with this change, a better solution would probably be to honor DbConstraintNaming in @Column / @JoinColumn annotations

### DIFF
--- a/src/test/java/org/tests/compositekeys/db/Item.java
+++ b/src/test/java/org/tests/compositekeys/db/Item.java
@@ -23,10 +23,10 @@ public class Item {
   private int region;
 
   @Embedded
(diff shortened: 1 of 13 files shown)


rPraml:
closing we should not merge this

Issue #1269, opened 2018-02-22
Paging queries should not ALWAYS append an order by ID ... but only when the order by is empty


rbygrave:
There is some background to this issue, refer to #410 

This has come about because currently we ALWAYS add the Id property to the order by for paging queries and actually this can be unnecessary and more expensive that initially thought.

So this change is to not always add the Id property to the order by clause but only do so when the query has no order by clause (and generally developers are expected to add an order by clause for a paging query).

### Steps to reproduce

```java
class Event {

   @Id
   private UUID id;

  @Column
  private Date created;
 ...
}
```
```java
query.order().desc("created").setMaxRows(50).findPagedList()
```
It seems that the generated SQL adds an id to the order by: 
```sql
... order by t0.created desc, t0.id limit 50;  -- this query takes around 1.4 seconds
```
We would like the query to not include the id in the order by because we don't need it in our case:
```sql
... order by t0.created desc limit 50;  -- this query takes around 0.015 seconds
```

PR #1275, opened 2018-02-24
Add ServerConfig.setDefaultOrderById() and query.orderById(boolean) ... to control automatically append "order by id" with paging query

rPraml:
This patch ensures, that paged results are always ordered by id as last column to give a deterministic result.

If persons are "order by name", persons with same name may change their position during paging.

It also fixes the `pagedList.getTotalCount();` which did not work for SqlServer. (a quick hack and may not work on complex queries, with subqueries)

@Rob: I think this code should be moved into the SqlLimiter. Instead of executing a second query with `select count (*) from ....`, you may query the total count on MySql with sql_calc_found_rows / found_rows https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows
and on SqlServer with 'COUNT(*) OVER () '


rbygrave:
So this is not consistent with https://github.com/ebean-orm/ebean/issues/1269

Specifically if the developer specifies an order by we need to trust they have done what they want (and that adding an additional 'order by id' to that is not what we should do).

rPraml:
Hmm. I understand, did not think that this affects performance so much.

The example in #1269 is only correct, if you can ensure, that `created` is unique. If there are two entities with same creation time (which is theoretically possible), the produced pagination result is not deterministic, if no ID column is specified.

I also think that the "fetch joins" may produce the wrong result, if not sorted by id column.

> Specifically if the developer specifies an order by we need to trust they have done what they want 
> (and that adding an additional 'order by id' to that is not what we should do).  

In my use case, I have datagrid, where the end user can sort columns by clicking the headers. I pass the sort arguments directly to ebean. Maybe I've to add the sort logic in my application code

rbygrave:
> this affects performance so much

Well, I suspect it is probably a case of sorting UUID values but even then it is surprising so it probably depends on the specifics of the DB and index.

> the produced pagination result is not deterministic

Yes, in theory we are leaving that up to the developers to decide now.

> I also think that the "fetch joins" may produce the wrong result, if not sorted by id column.

I don't see how that can be - we don't use limit/offset for fetch joins.

> Maybe I've to add the sort logic in my application code

Unless we make it a configuration option that people can turn on ... then yes.

rPraml:
>> I also think that the "fetch joins" may produce the wrong result, if not sorted by id column.
> I don't see how that can be - we don't use limit/offset for fetch joins.
You are right, limit with fetch joins makes no sense.

> Unless we make it a configuration option that people can turn on ... then yes.
I can make a ServerConfig flag `defaultOrderById=true/false` and maybe also a Query setter: orderById(boolean) to override the server setting if you want.

src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
```diff
@@ -254,17 +255,20 @@
...
+
   /**
    * Identity the query for profiling purposes (expected to be unique for a bean type).
    */
   private short profileId;
 
   private ProfileLocation profileLocation;
 
-  public DefaultOrmQuery(BeanDescriptor<T> desc, EbeanServer server, ExpressionFactory expressionFactory) {
+  public DefaultOrmQuery(BeanDescriptor<T> desc, SpiEbeanServer server, ExpressionFactory expressionFactory) {
```
rPraml: I changed this to SpiEbeanServer so that I have access to serverConfig

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -488,6 +488,11 @@
 
   private ProfilingConfig profilingConfig = new ProfilingConfig();
 
+  /**
+   * Controls the default order by id setting of queries. See {@link Query#orderById(boolean)}
+   */
+  private boolean defaultOrderById = false;
```
rbygrave: Maybe this should default to true ... The exception case is then to turn it off.

PR #1277, opened 2018-02-25
ENH: Support adding a '-- Generated by' comment to DDL including version and timestamp

rPraml:
Hello Rob, this PR adds some improvements for dbMigration - which is the first step for upcomming pull requests
- you can specifiy a header for each migration file in ebean.properties. e.g
```
ebean.ddl.header=-- Generated by ebean ${version} at ${timestamp}
```
This may make the work with migration scripts a bit easier

- Ebean prints the current version at start up - this may make debugging / bugreports easier
- Ebean sets identity only if it is not the platform default


rbygrave:
Looks good, thanks.

PR #1285, opened 2018-02-26
No effective code change: Fixwarnings, overrides, generics etc.

rPraml:
There are no code changes

### DIFF
--- a/src/main/java/io/ebean/config/dbplatform/DatabasePlatform.java
+++ b/src/main/java/io/ebean/config/dbplatform/DatabasePlatform.java
@@ -7,7 +7,7 @@ import io.ebean.annotation.Platform;
 import io.ebean.config.CustomDbTypeMapping;
 import io.ebean.config.DbTypeConfig;
 import io.ebean.config.ServerConfig;
-import io.ebeaninternal.util.JdbcClose;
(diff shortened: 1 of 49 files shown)


src/main/java/io/ebean/config/dbplatform/DatabasePlatform.java
```diff
@@ -7,7 +7,7 @@
 import io.ebean.config.CustomDbTypeMapping;
 import io.ebean.config.DbTypeConfig;
 import io.ebean.config.ServerConfig;
-import io.ebeaninternal.util.JdbcClose;
+import io.ebeaninternal.util.JdbcClose; // FIXME: should not import ebeaninternal here
```
rbygrave: Quite right, need to review this ... probably move JdbcClose

src/main/java/io/ebeaninternal/server/type/ArrayElementConverterEnum.java
```diff
@@ -13,7 +13,7 @@ public ArrayElementConverterEnum(ScalarType<?> scalarType, Class<? extends Enum<
 
   @Override
   public String toElement(Object rawValue) {
-
+    // FIXME: Don't understand for what is this.
```
rbygrave: Note to self - review

PR #1292, opened 2018-02-26
Fix for non-native / number boolean types (MySql, SqlServer, Oracle, Sqlite, SqlAnywhere) removing "default 0" + Fix for DB2 add foreign key update clause

rPraml:
Hello Rob, this is my last PR for today ;)

- It will remove the default value for booleans we already start to discuss here: https://groups.google.com/forum/#!topic/ebean/fb4A7hffIaA
- add default value for true/false and now for different platfoms
- fixes for DB2

- currently, it does not work here on my oracle express 11g (it does not support create table as identity, probably would need 12)

I have some questions, how run the tests (I use `mvn verify -Ddatasource.default=...`)
I see you have integrated something with docker? How does this work and what platforms covers your test setup?

cheers
Roland


rbygrave:
> I see you have integrated something with docker? How does this work and 

Yes. It is in the process of being refactored right at the moment. it used to be called 'ebean-docker-run' and only do docker stuff ... but for those who opt in to use it, it is going to "simplify" test configuration ... https://github.com/ebean-orm/ebean-test-config

... and you can pretty much just set `ebean.test.plaform` : `sqlserver` ... and it can use decent defaults, fire up a sqlserver as a docker container, setup the database etc.

> what platforms covers your test setup?

Postgres, MySql, SqlServer & Oracle. 

I need to add support for DB2 and SqlAnywhere (add support for them to the underlying https://github.com/avaje/docker-commands project such that it knows how to create/drop database and user, wait for DB ready state and wait for connectivity).

rbygrave:
Note that the Oracle docker container defaults to 12c ... she is a pretty big image (takes a while to pull) but it is really nice once pulled.

PR #1299, opened 2018-03-01
FIX: @Formula on @ManyToOne produces SQLException ... fix appends missing FROM clause for formula joins

rPraml:
This appends the from join, if the property was a formula.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeBean.java
+++ b/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeBean.java
@@ -531,7 +531,10 @@ class SqlTreeNodeBean implements SqlTreeNode {
    */
   @Override
   public void appendFrom(DbSqlContext ctx, SqlJoinType joinType) {
-
+    if (nodeBeanProp != null && nodeBeanProp.isFormula()) {
+      // add joins for formula beans
+      nodeBeanProp.appendFrom(ctx, joinType);
+    }
     ctx.pushJoin(prefix);
     ctx.pushTableAlias(prefix);
 
--- a/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeExtraJoin.java
+++ b/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeExtraJoin.java
@@ -126,6 +126,10 @@ class SqlTreeNodeExtraJoin implements SqlTreeNode {
       joinType = SqlJoinType.OUTER;
     }
     if (!manyToMany) {
+      if (assocBeanProperty.isFormula()) {
+        // add joins for formula beans 
+        assocBeanProperty.appendFrom(ctx, joinType);
+      }
       joinType = assocBeanProperty.addJoin(joinType, prefix, ctx);
     }
 
--- a/src/test/java/org/tests/family/TestInheritance.java
+++ b/src/test/java/org/tests/family/TestInheritance.java
@@ -219,4 +219,46 @@ public class TestInheritance extends BaseTestCase {
     assertEquals("Munich", grandparent1.getAddress());
     assertEquals(1, grandparent1.getEffectiveBean().getId().intValue());
   }
+
+  @Test
+  public void testFindCount() {
+    EBasic basicA = new EBasic();
+    basicA.setName("FamilyName A");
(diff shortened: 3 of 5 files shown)


rbygrave:
This description is not clear to me.  It seems this is somewhat of an enhancement to support using `@Formula` on non-scalar asscociated beans:

e.g.
```java
  @ManyToOne(optional = true, fetch = FetchType.EAGER)
  @Formula(select = "f3.id", join = "left join e_basic f3 on f3.name = ${ta}.family_name")
  private EBasic basicSameName;  
```

Is that right?


Note that is it important to have good descriptions of what the change actually does.  Specifically we need to be clear if we are bug fixing current behavior versus supporting new behavior - it is pretty important for people reading our PRs and issues such that they can assess the impact to them.
  
In this case, it looks to me like we are adding support for using `@Formula` with `@ManyToOne` ... and we couldn't do that before.

Cheers, Rob.

rbygrave:
The title could be: 

Add ability to use `@Formula` with `@ManyToOne` beans.

---
If I have understood correctly I'll update the title and merge. Cheers, Rob.

rPraml:
Using @Formula with @ManyToOne was already possible before, if formula returns the ID of the bean.
Just reading did work already before (because of lazy loading), but if there is a query, that has a property of the formula bean, it will fail. (So, I think this is more a fix than an enhancement)

This PR fixes the following issues:

1. `server().find(GrandParentPerson.class).where().eq("basicSameName.name", "FamilyName A").findCount()` fails with
```
javax.persistence.PersistenceException: Query threw SQLException:Feld "F3.ID" nicht gefunden
Column "F3.ID" not found; SQL statement:
select count(*) from grand_parent_person t0 left join e_basic t1 on t1.id = f3.id  where t1.name = ?  
```
(Fix in SqlTreeNodeExtraJoin)

2. the complex query in TestInheritance.testInheritance() fails with
```
javax.persistence.PersistenceException: Query threw SQLException:Feld "J1.SOME_BEAN_ID" nicht gefunden
Column "J1.SOME_BEAN_ID" not found; SQL statement:
select t0.identifier, t0.name, t0.age, coalesce(f1.child_count, 0), coalesce(f1.child_age, 0), t0.family_name, t0.address, t0.some_bean_id, f3.id, coalesce(t0.some_bean_id,1), t1.identifier, t2.id, t2.status, t2.name, t2.description, t2.some_date from grand_parent_person t0 left join (select i1.parent_identifier, count(*) as child_count, sum(i1.age) as child_age from parent_person i1 group by i1.parent_identifier) f1 on f1.parent_identifier = t0.identifier  left join e_basic f3 on f3.name = t0.family_name  left join parent_person t1 on t1.parent_identifier = t0.identifier  left join e_basic t2 on t2.id = coalesce(t1.some_bean_id, j1.some_bean_id)  left join e_basic t3 on t3.id = coalesce(t0.some_bean_id,1)  where t0.identifier = ?   and (t3.id in (? )  or t3.id is null )  order by t0.identifier, t1.identifier [42122-196] Bind values:[null] Query was:select t0.identifier, t0.name, t0.age, coalesce(f1.child_count, 0), coalesce(f1.child_age, 0), t0.family_name, t0.address, t0.some_bean_id, f3.id, coalesce(t0.some_bean_id,1), t1.identifier, t2.id, t2.status, t2.name, t2.description, t2.some_date from grand_parent_person t0 left join (select i1.parent_identifier, count(*) as child_count, sum(i1.age) as child_age from parent_person i1 group by i1.parent_identifier) f1 on f1.parent_identifier = t0.identifier  left join e_basic f3 on f3.name = t0.family_name  left join parent_person t1 on t1.parent_identifier = t0.identifier  left join e_basic t2 on t2.id = coalesce(t1.some_bean_id, j1.some_bean_id)  left join e_basic t3 on t3.id = coalesce(t0.some_bean_id,1)  where t0.identifier = ?   and (t3.id in (? )  or t3.id is null )  order by t0.identifier, t1.identifier
```
(Fix in SqlTreeNodeBean)

rbygrave:
Ah I see - awesome, thanks for that clarification.

When the failing test produces an exception ... that is good because readers/followers will know they have not hit the bug / scenario etc - so the clarification above is great - ta.

PR #1300, opened 2018-03-01
NEW: FetchCountDistinct - query can return distinct values with their counts

rPraml:
**Use Case**
Give me the top 10 customer orders.

So we need a way not only to get the distinct values but also the counts in a certain order

### DIFF
--- a/src/main/java/io/ebean/CountedValue.java
+++ b/src/main/java/io/ebean/CountedValue.java
@@ -1,0 +1,33 @@
+package io.ebean;
+
+import java.io.Serializable;
+/**
+ * Holds a distinct value with it's count.
+ * (Used with {@link Query#findSingleAttributeList()} and {@link Query#setCountDistinct(CountDistinctOrder)}.)
+ * @author Roland Praml, FOCONIS AG
+ */
+public class CountedValue<A> implements Serializable {
+  private static final long serialVersionUID = -2267971668356749695L;
+
+  private final A value;
+  private final long count;
+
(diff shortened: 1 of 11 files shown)


src/main/java/io/ebean/CountedValue.java
```diff
@@ -0,0 +1,33 @@
+package io.ebean;
+
+import java.io.Serializable;
+/**
+ * Holds a distinct value with it's count.
+ * (Used with {@link Query#findSingleAttributeList()} and {@link Query#setCountDistinct(CountDistinctOrder)}.)
+ * @author Roland Praml, FOCONIS AG
```
rbygrave: Note to self ... need to add an example in javadoc

rbygrave:
## Example

```java
    List<CountedValue<Order.Status>> orderStatusCount =

      Ebean.find(Order.class)
        .select("status")
        .setCountDistinct(CountDistinctOrder.ATTR_DESC)
        .findSingleAttributeList();

    for (CountedValue<Order.Status> entry : orderStatusCount) {
      System.out.println("orderStatus:" + entry.getValue() + " count:" + entry.getCount());
    }
```

```console
orderStatus:COMPLETE count:1
orderStatus:SHIPPED count:1
orderStatus:NEW count:3
```

And with `.setCountDistinct(CountDistinctOrder.COUNT_DESC_ATTR_ASC)`

```console
orderStatus:NEW count:3
orderStatus:SHIPPED count:1
orderStatus:COMPLETE count:1
```

PR #1301, opened 2018-03-01
ENH: extra-ddl can contribute to "drop-all" script, otherwise the drop all script is not appyable to h2:file

rPraml:
I discovered this problem, when I tried to run the test case against a h2:file database, that the second run could not successfully run the drop-all script.

It's a small enhancement (and not really needed, because I do not use the drop-all script)

### DIFF
--- a/src/main/java/io/ebeaninternal/dbmigration/DefaultDbMigration.java
+++ b/src/main/java/io/ebeaninternal/dbmigration/DefaultDbMigration.java
@@ -268,7 +268,8 @@ public class DefaultDbMigration implements DbMigration {
       if (extraDdl != null) {
         List<DdlScript> ddlScript = extraDdl.getDdlScript();
         for (DdlScript script : ddlScript) {
-          if (ExtraDdlXmlReader.matchPlatform(dbPlatform.getName(), script.getPlatforms())) {
+          if (script.isDrop() == false
+              && ExtraDdlXmlReader.matchPlatform(dbPlatform.getName(), script.getPlatforms())) {
             writeExtraDdl(migrationDir, script);
           }
         }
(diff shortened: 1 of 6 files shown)


src/main/java/io/ebeaninternal/dbmigration/DefaultDbMigration.java
```diff
@@ -268,7 +268,8 @@ private void generateExtraDdl(File migrationDir, DatabasePlatform dbPlatform) th
       if (extraDdl != null) {
         List<DdlScript> ddlScript = extraDdl.getDdlScript();
         for (DdlScript script : ddlScript) {
-          if (ExtraDdlXmlReader.matchPlatform(dbPlatform.getName(), script.getPlatforms())) {
+          if (script.isDrop() == false
```
rbygrave: Note to self - review line format

src/main/java/io/ebeaninternal/extraddl/model/ExtraDdlXmlReader.java
```diff
@@ -20,15 +20,16 @@
...
+  public static String buildExtra(String platformName, boolean drops) {
 
     ExtraDdl read = ExtraDdlXmlReader.read("/extra-ddl.xml");
     if (read == null) {
       return null;
     }
     StringBuilder sb = new StringBuilder(300);
     for (DdlScript script : read.getDdlScript()) {
-      if (matchPlatform(platformName, script.getPlatforms())) {
+      if (script.isDrop() == drops
```
rbygrave: Note to self - review line format

rPraml: What's your line format, line length? Maybe you can add an http://editorconfig.org/ and configure at least
indent_style = space
indent_size = 2
max_line_length = ???

rbygrave: Thanks - set it to 130, ta.

PR #1302, opened 2018-03-01
NEW: method ebean_recalc notifies the bean to perform additional computations

rPraml:
This is a quick hack, we need in our framework, so that we can execute java code if bean is modified. E.g. compute the hash of a field.

It's comparable to a SQL trigger, that can modify fields before the bean was saved.

Unfortunately, there was no suitable lifecycle method, as preperisist is also executed, if the bean is not dirty. And checking, if the bean is dirty did also not work because of the mutable properties, which are checked later.

Maybe you have an other idea, how to solve the issue (should we introduce a 3rd lifecycle, so we have PrePersist, Persisting, PostPersist)


src/main/java/io/ebean/bean/EntityBean.java
```diff
@@ -96,4 +96,9 @@
    */
   Object _ebean_getFieldIntercept(int fieldIndex);
 
+  /**
+   * Recalcs the bean. This is invoked by PersistRequestBean.
+   */
+  default void _ebean_recalc() {};
```
rbygrave: This is almost '_ebean_prePersistDirty' ?

We ought to javadoc the difference with normal PrePersist ...

rbygrave:
> should we introduce a 3rd lifecycle

That seems like a pretty good description ... add 'Persisting' lifecycle hook.

I'm not against having the method ... but maybe with a better method name? and with some javadoc explaining when it's used relative to PrePersist.

Yeah, I'll come back to this in a bit.

src/main/java/io/ebean/bean/EntityBean.java
rbygrave: How about `_ebean_onPersistTrigger()`

rbygrave:
> as preperisist is also executed, if the bean is not dirty. And checking

Is PrePersist just firing too early then? ... only fire on dirty, fire after mutable properties are checked.

rPraml:
I have to check again, why prePersist did not work. (I'm sure we tried this, but I do not remember what was the problem)
I also would agree to rename the method.

rbygrave:
I think we should close this.  Re-open with more information re PrePersist etc.

PR #1303, opened 2018-03-01
NEW: Ebean.checkUniqueness - you can check a bean, if save would work

rPraml:
**Use Case**
We have a model with `firstName`, `lastName` and `emailAddress`.
On firstName and lastName is a compound unique index. On email address also.

So, if I insert the user "Roland", "Praml" twice, I will get a DuplicateKey (or DataIntegrityException, depends on DB platform) with a cryptic error message like "cannot insert .... because UQ_FIRST_LAST..."
I cannot show that error message to the end user and it is nearly impossible to parse out the properties to show a message "The values 'firstName' and 'lastName' must be unique)"

To solve this, I have to check first, if the model is already in the database and show to the end user what is wrong.

For this I've added the checkUniqueness method, that checks the model against all unique indices and report the properties, that violates the index.

**Drawbacks**
This costs additional performance and there is a certain risk, that between check and save a model is created somewhere else.

### DIFF
--- a/src/main/java/io/ebean/Ebean.java
+++ b/src/main/java/io/ebean/Ebean.java
@@ -3,11 +3,13 @@ package io.ebean;
 import io.ebean.annotation.TxIsolation;
 import io.ebean.cache.ServerCacheManager;
 import io.ebean.config.ServerConfig;
+import io.ebean.plugin.Property;
 import io.ebean.text.csv.CsvReader;
 import io.ebean.text.json.JsonContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.persistence.OptimisticLockException;
 import javax.persistence.PersistenceException;
@@ -15,6 +17,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -116,7 +119,7 @@ public final class Ebean {
   static {
     EbeanVersion.getVersion(); // initalizes the version class and logs the version.
   }
- 
+
   /**
    * Manages creation and cache of EbeanServers.
    */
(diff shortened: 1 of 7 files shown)


rbygrave:
Wow - I just think this is such a nice little addition ... very useful, very cool.

src/main/java/io/ebean/Ebean.java
```diff
@@ -641,6 +644,30 @@ public static int saveAll(Collection<?> beans) throws OptimisticLockException {
     return serverMgr.getDefaultServer().saveAll(beans);
   }
 
+  /**
+   * This method checks the uniqueness of a bean. I.e. if the save will work. It will return the
+   * properties that violates an unique / primary key. This may be done in an UI save action to
+   * validate if the user has entered correct values.
+   *
+   * Note: This method queries the DB for uniqueness of all indices, so do not use it in a batch update.
+   *
```
rbygrave: Note to self ... add an example use into the javadoc

src/main/java/io/ebean/EbeanServer.java
```diff
@@ -1461,6 +1462,26 @@
    */
   int saveAll(Collection<?> beans, Transaction transaction) throws OptimisticLockException;
 
+  /**
+   * This method checks the uniqueness of a bean. I.e. if the save will work. It will return the
+   * properties that violates an unique / primary key. This may be done in an UI save action to 
+   * validate if the user has entered correct values.
+   * 
+   * Note: This method queries the DB for uniqueness of all indices, so do not use it in a batch update.
+   *
```
rbygrave: Add an example into javadoc ...

rbygrave:
## Example

```java

      // there is a unique constraint on title

      Document doc = new Document();
      doc.setTitle("One flew over the cuckoo's nest");
      doc.setBody("clashes with doc1");

      Set<Property> properties = server().checkUniqueness(doc);

      if (properties.isEmpty()) {
        // it is unique ... carry on
      } else {
        // build a user friendly message
        // to return message back to user

        String uniqueProperties = properties.toString();

        StringBuilder msg = new StringBuilder();

        properties.forEach((it)-> {
          Object propertyValue = it.getVal(doc2);
          String propertyName = it.getName();
          msg.append(" property["+propertyName+"] value["+propertyValue+"]");
        });

        System.out.println("uniqueProperties > "+uniqueProperties);
        System.out.println("      custom msg > " + msg.toString());

      }
```

```console
uniqueProperties > [title]
      custom msg >  property[title] value[One flew over the cuckoo's nest]
```

vnnv:
Very very cool :) need that thing last week :)

PR #1305, opened 2018-03-01
NEW: server.autostart can be disabled, so no migration is run

rPraml:
This is mainly to control executeDdl (so it would be OK if we rename this)

### DIFF
--- a/src/main/java/io/ebean/EbeanServer.java
+++ b/src/main/java/io/ebean/EbeanServer.java
@@ -1850,4 +1850,10 @@ public interface EbeanServer {
    * </p>
    */
   <T> Set<String> validateQuery(Query<T> query);
+
+  /**
+   * Execute the DDL generator manually (this makes only sense if {@link ServerConfig#isAutostart() is false)
+   * @param online
+   */
+  void executeDdlGenerator(boolean online);
 }
--- a/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -286,14 +286,17 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
     }
   }
 
+  @Override
+  public void executeDdlGenerator(boolean online) {
+    if (!serverConfig.isDocStoreOnly()) {
+      ddlGenerator.execute(online);
+    }
+  }
   /**
    * Execute all the plugins with an online flag indicating the DB is up or not.
    */
   public void executePlugins(boolean online) {
 
-    if (!serverConfig.isDocStoreOnly()) {
-      ddlGenerator.execute(online);
-    }
     for (Plugin plugin : serverPlugins) {
       plugin.online(online);
     }
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -368,6 +368,11 @@ public class ServerConfig {
   private boolean updatesDeleteMissingChildren = true;
 
   /**
+   * Should the server start all
+   */
+  private boolean autostart = true;
+
+  /**
    * Database type configuration.
    */
   private DbTypeConfig dbTypeConfig = new DbTypeConfig();
@@ -2230,6 +2235,14 @@ public class ServerConfig {
     this.updatesDeleteMissingChildren = updatesDeleteMissingChildren;
   }
 
+  public boolean isAutostart() {
+    return autostart;
+  }
+
+  public void setAutostart(boolean autostart) {
+    this.autostart = autostart;
+  }
+
   /**
    * Return true if the ebeanServer should collection query statistics by ObjectGraphNode.
    */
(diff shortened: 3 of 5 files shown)


rbygrave:
What is the use case this is used for?

rPraml:
I'll check if the autostart flag is really needed.

What I definitively need is the ability to run DDL migration again.

We have built a multi tenat application that works with DB_WITH_MASTER multi tenant mode (this is a kind of mixture between NONE and DB)

If there is no tenant set in the currentTeant variable(= null / root tenant), everything should behave like in NONE mode. that means, we use the DB that is configured in the ebean.properties.

When our application boots, it will access this master db at first and read the tenant-configuration from this database (connection string/user/password)
In the next step, it tells our TenantDataSourceProvider the connection settings, so that tenant switching works from this point.
Then it switches to each tenant and run dbmigration.

Pseudocode;
```
ThreadLocal currentTenant= new ThreadLocal();
MyMultitenantDataSourceProvider dsProvider = ...
config.setCurrentTenantProvider(currentTenant::get)
config.setTenantDataSourceProvider(dsProvider);
server = EbeanServerFactory.create(serverConfig);

tenants = server.find(TenantConfigEntity.class).findList(); // contains connection string
dsProvider.setTenants(tenants); // will create internally a datasourcePool for each tenant

for (tenant : tenants) {
   currentTenant.set(tenant.getId);
   server.executeDdlGenerator(true); // will create or update schema of each tenant.
}
```

rPraml:
The use case is:

I need a working ebean server, if the database is not yet up. 

So, we construct an ebean server, that does not start (i.e.try to access the database) 

Then we execute server.start & ddl run in a try catch block, which may fail, if DB is not up.

If so, we defer ddl run in the onDatasourceUp of the DataSourceAlert listener

rbygrave:
I must admit to not understanding the use case.

> I need a working ebean server, if the database is not yet up.

Why? For me from a Kubernetes perspective especially but in general we don't start the service if the DB service isn't up and ready - we fail early.

From a testing perspective we don't have a problem here either if we use ebean-test-config or a similar approach (that blocks until all the required test containers are ready for DDL) ... so why?

rPraml:
We have a spring-boot web application running in a tomcat container. Initializing of the ebean-server is done in a very early stage. The server-"bean" is needed on several places. So, if the server does not come up at application start, the whole application would not start at all.
This behaviour is not consistent, as the DB can go down after application start (.isDatasourceUp will be false then)

On an incomming request, we check, if datasource is up and if not, we return an error page.

So, if application starts and db goes down later (application runs) - a dataSourceDown alert happens.
If you restart the application now (datasource is still down) you cannot, because startup will fail as the ebean server needs a working state at start up.

I also have to adjust the datasource implementation to archive this: https://github.com/FOCONIS/avaje-datasource/commit/8be8a1beeb2d77dbe9eac33e89b090ad73341ed9 (have to create a PR for this, as the autostart option in ebean is useless without this change)

It would be OK for me to maintain that change in my branch or keep back that PR for later, if you think that this is an edge use case

rbygrave:
Right, ok I think I get it.  Roughly ...
- Expose the ability to run DDL scripts  ... yup, cool
- Ability to bring up EbeanServer instance without an "active DataSource" (aka initially offline) ... yup, cool.

Reviewing again.  Thinking "autostart = false" means allow EbeanServer to come up offline (without active DataSource).

Real life is always messier and harder that "the theory" and yes I can see the need.  Looking ...

src/main/java/io/ebean/EbeanServer.java
```diff
@@ -1850,4 +1850,10 @@
    * </p>
    */
   <T> Set<String> validateQuery(Query<T> query);
+
+  /**
+   * Execute the DDL generator manually (this makes only sense if {@link ServerConfig#isAutostart() is false)
+   * @param online
+   */
+  void executeDdlGenerator(boolean online);
```
rbygrave: So it seems to me that you'd want to "run migrations" rather than "generate and run the create-all.sql" ?

That is, create-all.sql ... generally aimed for testing and "db migrations" for production.

rbygrave: Should this not instead be one of:
- runMigrations()
- start()

rPraml: Yes, runMigrations would be a better naming.

Maybe autostart can be renamed to offline, too. So would be the application's task to create server,  try to run migration and start the server afterwards.

rbygrave: Yeah, it is almost `offline` and `startServer()` ... with start server usually meaning run migrations but that depends on ServerConfig. Maybe startServer() .. is a better more generic name.

Bring up EbeanServer "offine" ... then sometime later ... ebeanServer.startServer() ... which maybe runs migrations etc (stuff that is normally done when the server starts).

src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -443,6 +454,8 @@ public ReadAuditLogger getReadAuditLogger() {
    * Run any initialisation required before registering with the ClusterManager.
    */
   public void initialise() {
+    ddlGenerator.generateDdl();
```
rPraml: generation is done in initialize now, run in start()/runDbMigration()

rPraml:
@rbygrave I have rewritten this. The autostart flag is gone, so the ebean server tries to start up also if DB is down.

This PR depends on https://github.com/ebean-orm/ebean/pull/1597

I will check in our application, if the current implementation will fit our needs.

rbygrave:
Looking to close this PR.

PR #1306, opened 2018-03-01
NEW: flag to control, if javax.validation NotNull annotation should generate a NotNull column or not

rPraml:
This controls the behaviour of the javax.validation.constraints.NotNull annotation, which we use for bean validation but do not want to use for DDL not null generation.

We have the concrete problem, when we alter a table and add additional columns, that the operation will fail, when the column is `not null` in DDL

### DIFF
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -470,6 +470,14 @@ public class ServerConfig {
    */
   private boolean disableL2Cache;
 
+
+  /**
+   * Should the javax.validation.constraints.NotNull enforce a notNull column in DB.
+   * If set to false, you have to use io.ebean.annotation.NotNull to explicityl create
+   * a not null column.
+   */
+  private boolean useJavaxValidationNotNull = true;
+
   /**
    * Generally we want to perform L2 cache notification in the background and not impact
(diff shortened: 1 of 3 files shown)


src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -470,6 +470,14 @@
    */
   private boolean disableL2Cache;
 
+
+  /**
+   * Should the javax.validation.constraints.NotNull enforce a notNull column in DB.
+   * If set to false, you have to use io.ebean.annotation.NotNull to explicityl create
```
rbygrave: javadoc ... explicityl

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -2959,6 +2968,27 @@ public void setDisableL2Cache(boolean disableL2Cache) {
...
 
+  /**
+   * Returns if we use javax.validation.constraints.NotNull
+   */
+  public boolean isUseJavaxValidationNotNull() {
+    return useJavaxValidationNotNull;
+  }
+
+  /**
+   * Controlws when Ebean should generate a <code>NOT NULL</code> column.
```
rbygrave: javadoc .. Controlws

PR #1307, opened 2018-03-01
NEW: Json supports migration and enforcing to write reference beans only

rPraml:
**Use Case**
What do you do, if dataModel of a @DbJson model changes? e.g. jon the fields firstname + lastname to name. 
Do you write migration scripts in SQL that will handle the update.

So this PR adds version information in the json and also adds functionality, that you can migrate from an older json structure in database to a newer one

### DIFF
--- a/src/main/java/io/ebean/text/json/JsonReadOptions.java
+++ b/src/main/java/io/ebean/text/json/JsonReadOptions.java
@@ -25,6 +25,8 @@ public class JsonReadOptions {
 
   protected Object loadContext;
 
+  private JsonVersionMigrationHandler versionMigrationHandler;
+
   /**
    * Default constructor.
    */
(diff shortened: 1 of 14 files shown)


rbygrave:
>  Do you write migration scripts in SQL that will handle the update.

So to date ... my JSON columns have DTO beans ... and generally I'd expect to write code findEach() and load and mutate the document as desired.

My concern here is that JsonVersionMigrationHandler uses pretty low level API (not public ReadJson) ... and with this type of thing perhaps we could look at an 'expression language' to read and mutate JSON ala XPath.

I'm not 100% convinced by JsonVersionMigrationHandler yet. Thinking ...

rPraml:
My needs for this use case are
- be prepared for migration, so write bean version information from the beginning
- can convert from one version to the next 

If you do not want to expose the low level API to the user, It would be also OK, if I get a Map that represents the current JSON and modify that map.

I'm also thinking, if we can/should integrate this into the beans, as it is done by java serialization: 


```
class MyJsonBean {
  public static final int JSON_VERSION_UID=23

  public static int jsonMigrateFrom_22(Map current) {
    // do the modification in the map (inplace) 
    return 23;
  } 
} 
```

So, if ebean reads a json with version 22,it will invoke the method by reflection and the method can perform the migration steps. After this, it returns the version, where it has migrated to.

rPraml:
Still in progress

rPraml:
@rbygrave I've updated this PR and reduced the neccessary API changes a bit. It adds a writer to the jsonWriteOptions that can write an additional version field and a MigrationHandler, that can manipulate the ObjectNode before ebean tries to create a bean. (this is also a little drawback, if such a handler is set, we have to create always an ObjectNode and cannot use the jackson streaming features)

In our application, we have static methods in the models, that perform the migration steps. I also added some demo/test code to see this in action.

PR #1309, opened 2018-03-01
NEW: UUID version 1 generator

rPraml:
This adds an UUID version 1 generator. In conjunction with #1308 you can store UUIDs in a very efficient way


rbygrave:
Another very nice addition - very cool.

PR #1310, opened 2018-03-01
NEW: CustomDeployParser to add your own Formula annotations

rPraml:
We use this in our application, to implement own annotations (see test case)
This prevents us for writing complex @Formula annotations.

Started Introduce a Descriptor API, so I do not need to access the ebeaninternal in my application


rPraml:
blocked by other PR will investigate build errors later

PR #1317, opened 2018-03-01
FIX: limit the constraint names

rPraml:
The configured `maxConstraintNameLength` has to be honored for each platform separately.

@Rob Maybe this makes DbConstraintNaming.MaxLength obsolete, or I did not completely understand how this should work. If I set a max constraint name, this applies to all platforms?

My use case is, that I want to generate migration scripts for my target databases, which are currently sqlserver and mysql and maybe in future db2/oracle


rbygrave:
This change / fix looks correct to me.

I believe this has come about because originally ... the intention was that for constraint names in the 'orm logical model' and the 'db physical model' would **match** (like we do with table names).

However, I believe that approach has problems when we target many DB's and perhaps don't know all the target DB's initially.  For example, we start with Postgres and MySql ... then later on target DB2 - our logical model (in *.model.xml effectively) will have constraint names that are too long for DB2.

So as I see it this fix is correct in that ... we should allow the 'orm logical model' (in *.model.xml) to have relatively large constraint names that might be too long for some target platforms (well, DB2 especially)  but then when we generate platform specific DDL we can truncate them down to suit DB2.

I suspect what we should do is allow the 'logical constraint name' to be relatively large (not restricted by the platform limit and appear in *.model.xml as such) ... and then restrict on platform at DDL generation time.

I believe Oracle increased their constraint name limit (need to check that) ... so in theory we should all be good and match for everything but DB2 ... and then truncate at DDL there.

src/main/java/io/ebeaninternal/server/rawsql/DRawSqlService.java
```diff
@@ -37,7 +37,7 @@ public RawSqlBuilder unparsed(String sql) {
   public SqlRow sqlRow(ResultSet resultSet, String dbTrueValue) throws SQLException {
     ResultSetMetaData meta = resultSet.getMetaData();
     int estCap = (int) (meta.getColumnCount() / 0.7f) + 1;
-    DefaultSqlRow ret = new DefaultSqlRow(estCap, 0.75f, dbTrueValue);
+    DefaultSqlRow ret = new DefaultSqlRow(estCap, 0.75f, dbTrueValue, false);
```
rbygrave: Note to self ... this method takes binaryOptimizedUUID ... so we use that and not false here.

PR #1318, opened 2018-03-02
ENH: native UUID-support for SqlServer

rPraml:
This adds UUID support for SqlServer.
This may break things, if UUIDs were already used and migrating to this ebean version (May require to set dbuuid=VARCHAR explicitly)

### DIFF
--- a/src/main/java/io/ebeaninternal/server/type/ScalarTypeUUIDNative.java
+++ b/src/main/java/io/ebeaninternal/server/type/ScalarTypeUUIDNative.java
@@ -20,6 +20,9 @@ public class ScalarTypeUUIDNative extends ScalarTypeUUIDBase {
     if (value == null) {
       return null;
     }
+    if (value instanceof String) {
+      return UUID.fromString((String) value);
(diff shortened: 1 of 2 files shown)


src/main/java/io/ebeaninternal/server/type/ScalarTypeUUIDNative.java
```diff
@@ -20,6 +20,9 @@ public UUID read(DataReader dataReader) throws SQLException {
     if (value == null) {
       return null;
     }
+    if (value instanceof String) {
```
rPraml: The driver can write UUIDs, but returns it as string

rbygrave:
So yes, I think we need to split SqlServerPlatform into 2 platforms first.  Allow people to use SqlServerLegacyPlatform with no changes ... and then yes this could be merged to a "new" SqlServerPlatform that people would explicitly choose to use.

rbygrave:
Thanks Roland, we have got this one in.  Nice.

PR #1320, opened 2018-03-02
FIX: only visit DDL-columns

rPraml:
**wait** - I have to check this, if this is still needed

Note for myself: check what was the problem and does it still exist in our app


rPraml:
fixed by #1453

PR #1321, opened 2018-03-02
New fetchproperties

rPraml:
This adds a feature, that you can pass a set of el-paths to the query and it will fetch these properties automatically.

We use this in our datagrid, where the end user can select, which el-paths he wants to see in the view

### DIFF
--- a/src/main/java/io/ebean/Query.java
+++ b/src/main/java/io/ebean/Query.java
@@ -1525,4 +1525,9 @@ public interface Query<T> {
    */
   Query<T> orderById(boolean orderById);
 
+  /**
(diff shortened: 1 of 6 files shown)


rbygrave:
Ummm ... why didn't you use "fetchPaths" ? ... query.apply(fetchPaths) ??

rPraml:
Yes, this might be an option, but I need a way to convert a set of el-paths (as string) into a PathProperties object (can do that in an utility class in my application) - so I investigate, if I can use this in my application.

rPraml:
thanks @rbygrave  for pointing this out. query.apply(fetchPath) is a good working alternative. So closing this

PR #1324, opened 2018-03-02
Refactor dbconfig for multiple platforms

rPraml:
You can now specify different options when you generate multi platform migrationscripts

(e.g enforce binary uuid for SqlServer, but user native for PostgreSQL)

Supported options:
- `dbmigration.platform.<PLATFORM>.databaseSequenceBatchSize`
- `dbmigration.platform.<PLATFORM>.dbuuid`
- `dbmigration.platform.<PLATFORM>.uuidStoreAsBinary`
- `dbmigration.platform.<PLATFORM>.geometrySRID`
- `dbmigration.platform.<PLATFORM>.idType`
- `dbmigration.platform.<PLATFORM>.mapping`


src/main/java/io/ebean/config/PlatformConfig.java
```diff
@@ -0,0 +1,282 @@
...
+
+import io.ebean.annotation.Platform;
+import io.ebean.config.dbplatform.DbType;
+import io.ebean.config.dbplatform.IdType;
+import io.ebean.util.StringHelper;
+
+/**
+ * Configuration for Platforms such as UUID, Geometry etc.
+ */
+public class PlatformConfig {
```
rbygrave: So this is a rename plus adds allQuotedIdentifiers and databaseSequenceBatchSize ...

src/main/java/io/ebean/config/PlatformConfig.java
```diff
@@ -0,0 +1,282 @@
...
+   */
+  public List<CustomDbTypeMapping> getCustomTypeMappings() {
+    return customDbTypeMappings;
+  }
+
+  /**
+   * @param p
+   * @param object
+   */
+  public void loadSettings(PropertiesWrapper p) {
```
rbygrave: Plus a loadSettings() method ... (which is good)

src/main/java/io/ebean/config/PlatformConfig.java
```diff
@@ -0,0 +1,282 @@
...
+      for (Entry<String, String> entry : map.entrySet()) {
+        addCustomMapping(DbType.valueOf(entry.getKey()), entry.getValue());
+      }
+    }
+  }
+
+  /**
+   * Specify how UUID is stored.
+   */
+  public enum DbUuid {
```
rbygrave: This is an addition I think

src/main/java/io/ebeaninternal/dbmigration/DefaultDbMigration.java
```diff
@@ -222,6 +224,15 @@ public void addDatabasePlatform(DatabasePlatform databasePlatform, String prefix
   @Override
   public String generateMigration() throws IOException {
 
+
+    // configure the platform
```
rbygrave: Yeah ok ... anything DDL / migration related really need to be in PlatformConfig to support this

src/main/java/io/ebean/config/PlatformConfig.java
rbygrave: And ultimately this is to support ... a single ServerConfig with many platforms for migration, so IdType needs to move from ServerConfig to here.

rbygrave:
Ok, I have basically copied these changes as part of #1342

So I'll mark this as a duplicate now.

PR #1350, opened 2018-03-17
 ENH: Add mappingsLocation config to search xml mapping files.

yuanxuegui:
Users from mybatis want to use sql xml mappings, it's a good choose to add this feature, config mappingLocations for ebean server to search xml mappings.

### DIFF
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -14,14 +14,7 @@ import io.ebean.config.dbplatform.DbEncrypt;
 import io.ebean.config.dbplatform.DbType;
 import io.ebean.config.dbplatform.IdType;
 import io.ebean.config.properties.PropertiesLoader;
-import io.ebean.event.BeanFindController;
-import io.ebean.event.BeanPersistController;
-import io.ebean.event.BeanPersistListener;
-import io.ebean.event.BeanPostConstructListener;
-import io.ebean.event.BeanPostLoad;
-import io.ebean.event.BeanQueryAdapter;
-import io.ebean.event.BulkTableEventListener;
-import io.ebean.event.ServerConfigStartup;
+import io.ebean.event.*;
 import io.ebean.event.changelog.ChangeLogListener;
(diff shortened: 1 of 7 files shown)


rbygrave:
This looks good.  I'll have another review later today and look to merge. Thanks!!

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -14,14 +14,7 @@
...
 import io.ebean.config.properties.PropertiesLoader;
-import io.ebean.event.BeanFindController;
-import io.ebean.event.BeanPersistController;
-import io.ebean.event.BeanPersistListener;
-import io.ebean.event.BeanPostConstructListener;
-import io.ebean.event.BeanPostLoad;
-import io.ebean.event.BeanQueryAdapter;
-import io.ebean.event.BulkTableEventListener;
-import io.ebean.event.ServerConfigStartup;
+import io.ebean.event.*;
```
rbygrave: Note: imports

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -2862,6 +2854,12 @@ protected void loadSettings(PropertiesWrapper p) {
     tenantSchemaProvider = p.createInstance(TenantSchemaProvider.class, "tenant.schemaProvider", tenantSchemaProvider);
     tenantPartitionColumn = p.get("tenant.partitionColumn", tenantPartitionColumn);
     classes = getClasses(p);
+
+
+    if (mappingLocations != null) {
```
rbygrave: != null or isEmpty() ?

rbygrave:
Good stuff thanks !!

Note: I'm making a couple of minor adjustments.
- Change the property read from "mappingsLocations" to "mappingLocations"  (remove an extra "s" from mapping(s)).
- Not bother reading the extra "search.mappingsLocations" property (as it was in the other location for backwards compatibility we didn't really need it here in my opinion).
- Extra adjustment to modify getSearchList() to pass and return the existing `List<String>` ...

PR #1355, opened 2018-03-27
NEW: QueryPlanLogger

rPraml:
This was used to log the query plans for the "multivaluebind" optimization.
This might be useful in some other places, too.


rPraml:
@rbygrave what is with this PR? 

This PR adds the feature that you can log the database query plan (EXPLAIN SELECT...).
It is controlled by the "io.ebean.QUERYPLAN" logger setting and will be active if log level is at least DEBUG.

rPraml:
This PR is obsolete now

PR #1361, opened 2018-04-10
Inheritance - Add support for @DiscriminatorValue to be optional

ajcamilo:
JPA states:

"If the DiscriminatorValue annotation is not specified and a discriminator column is used, a provider-specific function will be used to generate a value representing the entity type. If the DiscriminatorType is STRING, the discriminator value default is the entity name." in https://docs.oracle.com/javaee/7/api/javax/persistence/DiscriminatorValue.html

This change adds support to this by setting the `DiscriminatorValue` to the class name when the annotation is not provided.

But there's a catch, ebean relied on the `DiscriminatorValue` annotation to know if the entity is concrete or not. I changed this behavior to look for the `abstract` modifier in the class.


rbygrave:
Yes, looks all good. Well done, thanks !!

PR #1370, opened 2018-05-09
trigger based history will recreate view properly and handle default …

rPraml:
…value also for history view

Hello Rob. This fixes some issues with trigger based history. In many cases the view is not recreated properly for MySql/H2. Please take a look.


rPraml:
Please hold back. I think there is still a bug in mariadb

rbygrave:
Is this ready for merging ?

PR #1373, opened 2018-05-11
FIX: NPE when fetching inherited beans from cache

rPraml:
**Steps to reproduce**
You need a base entity that references an inherited entity. (CInhRef -> CInhRoot)

Alt least CInhRef must be cacheable. When hitting the cache, the code tries to create a reference to CInhRoot, which is abstract and cannot be instantiated.

This fix hits the database (or cache) to get the concrete entity.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
+++ b/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
@@ -71,6 +71,7 @@ import io.ebeaninternal.server.query.STreePropertyAssocMany;
 import io.ebeaninternal.server.query.STreePropertyAssocOne;
 import io.ebeaninternal.server.query.STreeType;
 import io.ebeaninternal.server.query.SqlBeanLoad;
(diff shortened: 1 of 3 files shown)


rPraml:
@rbygrave Can you take a look at the build error:
```
Could not resolve dependencies for project io.ebean:ebean:jar:11.15.7-SNAPSHOT: Failed to collect dependencies at org.avaje.composite:avaje-composite-testing:pom:1.1 -> ch.qos.logback:logback-classic:jar:1.3.0-alpha1 -> org.slf4j:slf4j-api:jar:1.8.0-beta1-SNAPSHOT: Failed to read artifact descriptor for org.slf4j:slf4j-api:jar:1.8.0-beta1-SNAPSHOT: Could not transfer artifact org.slf4j:slf4j-api:pom:1.8.0-beta1-SNAPSHOT from/to apache-snapshots (https://repository.apache.org/snapshots/): Connect to repository.apache.org:443 [repository.apache.org/207.244.88.140] failed: Connection timed out (Connection timed out) -> [Help 1]
```
I think we have a version conflict with the new SLF4J API

Issue #1382, opened 2018-05-22
Optional OneToOne with PrimaryKeyJoinColumn fail with EntityNotFoundException


rPraml:
## Expected behavior

Fetching a OneToOne with optional "extra" should set the extra to null if it does not exist

## Actual behavior
```
javax.persistence.EntityNotFoundException: Lazy loading failed on type:org.tests.model.onetoone.OtoUPrimeExtra id:e4fd0a89-bb1a-4fcc-ae1c-6b9db37081b7 - Bean has been deleted.
	at io.ebean.bean.EntityBeanIntercept.loadBeanInternal(EntityBeanIntercept.java:846)
	at io.ebean.bean.EntityBeanIntercept.loadBean(EntityBeanIntercept.java:815)
	at io.ebean.bean.EntityBeanIntercept.preGetter(EntityBeanIntercept.java:915)
	at org.tests.model.onetoone.OtoUPrimeExtra._ebean_get_extra(OtoUPrimeExtra.java:6)
	at org.tests.model.onetoone.OtoUPrimeExtra.getExtra(OtoUPrimeExtra.java:37)
```	

### Steps to reproduce

```java
  @Test
  public void insertWithoutExtra() {
    String desc = "" + System.currentTimeMillis();
    OtoUPrime p1 = new OtoUPrime("u" + desc);
    Ebean.save(p1);

    Query<OtoUPrime> query = Ebean.find(OtoUPrime.class).setId(p1.getPid());

    OtoUPrime found = query.findOne();

    if (found.getExtra() != null) {
      found.getExtra().getExtra(); // fails here, because getExtra should be null
    }
    assertThat(found.getExtra()).isNull();
  }
```

### Suggestions / ideas how to fix
Either perform a left join when the OtoUPrime bean is fetched (would affect performance) - or on the first `getExtra()` call, check if the bean is a reference bean. This could be done in the preGetter like this:
```java
  public void preGetter(int propertyIndex) {
    preGetterCallback(propertyIndex);
    if (state == STATE_NEW || disableLazyLoad) {
      return;
    }
    if (!isLoadedProperty(propertyIndex)) {
      loadBean(propertyIndex);
    }
    // Quick & Ditry hack to fix that particular problem (may have side effects!)
    Object obj = getOwner()._ebean_getField(propertyIndex);
    if (obj instanceof EntityBean ) { // TODO Check if it is Oto Optional
[...]

PR #1383, opened 2018-05-23
Possible workaround for #1382

rPraml:


rPraml:
Closed. See #1382 for workaround

PR #1403, opened 2018-06-06
SaveManyBeans - fixes

JakubPetr:
1. enable soft delete when deleting missing children
2. remove orphans for stateless update

### DIFF
--- a/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java
+++ b/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java
@@ -76,7 +76,7 @@ class SaveManyBeans extends SaveManyBase {
       }
       if (cascade) {
         // potentially deletes 'missing children' for 'stateless update'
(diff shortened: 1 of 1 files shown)


rbygrave:
It is good to see the effort so good stuff there.  Ideally the first step is to produce a failing test case.  There might not be an existing set of test entity beans / model in Ebean's test suite to reproduce but that is really the first thing to check and add.

rbygrave:
I have made the fix/change for this as part of https://github.com/ebean-orm/ebean/issues/1425

... so I think we can close this PR now.

rbygrave:
Note that internally I changed to use an enum DeleteMode ... and that made the code easier to follow and understand I believe.  You were pretty much on the money ... SaveManyBeans wasn't doing the correct thing.

rbygrave:
Note that this issue was fixed with https://github.com/ebean-orm/ebean/issues/1425 ... so I think we should close this PR.   Are you happy to do that?

Thanks, Rob.

PR #1409, opened 2018-06-08
Potential resource leaks, as reported by Infer.

KoenDG:
Infer: https://github.com/facebook/infer

Infer is a static analysis tool that's a bit different. It goes very specifically into resource leaks, null dereference and thread safety violations.

Example: `infer -- mvn clean compile` and it will do the analysis on the code.

Afterwards a report is generated. Then you can use the command `infer-explore`, which will let you select a report and it will generate the code path through which it detected the potential problem.

Report for ebean was:
```
  THREAD_SAFETY_VIOLATION: 38
         NULL_DEREFERENCE: 23
            RESOURCE_LEAK: 6
```

I focussed on resource leaks as other subjects require a deeper knowledge of the code, which I don't have.

Fixed as best as possible. Comments added. Not all reported leaks are guaranteed to be leaks, as the algorithm seems to be very conservative in determining this.

Comments left in the code in places of uncertainty, will remove after getting comments about them.

That being said, I can't recall if PR's on this repo trigger Travis builds or something similar. Having all the tests run on these change would be nice and I just don't have the various databases installed for it.

There were more reports than the ones I put in this PR, but those seemed to be new instances that have to be passed along and kept in memory. I feel Infer reports resource leaks when it sees that the scope is being exited and the Closable that was created at some point in the scope hasn't been closed. Ignored that maybe that's intended.

That last part is something to keep in mind: the analysis is very conservative and you may spend a lot of time researching the code, to come to the conclusion that it's almost all false positives.

### DIFF
--- a/src/main/java/io/ebean/config/properties/YamlLoader.java
+++ b/src/main/java/io/ebean/config/properties/YamlLoader.java
@@ -2,6 +2,7 @@ package io.ebean.config.properties;
 
 import org.yaml.snakeyaml.Yaml;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.Map;
 
--- a/src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
+++ b/src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
@@ -173,10 +173,13 @@ public class DdlGenerator {
   protected void runResourceScript(String sqlScript) throws IOException {
 
     if (sqlScript != null) {
-      InputStream is = getClassLoader().getResourceAsStream(sqlScript);
-      if (is != null) {
-        String content = readContent(new InputStreamReader(is));
-        runScript(false, content, sqlScript);
+      try (InputStream is = getClassLoader().getResourceAsStream(sqlScript)) {
+        if (is != null) {
+          String content = readContent(new InputStreamReader(is));
+          runScript(false, content, sqlScript);
+        }
+      } catch (IOException ioe) {
+        // Nothing, because that's how it was before.
       }
     }
   }
@@ -273,7 +276,6 @@ public class DdlGenerator {
         buf.append(s).append("\n");
       }
       return buf.toString();
-
     }
   }
 
--- a/src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
+++ b/src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
@@ -97,6 +97,9 @@ public class ScalarTypeFile extends ScalarTypeBase<File> {
     } else {
       try {
         // stream from our file to the db
+        // This stream doesn't seem to be closed anywhere.
+        // It certainly can't be closed from here, but where then?
+        // Is it closed when the thread dies?
         InputStream fi = getInputStream(value);
         b.setBinaryStream(fi, value.length());
       } catch (IOException e) {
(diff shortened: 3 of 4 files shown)


src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
```diff
@@ -97,6 +97,9 @@ public void bind(DataBind b, File value) throws SQLException {
     } else {
       try {
         // stream from our file to the db
+        // This stream doesn't seem to be closed anywhere.
+        // It certainly can't be closed from here, but where then?
```
KoenDG: What I mean is: it seems like the way the code works is to pass along an InputStream, set it and use that later whenever necessary. Meaning it has to remain open and cannot be closed.

src/main/java/io/ebean/config/properties/YamlLoader.java
```diff
@@ -27,6 +28,12 @@ void load(InputStream is) {
       for (Object map : yaml.loadAll(is)) {
         loadMap((Map<String, Object>)map, null);
       }
+
+      try {
+        is.close();
+      } catch (IOException ioe) {
+        // Nothing, because that's what it was before this.
+      }
```
rPraml: It's not a good design pattern to close resource-parameters in methods. There are the following converntions (which may violated in certain cases)
- If you pass in a resource, the caller is responsible to close it
- If you wrap a stream (e.g. `new BufferedInputStream(is)`), closing the outer stream will close also the inner.

You should fix that resource leak at the caller side, there is only one in io.ebean.config.properties.Loader:
```java
void loadYaml(String resourcePath, Source source) {
  if (yamlLoader != null) {
    try (InputStream is = resource(resourcePath, source) {
      yamlLoader.load(is);
    } catch (Exception e) {
      log.warn("Failed to read yml from:" + resourcePath, e);
    }
  }
}
```

KoenDG: I was wondering about that, caller side it is.

src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
rPraml: This is really very tricky... I've written a small test and the stream stays open.
As far as I understand, the stream must not be closed until the data is in the database (i.e. until executeUpdate/executeBatch is called)
My current plan is to keep a reference of all open streams in DataBind/BatchedPstmt

KoenDG:
I updated the code as suggested. Should be safe. All the added code should close resources only in case of one-shot use cases, like reading a file and then after the method is done, the resource can be closed. Not like streaming resources that have to remain open for constant use.

src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
```diff
@@ -173,10 +173,13 @@ protected void runSeedSql() throws IOException {
...
-      if (is != null) {
-        String content = readContent(new InputStreamReader(is));
-        runScript(false, content, sqlScript);
+      try (InputStream is = getClassLoader().getResourceAsStream(sqlScript)) {
+        if (is != null) {
+          String content = readContent(new InputStreamReader(is));
+          runScript(false, content, sqlScript);
+        }
+      } catch (IOException ioe) {
+        // Nothing, because that's how it was before.
```
rPraml: the catch would be obsolete here. note that this change is already in #1411

src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
```diff
@@ -194,13 +197,19 @@ public long pump(InputStream is, OutputStream out) throws IOException {
     } finally {
       if (output != null) {
         try {
+          // Closing BufferedOutputStream does not close the underlying OutputStream.
+          // It only performs a flush() on the underlying OutputStream.
```
rPraml: Thats not true, close method of FilterOutputStream
```java
    public void close() throws IOException {
        try (OutputStream ostream = out) {
            flush();
        }
    }
```
(the .close is a bit hard to find... Note that there is a try-with-resources statement)

rbygrave:
If you modify this to remove the change to ScalarTypeFile (which I don't see as being correct) then I can look to merge the other 2 changes.  Do you want to do that?

rbygrave:
Actually I have just put these 2 changes into https://github.com/ebean-orm/ebean/issues/1412

So I think this PR can be closed.

KoenDG:
Neat 👍

PR #1411, opened 2018-06-09
Fix potential resource leaks

rPraml:
This is a bigger refactor of possible resource leaks inspired by the hints from @KoenDG

### DIFF
--- a/src/main/java/io/ebean/EbeanVersion.java
+++ b/src/main/java/io/ebean/EbeanVersion.java
@@ -13,31 +13,31 @@ import org.slf4j.LoggerFactory;
  */
 public class EbeanVersion {
   private EbeanVersion() {
-    
+
   }
   private static final Logger logger = LoggerFactory.getLogger(EbeanVersion.class);
-  
+
   private static String version = "unknown";
   static {
     try {
       Properties prop = new Properties();
-      InputStream in = Ebean.class.getResourceAsStream("/META-INF/maven/io.ebean/ebean/pom.properties");
-      if (in != null) {
-        prop.load(in);
-        in.close();
-        version = prop.getProperty("version");
+      try (InputStream in = Ebean.class.getResourceAsStream("/META-INF/maven/io.ebean/ebean/pom.properties")) {
+        if (in != null) {
+          prop.load(in);
+          version = prop.getProperty("version");
+        }
       }
       logger.info("ebean version: {}", version);
     } catch (IOException e) {
       logger.warn("Could not determine ebean version: {}", e.getMessage());
     }
   }
-  
+
   /**
    * Returns the ebean version (read from /META-INF/maven/io.ebean/ebean/pom.properties)
    */
   public static String getVersion() {
     return version;
   }
-  
+
 }
--- a/src/main/java/io/ebean/config/properties/LoadContext.java
+++ b/src/main/java/io/ebean/config/properties/LoadContext.java
@@ -5,7 +5,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -18,22 +18,22 @@ import java.util.Set;
  */
 class LoadContext {
 
-	private static final Logger log = LoggerFactory.getLogger(LoadContext.class);
+  private static final Logger log = LoggerFactory.getLogger(LoadContext.class);
 
   /**
    * Map we are loading the properties into.
    */
-	private final Map<String, String> map = new LinkedHashMap<>();
+  private final Map<String, String> map = new LinkedHashMap<>();
 
   /**
    * Names of resources/files that were loaded.
    */
-	private final Set<String> loadedResources = new LinkedHashSet<>();
+  private final Set<String> loadedResources = new LinkedHashSet<>();
 
   /**
    * Return the input stream (maybe null) for the given source.
    */
-	InputStream resource(String resourcePath, Loader.Source source) {
+  InputStream resource(String resourcePath, Loader.Source source) throws IOException {
 
     InputStream is = null;
     if (source == Loader.Source.RESOURCE) {
@@ -44,23 +44,19 @@ class LoadContext {
     } else {
       File file = new File(resourcePath);
       if (file.exists()) {
-        try {
-          is = new FileInputStream(file);
-          loadedResources.add("file:"+resourcePath);
-        } catch (FileNotFoundException e) {
-          throw new IllegalStateException(e);
-        }
+        is = new FileInputStream(file);
+        loadedResources.add("file:" + resourcePath);
       }
     }
 
-		return is;
-	}
+    return is;
+  }
 
   private InputStream resourceStream(String resourcePath) {
     InputStream is = getClass().getResourceAsStream("/" + resourcePath);
     if (is == null) {
       // search the module path for top level resource
-      is =  ClassLoader.getSystemResourceAsStream(resourcePath);
+      is = ClassLoader.getSystemResourceAsStream(resourcePath);
     }
     return is;
   }
@@ -68,30 +64,30 @@ class LoadContext {
   /**
    * Add a property entry.
    */
-	void put(String key, String val) {
-	  if (val != null) {
-	    val = val.trim();
+  void put(String key, String val) {
+    if (val != null) {
+      val = val.trim();
     }
-		map.put(key, val);
-	}
+    map.put(key, val);
+  }
 
   /**
    * Evaluate all the expressions and return as a Properties object.
    */
-	Properties eval() {
+  Properties eval() {
 
-		log.info("loaded properties from {}", loadedResources);
+    log.info("loaded properties from {}", loadedResources);
 
-		Properties properties = new Properties();
+    Properties properties = new Properties();
 
-		for (Map.Entry<String, String> entry : map.entrySet()) {
-			String key = entry.getKey();
-			String value = PropertyEval.eval(entry.getValue());
-			properties.setProperty(key, value);
-		}
+    for (Map.Entry<String, String> entry : map.entrySet()) {
+      String key = entry.getKey();
+      String value = PropertyEval.eval(entry.getValue());
+      properties.setProperty(key, value);
+    }
 
-		return properties;
-	}
+    return properties;
+  }
 
   /**
    * Read the special properties that can point to an external properties source.
--- a/src/main/java/io/ebean/config/properties/Loader.java
+++ b/src/main/java/io/ebean/config/properties/Loader.java
@@ -16,18 +16,17 @@ import java.util.Properties;
  */
 class Loader {
 
-	private static final Logger log = LoggerFactory.getLogger(Loader.class);
+  private static final Logger log = LoggerFactory.getLogger(Loader.class);
 
-	enum Source {
-	  RESOURCE,
-    FILE
+  enum Source {
+    RESOURCE, FILE
   }
 
   private final LoadContext loadContext = new LoadContext();
 
   private YamlLoader yamlLoader;
 
-	Loader() {
+  Loader() {
     String skipYaml = System.getProperty("ebeanSkipYaml");
     if (!"true".equals(skipYaml)) {
       try {
@@ -44,19 +43,19 @@ class Loader {
   /**
    * Load the configuration with the expected ordering.
    */
-	void load() {
+  void load() {
 
     loadMain(Source.RESOURCE);
     // external file configuration overrides the resources configuration
     loadMain(Source.FILE);
 
-		loadViaSystemProperty();
-		loadViaIndirection();
+    loadViaSystemProperty();
+    loadViaIndirection();
 
-		// test configuration (if found) overrides main configuration
+    // test configuration (if found) overrides main configuration
     // we should only find these resources when running tests
-		loadTest();
-	}
+    loadTest();
+  }
 
   /**
    * Load test configuration.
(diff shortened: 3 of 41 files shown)


src/main/java/io/ebean/EbeanVersion.java
```diff
@@ -13,31 +13,31 @@
...
   private static final Logger logger = LoggerFactory.getLogger(EbeanVersion.class);
-  
+
   private static String version = "unknown";
   static {
     try {
       Properties prop = new Properties();
-      InputStream in = Ebean.class.getResourceAsStream("/META-INF/maven/io.ebean/ebean/pom.properties");
-      if (in != null) {
-        prop.load(in);
```
rPraml: OLD: if prop.load throws an exception, the stream was not closed
NEW: try-with-resources will ensure, that the stream is closed properly

src/main/java/io/ebean/config/properties/LoadContext.java
```diff
@@ -44,54 +44,50 @@ InputStream resource(String resourcePath, Loader.Source source) {
     } else {
       File file = new File(resourcePath);
       if (file.exists()) {
-        try {
-          is = new FileInputStream(file);
-          loadedResources.add("file:"+resourcePath);
-        } catch (FileNotFoundException e) {
-          throw new IllegalStateException(e);
```
rPraml: OLD: FileNotFoundException was catched and wrapped in an IllegalStateException
NEW: Method throws IOException

src/main/java/io/ebean/config/properties/LoadContext.java
```diff
@@ -44,54 +44,50 @@ InputStream resource(String resourcePath, Loader.Source source) {
...
+    for (Map.Entry<String, String> entry : map.entrySet()) {
+      String key = entry.getKey();
+      String value = PropertyEval.eval(entry.getValue());
+      properties.setProperty(key, value);
+    }
 
-		return properties;
-	}
+    return properties;
+  }
```
rPraml: class was reformatted

src/main/java/io/ebean/config/properties/Loader.java
```diff
@@ -112,59 +112,42 @@ void loadWithExtensionCheck(String fileName) {
...
-	void loadYaml(String resourcePath, Source source) {
-		if (yamlLoader != null) {
-		  try {
-			  yamlLoader.load(resource(resourcePath, source));
+  void loadYaml(String resourcePath, Source source) {
+    if (yamlLoader != null) {
+      try (InputStream is = resource(resourcePath, source)) {
+        yamlLoader.load(is);
       } catch (Exception e) {
         log.warn("Failed to read yml from:" + resourcePath, e);
```
rPraml: Uses try with resources now.
@rbygrave  should we throw an exception instead of printing a warning if the yml file could not be read?

src/main/java/io/ebeaninternal/dbmigration/migrationreader/MigrationXmlReader.java
```diff
@@ -24,12 +24,16 @@ private MigrationXmlReader() {
...
+    try (InputStream is = MigrationXmlReader.class.getResourceAsStream(resourcePath)) {
+
+      if (is == null) {
+        throw new IllegalArgumentException("No resource found for path [" + resourcePath + "]");
+      }
 
-    return read(is);
+      return read(is);
+    } catch (IOException e) {
+      throw new RuntimeException("Could not read resource: " + resourcePath, e);
```
rPraml: I've added the file name to the error handlers.

src/main/java/io/ebeaninternal/dbmigration/migrationreader/MigrationXmlReader.java
```diff
@@ -40,22 +44,22 @@ public static Migration read(File migrationFile) {
...
+  public static Migration read(InputStream is) throws IOException {
 
     try {
       JAXBContext jaxbContext = JAXBContext.newInstance(Migration.class);
       Unmarshaller unmarshaller = jaxbContext.createUnmarshaller();
       return (Migration) unmarshaller.unmarshal(is);
 
     } catch (JAXBException e) {
-      throw new RuntimeException(e);
+      throw new IOException(e);
```
rPraml: if read will fail, the excepton is wrapped in an IoExcecption, this exception is catched by the caller and wrapped in a RuntimeException with the resource name.
So, this change should ensure, that the filename of the exception is in the stack trace.

src/main/java/io/ebeaninternal/server/autotune/service/DefaultAutoTuneService.java
```diff
@@ -106,11 +107,14 @@ private void loadTuningFile() {
       loadAutoTuneProfiling(AutoTuneXmlReader.read(file));
     } else {
       // look for autotune as a resource
-      InputStream stream = getClass().getResourceAsStream("/" + tuningFile);
-      if (stream != null) {
-        loadAutoTuneProfiling(AutoTuneXmlReader.read(stream));
-      } else {
-        logger.warn("AutoTune file {} not found - no initial automatic query tuning", tuningFile);
+      try (InputStream stream = getClass().getResourceAsStream("/" + tuningFile);) {
```
rPraml: We had a lot resource leaks, where the streams were not closed. :(

src/main/java/io/ebeaninternal/server/type/DataBind.java
```diff
@@ -25,11 +29,30 @@
...
+    this.connection = connection;
+  }
+
+  /**
+   * Wraps a {@link BatchedPstmt} and takes accout of batch requirements.
+   */
+  public DataBind(DataTimeZone dataTimeZone, BatchedPstmt batchedPstmt, Connection connection) {
+    this.dataTimeZone = dataTimeZone;
+    this.pstmt = batchedPstmt.getStatement();
+    this.batchedPstmt = batchedPstmt;
```
rPraml: So, here comes the complex part - closing File-Streams - see the test case: https://github.com/ebean-orm/ebean/pull/1411/files#diff-b9f83866e73ea4adc1eceeaf35449eb6R112

So, what is done here:
1. DataBind will no longer close the prepared statement
2. DataBind will track a list of all InputStreams that are bound to the preparedStatement in non batch mode.
3. When the DataBind is closed, it will close all streams from the list (in non batch mode)
4. In batch mode, it delegates that list to the BatchedPstmt which handles the resource cleanup

src/main/java/io/ebeaninternal/server/core/AbstractSqlQueryRequest.java
```diff
@@ -114,6 +116,12 @@ public void close() {
     requestComplete();
     JdbcClose.close(resultSet);
     JdbcClose.close(pstmt);
+    resultSet = null;
+    pstmt = null;
+    if (dataBind != null) {
+      dataBind.close();
+      dataBind = null;
+    }
```
rPraml: We must track the resultset, pstmt AND dataBind - and close them all

src/main/java/io/ebeaninternal/server/core/DatabasePlatformFactory.java
```diff
@@ -115,24 +115,13 @@ private DatabasePlatform byDatabaseName(String dbName) {
...
       throw new PersistenceException(ex);
-
-    } finally {
-      try {
-        if (connection != null) {
-          connection.close();
-        }
-      } catch (SQLException ex) {
-        logger.error(null, ex);
-      }
```
rPraml: @rbygrave I simplified the try catch blocks.
OLD: if connection.close throws an error, the error was logged
NEW: try-with-resources will catch that and a PersistenceException is thrown now.
Do you think this is OK?

src/main/java/io/ebeaninternal/server/core/bootup/ManifestReader.java
```diff
@@ -55,31 +55,21 @@ ManifestReader read(String resourcePath) {
...
+        url = resources.nextElement();
+        try (InputStream is = url.openStream()) {
           read(new Manifest(is));
-        } finally {
-          close(is);
         }
       }
     } catch (IOException e) {
-      logger.warn("Error reading " + resourcePath + " manifest resources", e);
+      logger.warn("Error reading " + resourcePath + "(" + url + ") manifest resources", e);
```
rPraml: Added the URL to the message

src/main/java/io/ebeaninternal/server/persist/BatchedPstmtHolder.java
```diff
@@ -53,7 +52,7 @@ public PreparedStatement getStmt(String stmtKey, BatchPostExecute postExecute) {
     if (bsSize > maxSize) {
       maxSize = bsSize;
     }
-    return bs.getStatement();
+    return bs;
```
rPraml: here the BatchedPstmt is returned. That will be passed to the DataBind, so that it can register the streams on the BatchedPstmt

src/main/java/io/ebeaninternal/server/persist/Binder.java
```diff
@@ -228,14 +226,8 @@ public Object bindObject(DataBind dataBind, Object value) throws SQLException {
...
-      ScalarType<?> type = getScalarType(value.getClass());
-      if (!type.isJdbcNative()) {
-        // convert to a JDBC native type
-        value = type.toJdbcType(value);
-      }
-
-      int dbType = type.getJdbcType();
-      bindObject(dataBind, value, dbType);
+      ScalarType type = getScalarType(value.getClass());
+      type.bind(dataBind, value);
```
rPraml: @rbygrave take a look at this change. IMHO we should delegate the bind to the ScalarType.
I fall over this, becase bindObject did not support files. (I added a workaround anyway)

src/main/java/io/ebeaninternal/server/persist/Binder.java
```diff
@@ -411,7 +403,15 @@ private void bindLongVarChar(DataBind dataBind, Object data) throws SQLException
    * Bind byte[] data to a LONGVARBINARY column.
    */
   private void bindLongVarBinary(DataBind dataBind, Object data) throws SQLException {
-
+    if (data instanceof File) {
```
rPraml: Workaround to support files

src/main/java/io/ebeaninternal/server/persist/ExeCallableSql.java
```diff
@@ -32,33 +33,28 @@ public int execute(PersistRequestCallableSql request) {
 
     boolean batchThisRequest = request.isBatchThisRequest();
 
-    CallableStatement cstmt = null;
-    try {
-      cstmt = bindStmt(request, batchThisRequest);
+    try (DataBind dataBind = bindStmt(request, batchThisRequest)) {
```
rPraml: we must close dataBind in any case

src/main/java/io/ebeaninternal/server/persist/ExeCallableSql.java
```diff
@@ -32,33 +33,28 @@ public int execute(PersistRequestCallableSql request) {
...
+        dataBind.getPstmt().addBatch();
         // return -1 to indicate batch mode
         return -1;
       } else {
-        // handles executeOverride() and also
-        // reading of registered OUT parameters
-        int rowCount = request.executeUpdate();
-        request.postExecute();
-        return rowCount;
+        try (PreparedStatement pstmt = dataBind.getPstmt()) {
```
rPraml: and prepared statement in non-batch mode

src/main/java/io/ebeaninternal/server/persist/dml/DmlHandler.java
```diff
@@ -77,24 +76,12 @@ public long now() {
...
-   * Get the sql and bind the statement.
-   */
-  @Override
-  public abstract void bind() throws SQLException;
-
-  /**
-   * Execute now for non-batch execution.
-   */
-  @Override
-  public abstract int execute() throws SQLException;
```
rPraml: methods are in interface (with same comment) - so I removed these

src/main/java/io/ebeaninternal/server/type/ScalarTypeFile.java
```diff
@@ -71,16 +71,15 @@ private OutputStream getOutputStream(File value) throws IOException {
...
+    try (InputStream is = dataReader.getBinaryStream()) {
+      if (is == null) {
+        return null;
+      }
       // stream from db into our temp file
       File tempFile = File.createTempFile(prefix, suffix, directory);
-      OutputStream os = getOutputStream(tempFile);
-      pump(is, os);
+      try (OutputStream os = getOutputStream(tempFile)) {
+        pump(is, os);
```
rPraml: pump does no longer close the stream.

src/test/java/org/tests/types/TestFileType.java
```diff
@@ -98,6 +108,84 @@ public void test_insertUpdateDelete() {
     Ebean.delete(bean3);
   }
 
+  @Test
+  public void test_canDeleteIfStreamIsOpen() throws IOException {
```
rPraml: this test tests only, if delete will fail, if a stream is open.

rbygrave:
> This is a bigger refactor

Yes and unfortunately I see this PR as being too big.  There was some nice good small obvious improvements in here but they are bundled into this PR that also pretty important changes to DataBind which puts the whole PR at risk.

rPraml:
I've updated the PR, unfortunately, the PR is still big and the whole DataBind & Stream tracking changes are very complex. I also wonder if this is the correct approach, maybe we can track the streams we have to close in the transaction. This might be simpler.

KoenDG:
It's certainly very big and probably requires some extensive testing. After all, you don't want to close a resource that's intended to be re-used at a later point.

I say that entirely theoretically, because I don't actually know if the code contains such constructs.

rbygrave:
> // find all beans that have the same file content

I think this is the crux of the difference in how I see this issue.  I am not keen/interesting in supporting query predicates with File as an argument.  I have never had that need and can't see it as a valuable feature. 

Hence I am not keen/interested in all the change to DataBind etc - I don't think we need it at all.

That is, I really only want to address the ScalarTypeFile use cases as they currently are - insert and update of beans with a property of type File. Binding as part of insert/update ... and not the select where clause.

I'll pursue that path ...

PR #1422, opened 2018-06-13
FIX: applied history drops were not recognized

rPraml:
when history was dropped, pending drops were not marked as applied


rbygrave:
Good work !!

PR #1428, opened 2018-06-15
ENH: Use java.time.Clock for @WhenModified and @WhenCreated and allow it to be Replaceable for testing

tobias-:
As per discussion https://groups.google.com/forum/#!topic/ebean/cvy6MDhji3U a simple way to replace the clock used for setting @UpdatedTimestamp and
@CreatedTimestamp. 

I'm pretty new to hacking in the Ebean source, but this solves the problem for us.

### DIFF
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -33,6 +33,7 @@ import io.ebean.util.StringHelper;
 import org.avaje.datasource.DataSourceConfig;
 
 import javax.sql.DataSource;
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


rbygrave:
Good stuff.  I have just adjusted it a little ... storing the Clock on the DefaultServer and allowing it to be set via ExtendedServer interface.  

Cheers, Rob.

tobias-:
@rbygrave Looks better now. I didn't see that the ServerConfig was only for initializing the DefaultServer.

They may need to be further enhancements for history tables and the Triggers, but I don't understand it well enough to do it.

rbygrave:
> They may need to be further enhancements for history tables and the Triggers,

That is it done with nothing more to do really.

There is a case for adding the option that `@WhenModified` etc uses DB `current_timestamp` rather than our "client side JVM clock" ... but that isn't on the agenda really. 

So yes, we can consider it done.  A nice and small but pretty cool addition - great stuff !!

PR #1435, opened 2018-06-18
Migration enum drop check constraints

focbenz:
When the Enum property in a model was changed to String the check constraint was not removed from the column.

Note: Starting with MariaDB 10.2.1 the CHECK constraints are supported by the storage engines. This might be a problem because the current implementation relies on check constraints are passed but ignored (which is true for MySQL in version 5.7/8.0)


rbygrave:
Yes, this looks all good to me. I'll look to merge tomorrow - thanks !!

rbygrave:
Darn, I forgot about this ... missed getting it into 11.18.1 - my bad.

focbenz:
No issues Rob ... we already put this in the fork but thanks for the heads up ;-)

PR #1449, opened 2018-07-11
Failing testcase with lazy fetch (when logical property name starts with underscore)

tobias-:
All entities with a renamed id starting with _ fails when
lazy-fetching due to sort not being properly renamed .

I noticed it with the following snippet wreaking havoc.

```kotlin
@Id
@Column(name = "id")
val _id: UUID
```

```
Caused by: org.h2.jdbc.JdbcSQLException: Column "_ID" not found; SQL statement:
select t0.id, t1.id, t1.name, t1.tune_id from tune t0 left join looney t1 on t1.tune_id = t0.id  where t0.id = ?   order by _id [42122-196]
```

### DIFF
--- a/src/test/java/org/tests/model/lazywithid/Looney.java
+++ b/src/test/java/org/tests/model/lazywithid/Looney.java
@@ -1,0 +1,36 @@
+package org.tests.model.lazywithid;
+
+import javax.persistence.Entity;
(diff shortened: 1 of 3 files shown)


rbygrave:
Thanks for the testcase !! Fixed in 11.18.5 which I will look to release shortly.

PR #1450, opened 2018-07-11
Fix Transaction Flushing when ID Property is Loaded

Mitsch164:
Hi Rob,

I have an import-routine that inserts/updates model in a certain order via batchmode. To do that,  I have to check the ID-property (not generated) of the model prior in the batch. Although the id-property is set in the previous model, the bean getter will call "flushBatch", so that the pervious model ist saved to the database, which is not necessary. This results in flushing the batch before every insert.

My testcase is similar to the existing test case "testFlushOnGetId"  and has the logging output:

ebean test log


rbygrave:
Yes, very good work - well done.

PR #1451, opened 2018-07-12
DB Migration with helper stored procedures (for MySql and SQL Server)

rPraml:
we use stored procedures in DDL migration now.


rbygrave:
So this looks good to me but probably needs to go into 11.19.x (bumped version).

PR #1454, opened 2018-07-16
FIX: support for ScalarTypeConverter and @Lob

rPraml:
*use case*
We have a ScalarTypeConverter that converts a String to an "EncryptedString". (see test case)
If we try to annotate that property with @lob, we got an error as ebean thinks it must use "blob" instead "clob" as column type.

This PR tries to fix this issue. Unfortunately, I do not understand exactly, what happens in DeployUtil.setLobType and why the ScalarTypes are replaced there.
(ScalarTypeString and ScalarTypeClob are very similar, maybe we do not need both)

Maybe you can take a look at this, I have provided some test cases that addresses my issue.


rbygrave:
> why the ScalarTypes are replaced there

Yes, it is a bit of a thin argument.  The Clob scalar type streams from the DB but basically that is the only difference.  For large Clob/Blob treatment the File type gives us the better streaming to result rather than just holding it all in memory as a String.

Anyway, I like what you have done here. I'll merge in and then look to tidy ScalarTypeClob and ScalarTypeString.

PR #1455, opened 2018-07-17
FIX: null handling in ScalarType.toBeanType

rPraml:
**Bug:** When setting java 8 time API values to null with `pathSet(bean, null)`, you'll get a NPE.

This PR tries to fix this bug. I've also checked all other toBeanType methods, if they can handle `null`


rbygrave:
Good stuff, thanks !!

PR #1467, opened 2018-08-09
Let enum arrays be saved as integers or strings

tobias-:
As it is now, all `@DbArray` are always stored as a varchar[] by `enum.name()`, instead of using the `@DbEnumValue` or `@EnumValue` data. This patch aims to fix that.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/type/ArrayElementConverter.java
+++ b/src/main/java/io/ebeaninternal/server/type/ArrayElementConverter.java
@@ -112,14 +112,14 @@ interface ArrayElementConverter<T> {
 
     @Override
     public Object toElement(Object rawValue) {
-      return scalarType.parse(rawValue.toString());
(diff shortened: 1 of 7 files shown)


rbygrave:
The change looks good apart from there being no tests yet.  Are you looking to add a test or 2 (add to an existing entity bean another property with the mapping of `@DbArray` and `@EnumValue` etc) with some insert, update, delete to exercise it. It will be hard to assert I think but I'll have a look at the test when its in).

tobias-:
I wanted some feedback from you before putting in the work, but yes, since you seem to approve of it, I'll add some tests for it. I want to add tests for both sql-generation and insert/update/delete for both integers and strings. Are there any more types?

tobias-:
@rbygrave Good! Tests added. I've run the postgresql test locally, and it worked.

rbygrave:
Yup, looks great - good stuff !!

PR #1479, opened 2018-08-31
FIX: NPE in StalarTypeLocalTime and StalarTypeMonthDay

rPraml:
See #1455 
I forgot two methods


rPraml:
STOP: there is still a bug in SqlServer, which I'm investigating

rPraml:
There's (still) a bug in the driver. Workaround: Set driver option `sendTimeAsDateTime` to false.

See https://github.com/Microsoft/mssql-jdbc/pull/559 and https://github.com/Microsoft/mssql-jdbc/pull/798

@rbygrave Do you have a SqlServer build plan? Then you should adjust the connectionString. (I will not do that here, as this is not part of that PR)

PR #1486, opened 2018-09-20
FIX: recursive BatchedPstmtHolder.flush

rPraml:
Hello Rob, please take a look at this fix, do you think it is worth to merge that?

**How to reproduce**

- Add your own ChangeLogFilter and perform a DB query in the filter (e.g. read config)
- Perform a Batch insert, so that BatchedPstmtHolder.flush() will be called.

The flush will try to insert the values in stmtMap and will notify the Filter (or BeanPersistListeners) for each value (in the loop)

If any of these events will try to query something from the database, it will enforce a flush again (because batchFlushOnMixed is true) and the flush will try to insert the same values again (because the first loop is not yet terminated)


rbygrave:
Yes, this looks good. Merging ...

PR #1489, opened 2018-09-24
removed the concat-operator "||" and replaced with "concat()"

rPraml:
Hi Rob,

this PR removes the concat operator "||" and replaces it with "concat()", because there is no working "+" or "||" operator in SqlServer for different datatypes and it does not work on MySql/MariaDb, too.

Note that SqlServer and MariaDb have also a different handling with null values.

*SqlServer*
`SELECT 'a' + null` = `null`
`SELECT 'a' + 'b'` = `ab`
`SELECT 'a' + 1` = `Conversion failed when converting the varchar value 'a' to data type int.`
`SELECT concat('a',null,'b')` = `ab`
`SELECT concat('a',1)` = `a1`

*MariaDb*
`SELECT 'a' || 'a' ` = `0`
`SELECT 'a' + 'a' ` = `0`
`SELECT 126 || 5` = `4` (logical OR)
`SELECT concat('a',null,'b')` = `null` **Returns null! - Maybe coalesce also needed**

*Postgres*
To be done (concat available since 9.1)

### DIFF
--- a/src/main/java/io/ebeaninternal/server/expression/InPairsExpression.java
+++ b/src/main/java/io/ebeaninternal/server/expression/InPairsExpression.java
@@ -95,13 +95,9 @@ class InPairsExpression extends AbstractExpression {
 
     StringBuilder sb = new StringBuilder();
 
-    sb.append("CONCAT(").append(property0).append(",'").append(separator).append("',").append(property1);
+    request.getDbPlatformHandler().concat(property0, separator, property1, suffix);
 
-    if (suffix != null && !suffix.isEmpty()) {
-      sb.append(",'").append(suffix).append('\'');
-    }
-    sb.append(')');
-    request.append(sb.toString());
+    request.append(request.getDbPlatformHandler().concat(property0, separator, property1, suffix));
     request.appendInExpression(not, concatBindValues);
   }
 
--- a/src/main/java/io/ebeaninternal/server/expression/InPairsExpression.java
+++ b/src/main/java/io/ebeaninternal/server/expression/InPairsExpression.java
@@ -92,14 +92,16 @@ class InPairsExpression extends AbstractExpression {
       return;
     }
 
-    String concat = request.getDbPlatformHandler().getConcatOperator();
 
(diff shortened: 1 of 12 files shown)


rbygrave:
So I am unlikely to merge this until I confirm that this can be used with functional indexes (in Postgres in particular).  That is, `||` was explicitly chosen over `concat` because it CAN be used with functional indexes in Postgres.

Going to `concat` would be a major problem (until that utilises functional indexes in Postgres).

rPraml:
I didn't know that postgres feature, but it makes sense.
I've updated the PR - what do you think now?

src/main/java/io/ebeaninternal/server/expression/InPairsExpression.java
```diff
@@ -92,14 +92,12 @@ public void addSql(SpiExpressionRequest request) {
...
 
-    String concatFormula = "(" + property0 + concat + "'" + separator + "'" + concat + property1;
-    if (suffix != null && !suffix.isEmpty()) {
-      concatFormula += concat + "'" + suffix + "'";
-    }
-    concatFormula += ")";
-    request.append(concatFormula);
+    StringBuilder sb = new StringBuilder();
+
+    request.getDbPlatformHandler().concat(property0, separator, property1, suffix);
```
rbygrave: This should be removed right?

src/main/java/io/ebeaninternal/server/expression/platform/BaseDbExpression.java
```diff
@@ -57,4 +46,15 @@ protected String functionName(BitwiseOp operator) {
     }
   }
 
+  @Override
+  public String concat(String property0, String separator, String property1, String suffix) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("CONCAT(").append(property0).append(",'").append(separator).append("',").append(property1);
```
rbygrave: generally as a style we don't use ALL CAPS keywords.  "concat"

PR #1492, opened 2018-09-25
NEW: collation support for orderBy

rPraml:
this adds collation (and functional) support for orderBy. 
You can specify a collation like `latin_1` which results in an `ORDER BY name COLLATE latin_1`
If the string contains the placeholder `${}`, the argument is considered as function which is required for DB2 or for complex orderings

DB2 example: `o1.desc("name", "COLLATION_KEY(${}, 'latin_1')");`
Complex ordering example: `o1.desc("name", "SOUNDEX(${})");`

### DIFF
--- a/src/main/java/io/ebean/OrderBy.java
+++ b/src/main/java/io/ebean/OrderBy.java
@@ -5,6 +5,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import io.ebean.util.StringHelper;
+
 /**
  * Represents an Order By for a Query.
  * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/io/ebean/OrderBy.java
```diff
@@ -273,6 +311,7 @@ public boolean equals(Object obj) {
       }
       Property e = (Property) obj;
       if (ascending != e.ascending) return false;
+      if (collation != e.collation) return false;
```
rbygrave: I think this should use equals() rather than instance check

src/main/java/io/ebean/OrderBy.java
```diff
@@ -284,19 +323,33 @@ public String toString() {
...
-        sb.append(property);
+
+        // collate
+        if (collation != null)  {
+          if (collation.contains("${}")) {
+            // this is a complex collation, e.g. DB2 - we must replace the property
+            sb.append(StringHelper.replaceString(collation, "${}", property));
+          } else {
+            sb.append(property);
+            sb.append(" COLLATE ").append(collation);
```
rbygrave: Lower case keywords ... "collate"  for consistent style ?

PR #1514, opened 2018-10-24
Add support for JacksonAnnotations at bean-properties

rPraml:
This adds mainly support for @JsonDeserialize annotation and the support for polymorphic types annotated with @JsonTypeInfo 
Example:
```java
public class MyBean {

  // declare concrete implementation
  @DbJson
  @JsonDeserialize(contentAs = SomeImpl.class) 
  Set<SomeInterface> mySet1;

  // use polymorphic types (jackson will add a "type":"a" or "type":"b" as 'discriminator' in this example)
  @DbJson
  Set<BaseType> myPolymorphicSet;
  
  // declare map content
  @DbJson
  @JsonDeserialize(keyAs = Long.class, contentAs = String.class)
  Map<Object, Object> valueMap = new LinkedHashMap<>();
}

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = As.PROPERTY, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = TypeA.class, name = "a"),
        @JsonSubTypes.Type(value = TypeB.class, name = "b")
})
public interface BaseType {

}

public class TypeA implements BaseType {

}

```

### DIFF
--- a/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanDescriptor.java
+++ b/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanDescriptor.java
@@ -223,6 +223,8 @@ public class DeployBeanDescriptor<T> {
 
   private short profileId;
 
+  private Object jacksonAnnotatedClass;
+
   /**
    * Construct the BeanDescriptor.
    */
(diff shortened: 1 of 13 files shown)


rbygrave:
Good stuff !!

PR #1527, opened 2018-11-02
FIX: improved support for @DbForeignKey(noConstraint = true)

rPraml:
**Use case**
I have two tables that references each other (file->owner), but there is no FK constraint.
So I can have files with an ownerId that do not longer exist. e.g:

Table file:

| fileName | ownerId |
|-----|--------|
| bash | 1|
|java | 500|

Table owner:

| userId | userName |
|-----|--------|
|1 | root|
|2 | nobody|

This PR adresses the following things:

**1. Return reference, instead of `null` when `@DbForeignKey(noConstraint = true)` is used with fetch path**

_Actual behavior:_
When using lazy load, you will get the entity `e.fileName=java, e.owner.userId=500`
When accessing `e.owner.userName` you will get a "bean has been deleted" error. (this is ok)

When using eager load with fetch path, you will get the entity `e.fileName=java, e.owner=null`
So you cannot get the information, that the userId was 500 (because the whole bean is null)

_Expected behavior:_
Lazy load and eager load must behave the same (in both cases you want to get `e.owner.userId=500`)

**2. Do not throw Bean has been deleted exception, if there is a softdelete property**

The PR ensures, that the softdelete property is maintained correctly, in lazy and eager load now.
Also ensure, that the bean is considered as fully loaded (all props are null) if it seems to be deleted, so that no subsequent lazy loades will be triggered.

**3. Support for virtual softdelete property**

In my use case, I do not want to get a "bean has been deleted" error and I also cannot enable SoftDelete for that bean. (because, the data come from a 3rd party table) 
So I changed the beavior of the `@SoftDelete` annotation. If there is also a `@Formula` annotation. Ebean treats the softdelete as virtual property and does not require a DB column.

The formula has to compute, if a given bean is deleted or not in left joins. This is archived by
`@Formula(select = "${ta}.user_id is null")`. So if there is no data in the left join, we assume that the bean is deleted. Using formula has also a nice side effect, It triggers the lazy load if bean is not loaded yet.
(I tried `@SoftDelete` in conjunction with `@Transient`, but then no lazy load would be triggered)

So a virtual property softdelete property means, that it will be set to TRUE, if bean has been deleted, but not persisted in database.

I also tried a different approach, you may see in this commit
https://github.com/FOCONIS/ebean/commit/dc9eb012d71aae3b8a3a8c33288790dd93e3c72b
but I think the implementation in the current PR fits better in my use case

**open TODOs**

- We should rename the `lazyLoadFailure` property in EBI to `beanDeleted` as this property is also set in eager load now.
- This PR addresses only the `@ManyToOne` case, the `@OneToMany` and `@ManyToMany` path do not work, yet. (See testcases marked with `@Ignore`)
- We may add an additional flag, so that we can set `@SoftDelete(virtual=true)` instead of specifying an additional formula

(they have low prio for me)

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -829,7 +829,7 @@ public final class EntityBeanIntercept implements Serializable {
 
     if (lazyLoadFailure) {
       // failed when batch lazy loaded by another bean in the batch
-      throw new EntityNotFoundException("Lazy loading failed on type:" + owner.getClass().getName() + " id:" + ownerId + " - Bean has been deleted");
+      throw new EntityNotFoundException("(Lazy) loading failed on type:" + owner.getClass().getName() + " id:" + ownerId + " - Bean has been deleted");
     }
 
     if (lazyLoadProperty == -1) {
--- a/src/main/java/io/ebeaninternal/api/LoadBeanRequest.java
+++ b/src/main/java/io/ebeaninternal/api/LoadBeanRequest.java
@@ -168,13 +168,8 @@ public class LoadBeanRequest extends LoadRequest {
         // necessary but allow processing to continue until it is accessed by client code
         Object id = desc.getId(ebi.getOwner());
         if (!loadedIds.contains(id)) {
-          if (desc.isSoftDelete()) {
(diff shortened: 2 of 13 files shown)


rbygrave:
Ok, this looks good.  I'm going to re-read it again later.  Cheers, Rob.

PR #1531, opened 2018-11-07
Transform extra-ddl.xml files when running in ddl mode (no migration)

focdanisch:
Hi Rob,

during our junit tests we faced a problem that ebean does not substitute variables (${...}) in extra-ddl.xml files when ebean is not running in migration mode (our junit tests run in ddl creation mode only). We fixed the problem temporarily by copying PlaceholderBuilder and ScriptTransform from the io.ebean.migration.runner package and integrating them into DdlGenerator.

Could you please take a look at the PR? Maybe you have an idea for a better solution (e.g., make the classes/methods in io.ebean.migration.runner public so that they can be used from other packages as well). With the changes in this PR, our junit tests succeed in ddl mode.

Best regards,

Manuel

### DIFF
--- a/src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
+++ b/src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
@@ -1,7 +1,9 @@
 package io.ebeaninternal.dbmigration;
 
+import io.ebean.config.DbMigrationConfig;
 import io.ebean.config.ServerConfig;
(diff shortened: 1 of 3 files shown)


rbygrave:
Yup, getting around to looking at this now.  

> make the classes/methods in io.ebean.migration.runner public so that ...

Yes, it is probably a good plan to make them public and reuse them.

I should be looking at this tonight sometime.

PR #1539, opened 2018-11-10
FIX: Pairs must not be modified by query

rPraml:
`Ebean.find(..).where().inPairs(pair)` may modify the passed value. So executing the same query twice may produce differen results.

I think you will agree, that this is a no-go. A query must not modify any passed value.

### DIFF
--- a/src/main/java/io/ebeaninternal/server/expression/InExpression.java
+++ b/src/main/java/io/ebeaninternal/server/expression/InExpression.java
@@ -48,8 +48,16 @@ class InExpression extends AbstractExpression {
 
   @Override
   public boolean naturalKey(NaturalKeyQueryData<?> data) {
-    // can't use naturalKey cache for NOT IN
-    return !not && data.matchIn(propName, bindValues);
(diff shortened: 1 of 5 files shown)


src/main/java/io/ebeaninternal/server/expression/InExpression.java
```diff
@@ -48,8 +48,16 @@
 
   @Override
   public boolean naturalKey(NaturalKeyQueryData<?> data) {
-    // can't use naturalKey cache for NOT IN
-    return !not && data.matchIn(propName, bindValues);
+    // can't use naturalKey cache for NOT IN or if multi values are used
```
rbygrave: >  or if multi values are used

Ah?

rbygrave: > || multiValueSupported

That does not look correct to me.

rbygrave: Cool. I'll look to merge and adjust that ...

PR #1540, opened 2018-11-10
FIX for the limit query

rPraml:
there are differen meanings, what setMaxRows(0) means.
- does it mean all rows?
- or does it mean zero rows?

in conjunction with setFirstRow, it may happen, that an "offset 3 limit 0" query is generated (it depends on the SqlLimiter implementation)

So, in this PR, I checke all SqlLimiters. They will ignore the maxRows property if it is zero and if firstRow is already set.

A better solution might be, if we use -1 (or MAX_INT) as default value for maxRows. Unfortunately, there are many places, with checks like `maxRow != 0` , `maxRow > 0` , `maxRow >= 1`...

On the other side, a query with maxRows = 0 makes no sense, it will always be empty

OT: The HanaSqlLimiter seems to be the same as the default  LimotOffsetSqlLimiter

### DIFF
--- a/src/main/java/io/ebean/config/dbplatform/LimitOffsetSqlLimiter.java
+++ b/src/main/java/io/ebean/config/dbplatform/LimitOffsetSqlLimiter.java
@@ -31,12 +31,11 @@ public class LimitOffsetSqlLimiter implements SqlLimiter {
     int firstRow = request.getFirstRow();
     int maxRows = request.getMaxRows();
 
-    if (maxRows > 0 || firstRow > 0) {
+    if (maxRows > 0) {
       sb.append(" ").append(LIMIT).append(" ").append(maxRows);
-      if (firstRow > 0) {
-        sb.append(" ").append(OFFSET).append(" ");
-        sb.append(firstRow);
-      }
+    }
+    if (firstRow > 0) {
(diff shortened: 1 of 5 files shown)


rPraml:
@rbygrave I think I was a litte bit over-eager. This breaks the Maria-DB / MySQL support now, as they do not allow offset without limit: "SELECT * FROM table OFFSET 3"

https://dev.mysql.com/doc/refman/8.0/en/select.html#id4651990

I think we should revert this change. And maybe we either initialize the limit with -1 or throw an exception if you set offset only.


| limit | offset | Old impl | Current impl   | logically Expected |
|------|------|-----------|----------------|-------------------|
|   0    |    0   | return all rows   | return all rows |   return 0 rows |
|   0    |   10  | limit 0 offset 10 (no rows) | offset 10 (broken) |  return 0 rows |
|   10    |   10  | rows 11-20 | rows 11-20 |  rows 11-20 |
|   10    |   0  | rows 1-10 | rows 1-10 |  rows 1-10 |
|   -1    |    0   | ?  | ? |   return all rows |
|   -1    |    10   | ?  | ? |   return all rows starting with row 10 (seems not very easy in MySQL)|

rbygrave:
Right, looking ...

PR #1542, opened 2018-11-10
Filter and ExpressionList share a common API

rPraml:
Hello Rob,

this is what I'm currently working on, I've improved the filters, so that you can convert a query to a filter and perform the filtering in-memory.

To keep the PRs small, this PR is the first step, others will follow (if you want one big PR with all changes, I can do this also)

So, the following shoud work for all beans, that are not too complex (no raw sql, temporal, and so on)

```java
List dbResult = anyQuery.findList();

Filter filter = anyQuery.filter();
List allBeans = Ebean.find(anyQuery.getBeanType());
List filteredBeans = filter.filter(allBeans);

assertThat(filteredBeans).isEqalTo(dbResult);
```

### DIFF
--- a/src/main/java/io/ebean/ExpressionList.java
+++ b/src/main/java/io/ebean/ExpressionList.java
@@ -38,7 +38,7 @@ import java.util.function.Predicate;
  *
  * @see Query#where()
  */
-public interface ExpressionList<T> {
+public interface ExpressionList<T> extends QueryDsl<T, ExpressionList<T>> {
 
   /**
    * Return the query that owns this expression list.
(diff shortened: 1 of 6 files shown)


rPraml:
Needs work. It seems that moving the method in other class breaks existing code (like querybean package, which needs to be recompiled)

rbygrave:
Ah, I should ask what the motivation for this is.  There is a chance I reject this idea if I think it hurts the API.  I don't place much if any priority on the in-memory DSL having compatibility to the query language.

rPraml:
>Ah, I should ask what the motivation for this is.

I have a List<> of beans, and I want to sort and filter them (the beans in the list are not stored in the database or come from the database are updated in-memory and not yet persisted)

Currently, we have a very complex mechanism in our application, that wraps the Query and implements its own filter logic (as the current ebean filter was not sufficient)
Our logic cannot use queryBeans.

With this enhancement, I'm able to write simple filters with QueryBeans like this:
```
Query<MyBean> = new QMyBean().where().name.eq("Roland").query();
Filter<MyBean> filter = query.filter();
List<MyBean> filteredList = filter.filter(unfilteredList);
```

> There is a chance I reject this idea if I think it hurts the API.  I don't place much if any priority on the in-memory DSL having compatibility to the query language.

I understand. Breaking the API is also no my intention. So I keep all methods in Filter / ExpressionList. Otherwise it will break existing code like query-beans (maybe they can be removed with the next major version)

src/main/java/io/ebeaninternal/server/el/ElFilterNode.java
```diff
@@ -0,0 +1,585 @@
...
+
+//  /**
+//   * Simplify nested expressions where possible.
+//   * <p>
+//   * This is expected to only used after expressions are built via query language parsing.
+//   * </p>
+//   */
+//  @SuppressWarnings("unchecked")
+//  @Override
+//  public void simplify() {
```
rbygrave: Keen to hear if this is handled somewhere else now ?

rbygrave:
> With this enhancement, I'm able to write simple filters with QueryBeans like this:

Yes, I see.

This PR is large so it takes a bit of reading.  I've looked through it but it will take a few more looks.  It also has comflicts now probably due to merging the other PR's first.  Let me know when it's resolved and I'll try to have another look.

PR #1548, opened 2018-11-14
add global configuration disableLazyLoading

ooknight:


rbygrave:
FYI:  I'll get to merge this in shortly (hopefully tonight).  There are tests with a specific EbeanServer that we can move this test to (I'll do that).

PR #1549, opened 2018-11-16
BeanType, Property and EBI can hold a custom object

rPraml:
This feature allows the application to store a customObject on 
- BeanType
- BeanPropery
- EntityBean(Intercept)

This can be used to store additional information per BeanType/Property. e.g. from own annotations.
The customObject in the entity-bean can be used in a similar way, to store meta-information per bean.

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -104,6 +104,11 @@ public final class EntityBeanIntercept implements Serializable {
   private int sortOrder;
 
   /**
+   * Custom object, to attach any object to a bean.
+   */
+  private Object customObject;
+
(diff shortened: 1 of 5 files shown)


rPraml:
Currently, we have the same concept in our app (descriptor with properties) where we store our additonal meta information from our annotations per bean/property. 

So we implemented also mechanisms like `getPropertyByPath(String elPath)` that finds a property by its el-path (which is rather complex to traverse relations etc)

With this change, we would be able to attach this information to the existing ebean structure. So we would at least need the customObject in "BeanType" and "BeanPropery"

The custom object in the EBI is rarely used and not so important (I can use a transient property in the beans where I need this)

rPraml:
I'm closing this (I will reopen it, if we really need it)

PR #1551, opened 2018-11-16
new comparison method Ine (caseInsensitive notEquals)

rPraml:
Adds the comparison method "Ine" to perform a case insensitive not equals


src/main/java/io/ebeaninternal/server/expression/DocQueryContext.java
```diff
@@ -48,6 +48,13 @@
    */
   void writeIEqualTo(String propName, String value) throws IOException;
 
+  /**
+   * Write a case insensitive notEqualTo expression.
+   */
+  default void writeINotEqualTo(String propName, String value) throws IOException {
+    throw new AbstractMethodError("writeINotEqualTo not implemented");
```
rPraml: TODO: The ElasticSearch API should be extended by this method

PR #1584, opened 2018-12-20
Query cache bug, which produces wrong results

rPraml:
Hello Rob,

we discovered a strange issue, where one query produces wrong results, which we could not explain.

Sometimes, the result was correct, sometimes not. We investigated a bit, and it seems that the result depends, if a similar query was executed first, so I expect this may be a query/queryplan cache issue.

We stripped down the object graph and query from our application as far as possible to a test case.

Please see "Main1" & "Main2" class, where you should be able to reproduce the issue.

### DIFF
--- a/src/test/java/org/tests/query/cache/Acl.java
+++ b/src/test/java/org/tests/query/cache/Acl.java
@@ -1,0 +1,20 @@
+package org.tests.query.cache;
+
+import javax.persistence.Entity;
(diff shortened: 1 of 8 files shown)


rbygrave:
At this point the issue is very likely that conceptually `SpiQuery.setSqlDistinct(true)` ... should not be on the query itself but instead on the query plan (likely end up on SqlTree).  I will confirm ...

PR #1593, opened 2019-01-03
TempFiles will be deleted automatically if the reference is gc'ed

rPraml:
This feature enhances the ScalarTypeFile, so that temp files are deleted if the file is no longer referenced.

**Old behaviour**
TempFiles created by ScalarTypeFile were never deleted by java code (relies on the OS, that the temp dir will be cleared). 
Reading a lot of entities, that contains files could cause an out-of-disk space

**New behaviour**
There are two tempFileProviders:

_1. DeleteOnShutdownTempFileProvider_
This one will delete all created temp files on Ebean.shutdown().

Note 1: Ebean.shutdown normally occurs earlier than File.deleteOnExit()
Note 2: In a long running web application, microservice etc, Ebean.shutdown occurs rarely (=nearly never)

_2. WeakRefTempFileProvider_
This provider tracks a weak reference to the _File_ object, that is returned by ScalarTypeFile. 

As long as this file object is NOT gc'ed, the according temp file will NOT be deleted.
As the File object is referenced by the entity, it is ensured, that the temp-file will not be deleted, as long as you have a reference to the entiy. (which is sufficient for most cases)

Note: If you convert the _File_ to a _Path_ or a _String_, (although if you convert it back to _File_) you will give up the reference to the original _File_ object. So the WeakFileReference will be queued by the next gc run and the file may be deleted, so that your _Path_ or _String_ will point to a non existent file.
This is a bit non-deterministic, so if there are too many problems with the WeakRefTempFileProvider (which is the current default), you probably should use the DeleteOnShutdownTempFileProvider or implement your own provider.

### DIFF
--- a/src/main/java/io/ebean/config/DeleteOnShutdownTempFileProvider.java
+++ b/src/main/java/io/ebean/config/DeleteOnShutdownTempFileProvider.java
@@ -1,0 +1,68 @@
+package io.ebean.config;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * TempFileProvider implementation, which deletes all temp files on shutdown.
+ *
+ * @author Roland Praml, FOCONIS AG
+ *
+ */
+public class DeleteOnShutdownTempFileProvider implements TempFileProvider {
+
+  private static final Logger logger = LoggerFactory.getLogger(DeleteOnShutdownTempFileProvider.class);
+
+  List<String> tempFiles = new ArrayList<>();
+  private final String prefix;
+  private final String suffix;
+  private final File directory;
+
+  /**
+   * Creates the TempFileProvider with default prefix "db-".
+   */
+  public DeleteOnShutdownTempFileProvider() {
+    this("db-", null, null);
+  }
+
+  /**
+   * Creates the TempFileProvider.
+   */
+  public DeleteOnShutdownTempFileProvider(String prefix, String suffix, File directory) {
+    this.prefix = prefix;
+    this.suffix = suffix;
+    this.directory = directory;
+  }
+
+  @Override
+  public File createTempFile() throws IOException {
+    File file = File.createTempFile(prefix, suffix, directory);
+    synchronized (tempFiles) {
+      tempFiles.add(file.getAbsolutePath());
+    }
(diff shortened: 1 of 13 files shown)


rbygrave:
I'm not sure if this is a good idea.

The File is exposed to the application code and the application code can move the file and or delete the file when the application is done processing it etc.  Why not leave it to application code?

rPraml:
> Why not leave it to application code?

The application code does not know it. E.g. you have an entity `User` and you will add a `File picture` (probably with fetch=eager), you need to extend every code in your application, that uses the User entity, so the file is deleted afterwards.

> The File is exposed to the application code and the application code can move the file and or delete the file.

I only agree partially with this. The file is in the temp directory, so the application cannot rely that the file will be there forever, especially, after an application restart. So the `DeleteOnShutdownTempFileProvider` should cause no problems here.

If the application moves the file away, the file should be _safe_ as both temp file providers will store the creation-filename as string:
```
      for (String path : tempFiles) {
        if (new File(path).delete()) {
          logger.trace("deleted {}", path);
        } else {
          logger.warn("could not delete {}", path);
        }
      }
```
The only thing that can be optimized here, is a check, if the file still exists. I'll do that and provide a test case.

rbygrave:
> The application code does not know it. E.g. you have an entity User and you will add a File picture (probably with fetch=eager), you need to extend every code in your application, that uses the User entity, so the file is deleted afterwards.

I'd disagree.  File is a Lob ... and that should never really be fetch=eager basically.  I'd only be fetching File/Lob when the app is going to use it.

At this stage I'm highly unlikely to ever merge this into master.

rPraml:
> I'd disagree. File is a Lob ... and that should never really be fetch=eager basically. I'd only be fetching File/Lob when the app is going to use it.
> 
> At this stage I'm highly unlikely to ever merge this into master.

I understand your concerns and maybe I also have a rare use case:

I have a bean with attachment (e.g. profile pic), that is in the HttpSession of a web application. Whenever the browser requests the profile pic, the webserver will send a datastream from the bean's file.
So, to delete the file I would have to hook on the web application. (or delete the file after each request, which means, that I have to fetch the bean directly from the database)
 
Woud you at least accept the changes with the TempFileProvider with a default implementation, that never (or only on shutdown) deletes the files, so that I can set my own implementation in serverconfig.

Cheers Roland

rbygrave:
I think we should close this PR.

PR #1594, opened 2019-01-03
Support loading of beans with invalid JSON content

rPraml:
**Use Case:**
You have a non-parseable JSON in your database. This may happen if your JSON-DAOs can not be mapped, because their implementation has changed. There is no (easy) way, to load that bean even partially.

**Old behaviour**
- You will get an SQLException, PersistenceException or TextException, if JSON could not be mapped. 
- You could not load that bean (even not partially)

**New behaviour**
- Whenever text could not be converted to JSON (or other types like URI), a TextException is thrown.
There is a `TextException.getText()` to provide access to the causing JSON (so you may code a recovery strategy for this)
- You can set an other LoadErrorHandler, that is invoked, if a property could not be loaded. 

The default LoadErrorHandler throws a PersistenceException and the bean would not be loaded.

If you set `ebean.failOnLoadError=false` the exceptions will go to the log and the bean will be loaded at least partially. `BeanState.getLoadErrors()` will provide you access to the exceptions.

So now we have ability, to load corrupt beans parially and notify the user, that there were load-errors.

You can also implement own recovery strategies in your LoadErrorHandler

### DIFF
--- a/src/main/java/io/ebean/config/ServerConfig.java
+++ b/src/main/java/io/ebean/config/ServerConfig.java
@@ -30,9 +30,15 @@ import io.ebean.event.readaudit.ReadAuditLogger;
 import io.ebean.event.readaudit.ReadAuditPrepare;
 import io.ebean.meta.MetaInfoManager;
 import io.ebean.migration.MigrationRunner;
+import io.ebean.plugin.LoadErrorHandler;
 import io.ebean.util.StringHelper;
 
+import javax.persistence.PersistenceException;
 import javax.sql.DataSource;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -517,6 +523,10 @@ public class ServerConfig {
    * When true we do not need explicit GeneratedValue mapping.
    */
   private boolean idGeneratorAutomatic = true;
+
+  private LoadErrorHandler loadErrorHandler = (bean, prop, fullName, e) -> {
+    throw new PersistenceException("Error loading on " + fullName, e);
+  };
 
   /**
    * Construct a Server Configuration for programmatically creating an EbeanServer.
@@ -2985,6 +2995,13 @@ public class ServerConfig {
 
     String mappingsProp = p.get("mappingLocations", null);
     mappingLocations = getSearchList(mappingsProp, mappingLocations);
+
+    if (!p.getBoolean("failOnLoadError", true)) {
+      Logger logger = LoggerFactory.getLogger("io.ebean.SQL");
+      loadErrorHandler = (bean, prop, fullName, e) -> {
+        logger.error("Error loading on {}", fullName, e);
+      };
+    }
   }
 
   private NamingConvention createNamingConvention(PropertiesWrapper properties, NamingConvention namingConvention) {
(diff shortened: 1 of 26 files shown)


src/main/java/io/ebeaninternal/server/deploy/DynamicPropertyAggregationFormula.java
```diff
@@ -49,15 +49,16 @@ public Object read(DataReader dataReader) throws SQLException {
...
-        sqlBeanLoad.load(asTarget, value);
-      }
+      value = scalarType.read(sqlBeanLoad.ctx().getDataReader());
 
     } catch (Exception e) {
-      throw new PersistenceException("Error loading on " + fullName, e);
+      sqlBeanLoad.ctx().handleLoadError(null, asTarget, fullName, e);
+      return;
+    }
+    if (asTarget != null) {
```
rbygrave: Why did this move?

src/main/java/io/ebeaninternal/server/query/SqlBeanLoad.java
```diff
@@ -88,8 +86,9 @@ public Object load(BeanProperty prop) {
       return dbVal;
 
     } catch (Exception e) {
```
rbygrave: We should probably only setLoadError on TextException ... and let everything else fail by throwing PersistenceException

src/main/java/io/ebeaninternal/server/deploy/DynamicPropertyAggregationFormula.java
```diff
@@ -49,15 +49,16 @@ public Object read(DataReader dataReader) throws SQLException {
...
-
+    Object value;
     try {
-      Object value = scalarType.read(sqlBeanLoad.ctx().getDataReader());
-      if (asTarget != null) {
-        sqlBeanLoad.load(asTarget, value);
-      }
+      value = scalarType.read(sqlBeanLoad.ctx().getDataReader());
 
     } catch (Exception e) {
```
rbygrave: Only on TextException ?

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -2985,6 +2995,13 @@ protected void loadSettings(PropertiesWrapper p) {
 
     String mappingsProp = p.get("mappingLocations", null);
     mappingLocations = getSearchList(mappingsProp, mappingLocations);
+
+    if (!p.getBoolean("failOnLoadError", true)) {
```
rbygrave: failOnLoadJsonError ?

src/main/java/io/ebean/config/ServerConfig.java
```diff
@@ -518,6 +524,10 @@
    */
   private boolean idGeneratorAutomatic = true;
 
+  private LoadErrorHandler loadErrorHandler = (bean, prop, fullName, e) -> {
```
rbygrave: I think this should instead be on the query itself rather than global.  query().setAllowLoadErrors().where()....findOne()

src/main/java/io/ebean/config/ServerConfig.java
rPraml: It does not affect only to json errors. It also may handle also unparseable Durations and so on, that were stored as String

src/main/java/io/ebean/config/ServerConfig.java
rPraml: Does it also trigger on a lazy load then? (I think in my use case it could be OK to have that on the query - have to check)

src/main/java/io/ebeaninternal/server/deploy/DynamicPropertyAggregationFormula.java
rPraml: ScalarTypePeriod may throw a DateTimeParseException

src/main/java/io/ebeaninternal/server/deploy/DynamicPropertyAggregationFormula.java
rPraml: Do you mean, why I moved this out of the exception block?
I don't see when `sqlBeanLoad.load(asTarget, value);` can throw an exception that is worth to catch. (Of course, NPE, when the bean is null)

PR #1595, opened 2019-01-03
alternative dirty detection on mutable properties

rPraml:
This PR provides an alternative dirty detection.

We had the problem, that complex entities (e.g. List of json objects) were not updated properly as ebean did not realize, that the object has changed or were updated (but should not) because ebean updates some types always (e.g. JsonNode is always dirty). This results in unneccessary updates and unneccessary version increments.


soberich:
+1

rbygrave:
Looking to close this PR.  The simple approach is to use MutableAware types rather than JsonNode etc.

PR #1601, opened 2019-01-04
FIX: possible NPE in DeleteUnloadedForeignKeys.deleteCascade and in i…

rPraml:
…nitPostTarget

### DIFF
--- a/src/test/java/org/tests/family/TestInheritance.java
+++ b/src/test/java/org/tests/family/TestInheritance.java
@@ -1,6 +1,7 @@
 package org.tests.family;
 
 import io.ebean.BaseTestCase;
+import io.ebean.Ebean;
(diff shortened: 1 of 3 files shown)


src/test/java/org/tests/family/TestInheritance.java
```diff
@@ -218,6 +219,10 @@ public void testInheritance() {
     assertEquals("Foo", grandparent1.getFamilyName());
     assertEquals("Munich", grandparent1.getAddress());
     assertEquals(1, grandparent1.getEffectiveBean().getId().intValue());
+
+    Ebean.find(ChildPerson.class).delete();
```
rPraml: tried to clean up here and triggered the NPE in DeleteUnloadForeignKeys

PR #1609, opened 2019-01-14
FIX: insert in batch OneToOne relationship

ajcamilo:
When we have multiple `@OneToOne` relationships with `Cascade.PERSIST` in one entity, the `BatchControl` would simple look to the depth of the first relation, which in some cases would make the inserts out of order, and cause constraint violations.

The fix will make the entity have the maximum depth of all the `@OneToOne` relationships plus 1.

This also includes a test case for this: `TestBatchInsertOneToOne`


rbygrave:
Nice - well done !!

PR #1635, opened 2019-02-08
Fixed missing EBEAN_PROPS_FILE loading

mrmx:
Trying to honor EBEAN_PROPS_FILE env property found that bug..


rbygrave:
Thanks !!

PR #1637, opened 2019-02-10
build on jdk11

rPraml:
Hello Rob,

please take a look at these changes, they would add JDK11 build support.
- They require a new agentloader: https://github.com/avaje-common/avaje-agentloader/pull/6
- There is the same bug in javadoc generation, so skip javadoc generation

cheers
Roland

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,6 +188,22 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- Provided scope to compile with java 11 -->
+    <dependency>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -222,7 +238,7 @@
     <dependency>
       <groupId>org.avaje</groupId>
       <artifactId>avaje-agentloader</artifactId>
-      <version>2.1.2</version>
+      <version>3.0.0-SNAPSHOT</version>
```
rPraml: Requires a release first

Issue #1630, opened 2019-02-05
ENH: Improve jdbc batch use when persisting ManyToMany intersection changes


rbygrave:
In essence currently Ebean is flushing more than ideal when batch persisting beans that have ManyToMany changes. 

This change would queue and batch the ManyToMany intersection changes to get better jdbc batch utilisation.

PR #1638, opened 2019-02-11
 FIX: When saving multiple m2m beans, only last m2m wins (Regression due #1630)

rPraml:
When saving multiple m2m relations, only the last relation did win.


rbygrave:
Cool - thanks !!!

Issue #1640, opened 2019-02-17
For Postgres support mapping InetAddress to INET db column


rbygrave:
Note that Postgres INET DB column supports IpV6 address types + address ranges.

For this we are adding the `ebean-types` project that contains `io.ebean.type.Inet` and `io.ebean.type.Cdir`.  These are value types that:
- Wrap string values of IP addresses and ranges
- Don't have any validation (that is expected to be done externally)
- Don't have any DNS lookup or network validation (unlike some use of InetAddress).

That is, `io.ebean.type.Inet` and `io.ebean.type.Cdir` are types that can be used instead of `java.net.InetAddress` (and for Postgres these map to `INET` and `CDIR` database types.

yholkamp:
Regarding the note, the [PostgreSQL docs](https://www.postgresql.org/docs/11/datatype-net-types.html) note that the `inet` column maybe be used to store both IPv4 and IPv6 addresses, there may be another reason the test is failing in this scenario.

rbygrave:
@yholkamp   Quite right.  There was a problem with the IPv6 conversion being used.  Test using INET with IPv6 address working now.

java.net.InetAddress looks like a problem for range addresses like `2001:4f8:3:ba::/64` and `2001:4f8:3:ba:2e0:81ff:fe22:d1f1/128`.  

I am thinking it might be good to have a type (value type) that we could use instead of java.net.InetAddress that works in raw string values avoiding any dns lookup and supports all valid `inet` input such as ranges.  For this `Inet value type` DNS lookup and validation would be external to it ... developers would translate it to java.net.InetAddress if/when needed.

PR #1641, opened 2019-02-17
#1640 - For Postgres support mapping InetAddress to INET db column

rbygrave:


rbygrave:
With this second commit, this PR is now ready.

PR #1668, opened 2019-04-08
Fix check constraint support for MariaDB >= 10.2.1

rPraml:
Starting with MariaDB 10.2.1, the "check constraint" on columns are honored.
This conflicts with the migration scripts. See details https://groups.google.com/forum/#!topic/ebean/luFN-2xBkUw

So, this patch disables check constraints by default. It is currently possible to enable check constraints with the system property `ebean.mysql.useCheckConstraint` - but this is only recommended, if you use  mysql >=8.0.16 or mariadb  10.2.1


rbygrave:
Good stuff - thanks.

Issue #1447, opened 2018-07-05
Delete query ... should honour cascading delete if defined


rbygrave:
Unit cases added by PR #1445

### Steps to reproduce

Deleting OtoUser ... should cascade delete to OtoUserOptional due to the cascade ALL on the relationship.

The cascade should happen with both delete a bean and delete by query. 

```java
@Entity
@Table(name = "oto_user_model")
public class OtoUser extends BaseModel {

  String name;
  
  @OneToOne(optional = true, cascade = CascadeType.ALL)
  OtoUserOptional userOptional;
```

```java

OtoUser someUser = new OtoUser("someUser");
OtoUserOptional userPref = new OtoUserOptional(...);
...

Ebean.find(OtoUser.class)
  .where() ...
  .delete();

// should delete the associated OtoUserOptional
```

PR #1669, opened 2019-04-08
When deleting complex object graphs, entities are deleted multiple times

rPraml:
When you have a complex object graph, with one2many relations and cascade=REMOVE on both sides, you cannot delete the graph, if you have more than one relation
 
The test `testDeleteMultiMasterTxn()` shows this issue. It fails with an OptimisticLockException in https://github.com/ebean-orm/ebean/blob/f928923b5b380c1dafd261fac7c7cfaef2387f31/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java#L876

Note: 
- you need at least two master beans
- It does only happen in a transaction.
- It only happens, when a version-property is in the entity

So in the test, there are two RelMaster pointing to one RelDetail.
When deleteing all RelMaster, it deletes the dependent RelDetail, which deletes all RelMasters, which deletes all RelDetails...
In the SQLLog, it seems that the RelDetail (1,1) is deleted two times
```java
-- DeleteById of RelMaster ids[[1, 2]] requires fetch of foreign key values
select t0.id, t0.detail_id from rel_master t0 where t0.id in (?, ? ) ; --bind(Array[2]={1,2})
FindMany type[RelMaster] origin[Cjorkz.A.A] exeMicros[3359] rows[2] predicates[t0.id in (?, ? ) ] bind[Array[2]={1,2}]
delete from rel_master where id=?; -- bind(1)
Deleted [RelMaster] [1]
select t0.id from rel_master t0 where detail_id=? ; --bind(1)
FindAttr exeMicros[313] rows[1] type[RelMaster] predicates[detail_id=? ] bind[1]
-- DeleteById of RelMaster ids[[2]] requires fetch of foreign key values
select t0.id, t0.detail_id from rel_master t0 where t0.id in (? ) ; --bind(Array[1]={2})
FindMany type[RelMaster] origin[Cjorkz.A.A] exeMicros[260] rows[1] predicates[t0.id in (? ) ] bind[Array[1]={2}]
BatchControl flush [RelMaster:100 d:1, RelDetail:101 d:1]
delete from rel_master where id=?
 -- bind(2)
Deleted [RelMaster] [2]
delete from rel_detail where id=? and version=?
 -- bind(1,1)
Deleted [RelDetail] [1]
delete from rel_master where id=?; -- bind(2)
Deleted [RelMaster] [2]
select t0.id from rel_master t0 where detail_id=? ; --bind(1)
FindAttr exeMicros[181] rows[0] type[RelMaster] predicates[] bind[1]
BatchControl flush [RelDetail:100 d:1]
delete from rel_detail where id=? and version=?
 -- bind(1,1)
---> OptimisticLockException
```

It seems that the behaviour has changed with #1447 - before, no OptimisticLockException was thrown (but dependent entities were also not deleted)

### DIFF
--- a/src/test/java/org/tests/cascade/RelDetail.java
+++ b/src/test/java/org/tests/cascade/RelDetail.java
@@ -1,0 +1,44 @@
+package org.tests.cascade;
+
+import java.util.List;
+
+import javax.persistence.CascadeType;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.OneToMany;
+import javax.persistence.Version;
+
+@Entity
+public class RelDetail {
+
+  @Id
+  Long id;
+
+  String name;
+
+  @Version
+  int version;
+
+  @OneToMany(cascade = CascadeType.ALL, mappedBy = "detail")
+  private List<RelMaster> masterRel;
+
+  public Long getId() {
+    return id;
+  }
+
+  public void setId(Long id) {
+    this.id = id;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  public List<RelMaster> getMasterRel() {
+    return masterRel;
+  }
+}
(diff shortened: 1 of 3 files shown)


rbygrave:
So yes the fix is to skip the delete of the second attempt to delete the detail bean.  We can do this by not removing the "deleted bean hash" per object graph.

PR #1687, opened 2019-04-27
Fix implicit conversion from array to string

thibaultmeyer:
According to other object implementing `BindCaptureEntry`. Don't use a simple `String.valueOf()` for one-dimension array. When converting an array to a readable string, use Arrays.toString for one-dimensional arrays. This function iterate over the contents of the array and produce human-readable output.


rbygrave:
Cool - thanks !!

PR #1688, opened 2019-04-28
Remove useless comparison test

thibaultmeyer:
The result of certain comparison tests can sometimes be inferred from their context and the results of other comparisons. Inspect the code to check whether the logic is correct, and consider simplifying the logical expression.


rbygrave:
Good stuff, thanks !

Issue #1695, opened 2019-05-08
For MySql and SQL Server LIKE remove the use of binary keyword / collation


rbygrave:
The binary keyword for MySql like was brought in as part of: https://github.com/ebean-orm/ebean/pull/1162/files

We are going to treat the addition of the binary keyword as a bug.  The fix will be to remove the binary keyword from the MySql like cause.

See discussion on google group at:
https://groups.google.com/d/msgid/ebean/c2b9f87c-0983-44bf-a4b4-019fef2baa15%40googlegroups.com?utm_medium=email&utm_source=footer

rPraml:
#1162 was introduced by us
We use 'utf8mb4_bin' as DB default in the meantime for all our backend operations (so binary should be used autmatically) - with other collations, like 'utf8mb4_german2_ci' we have strange effects, as 'ä' == 'ae' or 'ss' == 'ß'
We also introduced #1492 so that we can specify special collations when sorting for UI.

I have no objection removing the 'binary' option. But then we should it remove for all platforms (There is a similar workaround in the sqlserver like)
This means:
- like use the db-default collation
- ilike performs a tolower
- the same should count for eq/ieq/...

So, if you using a 'non-case-sensitive' collation (=case-insensitive, like utf8mb4_german2_ci) the result of "like" and "ilike" should be the same. 


Maybe it would be also a good idea to specify the 'ilike' collation in serverconfig, because a tolower(...) is also not always correct (see http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html)
So: "like" should use DB default collation (which can be case sensitive) and "ilike" should use the one in server-config

PR #1697, opened 2019-05-08
#1695 - For MySql like remove the binary keyword + Add caseSensitiveCollation configuration option

rbygrave:


rbygrave:
I believe this PR is ready now to merge.  Would you able to review it for me @rPraml  ?

src/test/java/org/tests/transaction/TestBatchModelFlush.java
```diff
@@ -75,17 +75,22 @@ public void multipleTopLevel_expect_singleFlush() {
...
 
     // first saved to batch - (depth 100)
+    boolean hasSequence = isSqlServer();
     assertThat(sql.get(0)).contains("insert into mny_b");
-    assertThat(sql.get(1)).contains(" -- bind(BatchMultipleTop_0");
-    assertThat(sql.get(2)).contains(" -- bind(BatchMultipleTop_1");
+    if (!hasSequence) {
+      assertThat(sql.get(1)).contains(" -- bind(BatchMultipleTop_0");
+      assertThat(sql.get(2)).contains(" -- bind(BatchMultipleTop_1");
+    }
```
rPraml: I'm just wondering, how these tests could pass in my local test env. My connection settings are
```
ebean.mssql.idType=SEQUENCE
# Database has Collation: Latin1_General_100_BIN2
datasource.mssql.databaseUrl=jdbc:sqlserver://srv-01-docker02.foconis.local:1433;databaseName=ebean_unittest_bin;sendTimeAsDatetime=false
datasource.mssql.databaseDriver=com.microsoft.sqlserver.jdbc.SQLServerDriver
```
And I explicit enabeld SEQUENCE for SqlServer (by default, identity is used)
regarding your config file, SEQUENCE is commented out
https://github.com/ebean-orm/ebean/blob/master/src/test/resources/ebean.properties#L157

So sqlserver means *no sequencence* in your setup.
Nevertheless it would be better to use `spiEbeanServer().getServerConfig().getIdType() == IdType.SEQUENCE` instead of relying against a certain platform

src/test/resources/ebean.properties
```diff
@@ -160,8 +160,8 @@ ebean.sqlserver.databasePlatformName=sqlserver17
 
 datasource.sqlserver.username=test_ebean
 datasource.sqlserver.password=SqlS3rv#r
-datasource.sqlserver.databaseUrl=jdbc:sqlserver://localhost:1433;databaseName=test_ebean
-datasource.sqlserver.databaseDriver=com.microsoft.sqlserver.jdbc.SQLServerDriver
+datasource.sqlserver.url=jdbc:sqlserver://localhost:1433;databaseName=test_ebean;sendTimeAsDateTime=false
```
rPraml: Ah, the sendTimeAsDateTime flag ;) https://github.com/microsoft/mssql-jdbc/pull/798

rbygrave:
> We use 'utf8mb4_bin' for mysql and 'Latin1_General_100_BIN2' for SqlServer

Quite right - I'd do the same myself given the choice.  

I believe the current implementation of `isPlatformCaseSensitive()` needs to actually move to the platform itself (for at least testing purposes) - something like platform.isCaseSensitiveCollation().  We can then set that such that the tests pass for different collation setups. 

Hmmm, can we determine the case sensitivity of the collation dynamically - is that a good idea?

rPraml:
Autodetection seems to be a good idea at the first place, but it is not so trivial as I thought

On SqlServer you can perform
`select case when ('a' = 'A') then 1 else 0 end`

On mysql I tried several things
`select * from (select 'a' as c1) as x where x.c1 = 'A'`
`select * from db_migration where 'a' = 'A'`

but it seems that the check runs against the defaut mysql configuration as long as there is no column involved.

this check also does not work:
`select * from information_schema.schemata where schema_name = (select database() from dual)`
as `database()` is always pointing to information_schema instead to current database

I think other platforms (DB2, ...) have different surprises. 
So maybe we drop the idea of auto-detection and add the flag to platform, which has to be configured for the test setup in serverconfig/platformconfig

rbygrave:
> So maybe we drop the idea of auto-detection and add the flag to platform, which has to be configured for the test setup in serverconfig/platformconfig

Yes, I agree.

rbygrave:
Ok, I'm happy with this now. Looking to merge it.

rbygrave:
> ebean.mssql.idType=SEQUENCE

Hi @rPraml  ... note that the `ebean.sqlserver.databasePlatformName=sqlserver17` property ... sets the platform to sqlserver17 and that platform defaults to idType=SEQUENCE (and I'm testing against SqlServer 2017 with sequences).

PR #1698, opened 2019-05-08
FIX for IAE "key can't be empty"  in System.getProperty() 

rPraml:
*How to reproduce*
use a property placeholder like `${}` somewhere in the config file

### DIFF
--- a/src/main/java/io/ebean/config/properties/PropertyEval.java
+++ b/src/main/java/io/ebean/config/properties/PropertyEval.java
@@ -58,7 +58,9 @@ final class PropertyEval {
 	 * or existing an property in SystemProperties itself.
 	 */
 	private static String evaluateExpression(String exp) {
-
+	    if (exp.isEmpty()) {
+	        return null;
+	    }
 		if (isJndiExpression(exp)) {
 			// JNDI property lookup...
 			String val = getJndiProperty(exp);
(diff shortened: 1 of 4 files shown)


rPraml:
@rbygrave changed file should be reformatted, as it uses tabs instead of spaces

rbygrave:
Why do we think `${}` is valid ?  Should it not through an IAE ?

rPraml:
I use it for collation setting. e.g. `lcase(${})` which will be translated to an `order by lcase(property)` statement.
See: https://github.com/ebean-orm/ebean/blob/master/src/main/java/io/ebean/OrderBy.java#L336
(maybe the `${}` placeholder was not the best choice here)

rbygrave:
I'm confused.  OrderBy uses StringHelper but not PropertyEval so I don't see how the suggested change impacts OrderBy ?

rPraml:
I have one config file (spring application.properties) where all the config parameters are set. (ebean.*, spring.*, de.foconis.*)
I initialize the ServerConfig with config.loadFromProperties(applicationProperties)
So, if there is any property containing a ${} (in my case something like `de.foconis.customOrderByCollation=lcase(${})`) , PropertiesLoader.eval fails.
I would expect that PropertiesLoader.eval leaves the string untouched, if the property could not be resolved and no default value exists.

rbygrave:
We need a test case for this.

rbygrave:
Oh, in this case we are loading properties into ServerConfig that are not expected to be used by ServerConfig ... but instead used by the application - **all** the properties are going into ServerConfig regardless.

I see.

> maybe the ${} placeholder was not the best choice here

Right, because in the application properties sense `${}` doesn't make sense.

rPraml:
I have provided a test case. The main issue is, that properties that are not ebean related will prohibit ebean from start.
Note that you often have no control what strings are in the property files.

rbygrave:
Ok, I've pushed the change with a test.  We can close this PR (fixed in master).

PR #1699, opened 2019-05-08
Fix: equals() and hashCode() in ModifyAwareXXX

rPraml:
This adds equals (and correct) hashCode to ModifyAwareList/Set/Map, so that they are equals with an other collection, that contains the same elements


rbygrave:
Cool - thanks.

PR #1702, opened 2019-05-14
Fix: Migration when dbdefault is involved

rPraml:
This PR fixes issues in 
- MySql
- Sqlserver
- Hana

The first two did not alter the type when dbdefault was dropped
Hana did alter the type twice


rbygrave:
Yes, looks good - thanks !!

PR #1703, opened 2019-05-16
Provide test cases for transaction leak problem

rPraml:
As a first step to #1701 I try to provide test cases with minor code fixes.


src/main/java/io/ebeaninternal/api/SpiTransactionProxy.java
```diff
@@ -325,7 +325,7 @@ public Object getUserObject(String name) {
 
   @Override
   public void depth(int diff) {
-    transaction.depth();
+    transaction.depth(diff);
```
rPraml: This is probably a bug.

src/main/java/io/ebeaninternal/server/transaction/DefaultTransactionThreadLocal.java
```diff
@@ -8,12 +8,7 @@
  */
 public final class DefaultTransactionThreadLocal {
 
-  private static final ThreadLocal<TransactionMap> local = new ThreadLocal<TransactionMap>() {
-    @Override
-    protected synchronized TransactionMap initialValue() {
-      return new TransactionMap();
-    }
-  };
+  private static final ThreadLocal<TransactionMap> local = ThreadLocal.withInitial(TransactionMap::new);
```
rPraml: This creates no anonymous class, which could be a problem for leaks

src/test/java/org/tests/transaction/TestExecuteComplete.java
```diff
@@ -11,75 +11,89 @@
...
-        Ebean.save(customer);
-      });
-    } catch (DataIntegrityException e) {
+    assertThatThrownBy(
+        () -> DB.execute(TxScope.required().setBatch(PersistBatch.ALL),
+            () -> {
+              Customer customer = DB.getReference(Customer.class, 42424242L);
+              Order order = new Order();
+              order.setCustomer(customer);
+              DB.save(order);
```
rPraml: I changed the tests to use DB. `Ebean.save(customer)` was wrong in my oppinion.

src/test/java/org/tests/transaction/TestExecuteComplete.java
```diff
@@ -11,75 +11,89 @@
...
 public class TestExecuteComplete extends BaseTestCase {
 
+  @Rule public TestName name = new TestName();
+
+  @After
+  public void checkForLeak() {
+    if (DefaultTransactionThreadLocal.clear()) {
+      System.out.println(name.getMethodName() + " had stuck transactionmap.");
+    } else {
+      System.out.println(name.getMethodName() + " cleared everything.");
```
rPraml: Currently, all tests are leaking a threadlocal object:
Transactionmap, key='h2', value = State(where transaction is null)

The next step I want to do, is to change the `TransactionMap.Map<String, State>` to `TransactionMap.Map<String, SpiTransaction>`

src/main/java/io/ebeaninternal/server/transaction/DefaultTransactionThreadLocal.java
```diff
@@ -123,4 +118,13 @@ public static void end(String serverName) {
     }
   }
 
+  /**
+   * This is for testing purposes only. It will return TRUE, if a TransactionLocal has stuck.
+   */
+  public static boolean clear() {
```
rbygrave: I think this is a question of whether DefaultTransactionThreadLocal can be garbage collected vs whether it is empty. That is, on tomcat webapp reload the expectation is that the threads are stopped and webapp classloader should be garbage collected (but the DefaultTransactionThreadLocal isn't being GC'ed at the moment and that I believe that leads to the tomcat webapp classloader not being GC'ed due to the reference to the TransactionMap class).

rPraml:
On hold... @rbygrave I have to investigate this weekend, if memory leak is gone now. maybe you extract the fixes in TestExecuteComplete, because the existing test cases are wrong. (customer was saved instead of order)

PR #1704, opened 2019-05-16
#1701 (v2) - Replace TransactionMap with HashMap in DefaultTransactionThreadLocal

rbygrave:
Alternative to PR 1701.  This replaces TransactionMap with HashMap to remove classloader leak with DefaultTransactionThreadLocal


rPraml:
The problem is, you must not store strong references of objects 'outside' the web-app-classloader.
In a ThreadLocal, the ThreadLocal is used as *weak* key in a map maintained by thread.
So the threadlocal instance itself should not be the problem, if it is stored in an other thread.
But anonymous classes could be a problem (I'm not really sure about that, but better to avoid it)

It should be also no problem, if you store an empty hashmap in the ThreadLocal (as all instances are loaded by the system-classloader). So a threadLocal.remove() should be also not neccessary. I merged this in my branch where I fixed the tests. I have still to investigate, if this will avoid the leak, but I'm pretty sure, that your change go into the right direction

rbygrave:
The second commit here simplifies the code via:

- Using ThreadLocal.withInitial to always have a map.
- Remove the calls to local.remove(); when the map is empty (with the expectation that the thread local will GC with an empty map).

I think I'm happy with this. I'll look to merge it soon.

PR #1708, opened 2019-05-17
Refactor Tests: Check idType instead of assuming idType for a platform

rPraml:
This PR checks against the IdType and not against the platform.
It simplifies the code a bit, but should not change any functionality


rbygrave:
Great thanks - that is a nice tidy up that we want.  Ta.

PR #1709, opened 2019-05-17
Re-enabled some asserts (in relation #1298)

rPraml:
Here I re-enabled some asserts, that were disabled, because other tests had interefered this test before #1298 was done


rPraml:
It seems that there are still any differences between the ebean and foconis branch and this test is not yet ready to merge in.

PR #1711, opened 2019-05-17
Fix memory leak in threadLocal on implicit transactions

rPraml:
In continuation to #1704:
When an implicit transaction is used, the transaction will not be removed from the threadLocal.

So the JDBC (and JTA) transaction have to be removed from the ThreadLocal, when they are complete.

/edit: reverted the JTA part, as I cannot test it.

@rbygrave I tried to keep the changes as mininal now ;)
Can you take a look at this commit: https://github.com/ebean-orm/ebean/commit/c36e65e645cc56a17e3f5793e975ec10fe1a8229 - do you think it is good to check if there are leaked transactions after each test?

### DIFF
--- a/src/main/java/io/ebeaninternal/server/transaction/DefaultTransactionScopeManager.java
+++ b/src/main/java/io/ebeaninternal/server/transaction/DefaultTransactionScopeManager.java
@@ -42,5 +42,9 @@ public class DefaultTransactionScopeManager extends TransactionScopeManager {
     DefaultTransactionThreadLocal.set(serverName, trans);
   }
 
+  @Override
+  public void clear(SpiTransaction trans) {
+    DefaultTransactionThreadLocal.clear(serverName, trans);
+  }
 
 }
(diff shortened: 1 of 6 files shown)


rbygrave:
> When an implicit transaction is used, the transaction will not be removed from the threadLocal.

I'll look at that.

So yes this is a nice clean change and I'm happy with it.  I just think there is an underlying issue there with implicit transactions that I need to look at (ideally we fix the underlying issue).

> Can you take a look at this commit: c36e65e - do you think it is good to check if there are leaked transactions after each test?

Yes, that looks like a really good idea.  I'm happy for us to expose currentTransactions() for testing purposes.

rPraml:
> So yes this is a nice clean change and I'm happy with it. I just think there is an underlying issue there with implicit transactions that I need to look at (ideally we fix the underlying issue).

The underying issue is, that in the one case, a `ScopedTransaction` is in the ThreadLocal which will clear the TL properly, after the last transaction is popped from the stack.

When an implicit transaction is started, there is no scoped transaction in the TL, but the jdbc transaction itself.
On a txn.commit it just goes to 'inactive' state and relies in the TL
So, we must either
- always use scopedTransaction
- not putting non scoped transactions in the TL
- change the jdbc transaction that it will get removed on deactivate

rbygrave:
Note that I have a PR that adjusts this a little (I think more explicit with respect to clearing implicit inactive transactions from thread local scope).

PR #1712, opened 2019-05-17
The BaseTestCase ensures, that the threadLocal is cleared after each test.

rPraml:


rbygrave:
Yeah cool I like this, thanks.

PR #1714, opened 2019-05-17
Adjustment on #1711, more explicit clearing of inactive transactions from thread local

rbygrave:
Hi @rPraml , this is an adjustment that I think gives us more explicit clearing of the inactive implicit transactions.

The specific case is implicit persistence transactions (insert, update, delete etc) and note that these are always JdbcTransactions or more specifically NOT ScopedTransactions (as there is no possibility of nested transactions for this case).

### DIFF
--- a/src/main/java/io/ebeaninternal/server/transaction/JdbcTransaction.java
+++ b/src/main/java/io/ebeaninternal/server/transaction/JdbcTransaction.java
@@ -920,9 +920,6 @@ public class JdbcTransaction implements SpiTransaction, TxnProfileEventCodes {
     }
     connection = null;
     active = false;
-    if (manager != null) {
-      manager.scope().clear(this);
-    }
     profileEnd();
   }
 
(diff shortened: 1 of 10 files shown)


src/main/java/io/ebeaninternal/server/transaction/JdbcTransaction.java
```diff
@@ -920,9 +920,6 @@ protected void deactivate() {
     }
     connection = null;
     active = false;
-    if (manager != null) {
-      manager.scope().clear(this);
-    }
     profileEnd();
```
rPraml: yes, remove that.

rbygrave:
> This looks good for me.

Yeah, your adjustment to BaseTestCase also picked up issues with delete by id and update/delete queries leaving inactive transactions but now we are all good - good stuff.

> refactor ... end, rollback, commit and obtain are no longer used in ebean.

Yes. I'm going to look at that now.

> your intentions to use one threadlocal for all servers with a hashmap instead of a threadlocal per server.

I'm keen on that now, I think that would be good. I'll square away that refactor and look to get that into master now'ish.

PR #1715, opened 2019-05-18
Refactor tidy on DefaultTransactionThreadLocal

rbygrave:
- remove unused methods
- use clear() rather than set(null)


rbygrave:
Hi @rPraml , this is the refactor tidy on DefaultTransactionThreadLocal

PR #1716, opened 2019-05-18
Refactor DefaultTransactionThreadLocal, convert ThreadLocal to field (from map)

rbygrave:
With this we effectively remove DefaultTransactionThreadLocal.java moving the ThreadLocal into DefaultTransactionScopeManager.java


rbygrave:
Hi @rPraml  ..l this is the refactor we were talking about removing the map and making the ThreadLocal a field in DefaultTransactionScopeManager.

PR #1717, opened 2019-05-18
Query Cache will not always be invalidated

rPraml:
When you have 2 beans, with enabled query cache, cache invalidation will not always happen
(see provided test case)

Suggested fix: everytime when `changeSet.addClearQuery` is executed, it must also execute `changeSet.addInvalidate` so that the tableModState is updated.

A bit off topic I noticed during debugging: 
TableModState uses System.currentTimeMillis(), which has a resolution of 10ms on my machine, so it took a while until I had a proper failing test case, because I do a `DB.save` and immediately (in the next 10ms) a `DB.query`, so that I ended up with the situation that  `tableModTimestamp == queryCacheEntryTimeStamp`.
In this case, https://github.com/ebean-orm/ebean/blob/master/src/main/java/io/ebeaninternal/server/transaction/TableModState.java#L51 considers this entry as invalid.

This may make some test cases non deterministic, but should not make problems in production use **as long as no one turns back the system clock**
 
@rbygrave what would you suggest?
- leave it as it is?
- switch to System.nanoTime
- switch to an monotonic counting AtomicInteger/Long?

### DIFF
--- a/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
+++ b/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
@@ -220,6 +220,7 @@ final class BeanDescriptorCacheHelp<T> {
    */
   private void queryCacheClear(CacheChangeSet changeSet) {
     if (queryCache != null) {
+      changeSet.addInvalidate(desc);
       changeSet.addClearQuery(desc);
     }
   }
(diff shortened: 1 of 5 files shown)


rbygrave:
Yup, looks good - thanks.

>  as long as no one turns back the system clock

Well, I was happy before but I'll have another look.

rbygrave:
Actually both of these tests pass without the commit for `FIX: proper cache invalidation when dependent table will be modified`.

The first test on Address with `@InvalidateQueryCache` ... at BeanDescriptorCache line 877 (via persistUpdate() with invalidateQueryCache as true).

The second on PersistRequestBean line 1442 ... changeSet.addBeanUpdate(...).

So technically we can remove the change that is in the PR, the tests still pass.

rbygrave:
I have removed the changes this added into src/main ... as the tests still pass.

PR #1720, opened 2019-05-20
Refactor TableModState to use nanoTime refactor

rbygrave:


rPraml:
@rbygrave cool, now we can remove the Thread.sleeps from my last test

PR #1721, opened 2019-05-20
changed test to detect issue #1717

rPraml:
This is a followup to #1717, hope it is clear now, that there is a problem
(sorry, it was my mistake I ported the test wrong from our application)


rPraml:
Yeah, the test is failing now ;)
@rbygrave The reverted commit would fix the problem, maybe you can take a look if you have a better solution, maybe CacheChangeSet.addClearQuery can/should also touch the table

rbygrave:
> maybe CacheChangeSet.addClearQuery can/should also touch the table

Yes, I was thinking the same thing, put it internal to addClearQuery().

PR #1723, opened 2019-05-23
Testcase for ebean with circular dependences

tobias-:
Two failing test cases. 
1. While the javax.persistance api discourages the use of bidirectional
   cascades, it does not prohibit them. They're very useful and natural
   if you think of the database as a graph instead of a tree.
   test case 1 (testCircularCascade) tests this.
2. Second test case fetches the child, but saves the parent. If used like
   that, the child will not have it's changes saved, even with cascade = ALL.

### DIFF
--- a/src/test/java/org/tests/update/TestUpdateCircularSave.java
+++ b/src/test/java/org/tests/update/TestUpdateCircularSave.java
@@ -1,0 +1,75 @@
+package org.tests.update;
+
+import static org.junit.Assert.assertTrue;
+
+import io.ebean.BaseTestCase;
+import io.ebean.Ebean;
+import org.junit.Test;
+import org.tests.update.objects.SiblingA;
+import org.tests.update.objects.Parent;
+import org.tests.update.objects.SiblingB;
+import org.tests.update.objects.Child;
+
+public class TestUpdateCircularSave extends BaseTestCase {
+
+  @Test
+  public void testCircularCascade() {
+
+    long aId = createA();
+    modifyPropertyToTrue(aId);
+
+    SiblingA siblingA = Ebean.find(SiblingA.class, aId);
+    assert siblingA != null;
+    assertTrue(siblingA.getSiblingB().isTestProperty());
+  }
+
+  private void modifyPropertyToTrue(long aId) {
+    SiblingA siblingA = Ebean.find(SiblingA.class, aId);
+    assert siblingA != null;
+
+    final SiblingB siblingB = siblingA.getSiblingB();
+    siblingB.setTestProperty(true);
+    // Will get optimistic lock as version is increased twice on B
+    Ebean.save(siblingB);
+  }
+
+  private long createA() {
+    SiblingA siblingA = new SiblingA();
+    SiblingB siblingB = new SiblingB();
+    siblingA.setSiblingB(siblingB);
+
+    Ebean.save(siblingA);
+    return siblingA.getId();
+  }
+
+
+  @Test
+  public void testFetchChildSaveParent() {
+
+    long dId = createCAndD();
+    modifyPropertyToTrue2(dId);
+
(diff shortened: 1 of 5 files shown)


src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java
```diff
@@ -105,7 +105,9 @@ public T jsonRead(JsonParser parser) throws IOException {
       }
       default: {
         String jsonDateTime = parser.getText();
-        return convertFromTimestamp(dateTimeParser.parse(jsonDateTime));
+        Instant instant = Instant.parse(jsonDateTime.replace(" ", "T"));
+        Timestamp ts = Timestamp.from(instant);
+        return convertFromTimestamp(ts);
```
rPraml: I'm not happy with this change
1. why do you need to replace ` ` with `T`? (btw when replacing single characters it is better to use `replace(' ', 'T')`)
2. the change makes the dateTimeParser.parse method obsolete. So it would be better to change the 'parse' implementation.

What was the problem here? I assume, the precision of fractions of the DateTimeParser. Newer java versions (11+) probable needs up to 9 digits: `yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS'Z'
@rbygrave I think it would be a good idea to refactor the DateTimeParser, so that it uses Instant.parse instead of SimpleDateFormat

rbygrave:
Nice tests.  I should get to them shortly (tomorrow night).

src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java
rbygrave: > good idea to refactor the DateTimeParser, so that it uses Instant.parse instead of SimpleDateFormat

Yes, I agree @rPraml  @tobias-  ... Refactoring UtilDateTimeParser to use Instant.parse() looks like the good option.

tobias-: @rPraml @rbygrave Sorry, I managed to put extra stuff on this branch (it's been force pushed away not long after @rPraml commented on this PR). The java.time issue (#1725) is PR #1724 and nothing else. This PR is about test cases for circular dependencies.

tobias-:
@rbygrave I can make these test cases work by moving the `request.executeOrQueue()` to before `saveAssocOne()`. That makes a lot of other test cases fail though. Below is what I mean.

```diff
diff --git a/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java b/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
index 0d02a49ef..3d8134b7b 100644
--- a/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
+++ b/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
@@ -546,10 +546,6 @@ public final class DefaultPersister implements Persister {
     }
     request.flagUpdate();
     try {
-      if (request.isPersistCascade()) {
-        // save associated One beans recursively first
-        saveAssocOne(request);
-      }
 
       if (request.isDirty()) {
         request.executeOrQueue();
@@ -558,6 +554,11 @@ public final class DefaultPersister implements Persister {
         logger.debug(Message.msg("persist.update.skipped", request.getBean()));
       }
 
+      if (request.isPersistCascade()) {
+        // save associated One beans recursively first
+        saveAssocOne(request);
+      }
+
       if (request.isPersistCascade()) {
         // save all the beans in assocMany's after
         saveAssocMany(request);
```

rbygrave:
Note that there is not a fix for testFetchChildModifyChildSaveParent() yet.  This test just marked with `@Ignore` until I can look at it more.

PR #1724, opened 2019-05-27
Add (failing) java 11 to travis - date time parsing with nanos precision

tobias-:


tobias-:
Possible fix is 
```diff
diff --git a/src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java b/src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java
index 17f7c320b..d41958e6e 100644
--- a/src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java
+++ b/src/main/java/io/ebeaninternal/server/type/ScalarTypeBaseDateTime.java
@@ -12,6 +12,7 @@ import java.math.BigDecimal;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.Instant;
 
 /**
  * Base type for DateTime types.
@@ -105,7 +106,9 @@ public abstract class ScalarTypeBaseDateTime<T> extends ScalarTypeBase<T> {
       }
       default: {
         String jsonDateTime = parser.getText();
-        return convertFromTimestamp(dateTimeParser.parse(jsonDateTime));
+        Instant instant = Instant.parse(jsonDateTime.replace(" ", "T"));
+        Timestamp ts = Timestamp.from(instant);
+        return convertFromTimestamp(ts);
       }
     }
   }
```

tobias-:
As per what @rPraml said, yes, it should be fixed in a more permanent way than that just ignoring parts of the dateTimeParser and throwing `Instant.parse` on it. I've not checked that all the different patterns that dateTimeParser supports still work with the vanilla `Instant.parse`-parser. The whole reason for replacing " " with "T" is that a stackoverflow post suggested it.

I share the opinion that since ebean requires at least java 8, java.time.* should be used instead of the old java.text.* api's or even java.sql.Timestamp (although that, while missing time zone, has nanos), since they're more accurate and harder to shoot yourself in the foot with, especially with time offsets and timezones (which are weirdly difficult to get right).

Issue #1726, opened 2019-05-28
`createEntityBean` unloads initialized properties


jonasPoehler:
Also see: https://groups.google.com/forum/#!topic/ebean/D0RAXlni1aA

## Expected behavior

Calling `DB.getDefault().createEntityBean(Model.class)` is expected to behave analogously to calling `new`.

## Actual behavior

`createEntityBean(Model.class)` returns an instance of the model where properties initialized in the constructor have been "unloaded". When the bean then gets persisted, cascading references initialized in the constructor are not being persisted. The error only shows, when retrieving the bean from the database and checking the reference.

### Steps to reproduce

**Models**

```java
@Entity
class MainModel {

  @Id
  Integer id;

  @OneToOne(cascade = CascadeType.ALL, mappedBy = "main")
  ReferencedModel referencedModel = new ReferencedModel("reference from main");
}
```

```java
@Entity
class ReferencedModel {
  
  @Id
  Integer id;

  String name;

  @OneToOne(cascade = {})
  MainModel main;

  ReferencedModel (String name) {
    this.name = name;
  }
}
```

**Test**

```java
MainModel main = DB.getDefault().createEntityBean(MainModel.class);
Ebean.save(main);
main = Ebean.find(MainModel.class).findOne();
assertNotNull(main.getReferencedModel()); // <-- This fails
```

rPraml:
@jonasPoehler are you using our latest internal ebean version (11.38.2-FOC5) I've made this commit last week as you told me the problem:

https://github.com/FOCONIS/ebean/commit/270721bdb54373a83ef00b19a3c35da50f2a2bf5

I did not have time yet, to write test nor to create a PR for upstream. Maybe you can assist?

jonasPoehler:
> @jonasPoehler are you using our latest internal ebean version (11.38.2-FOC5)

I didn't test with -FOC-5 yet, but will do so as soon as I got time for it.
A soon as I can verify the problem fixed, I could create a test and PR.

rbygrave:
Fix merged, closing this one.

PR #1729, opened 2019-06-07
Bugfix/create entity bean unload

jonasPoehler:
This PR fixes the unexpected behaviour of `createEntityBean` described in issue #1726.


rbygrave:
Yes good - merging.

PR #1730, opened 2019-06-11
Add testcase for wrong batch sorting

jonasPoehler:
This PR adds a testcase for a problem that showed up after a fix in c66e954. This test shows that the order of statements in a batch can be wrong when first having a *ToMany reference and then from there two or more *ToOne reference, when using UUID as identity type.

### DIFF
--- a/src/test/java/org/tests/cascade/TestMultiCascadeBatch.java
+++ b/src/test/java/org/tests/cascade/TestMultiCascadeBatch.java
@@ -1,0 +1,55 @@
+package org.tests.cascade;
+
+import io.ebean.BaseTestCase;
+import io.ebean.Ebean;
+import io.ebean.Transaction;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.tests.model.site.DataContainer;
+import org.tests.model.site.Site;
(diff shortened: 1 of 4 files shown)


rbygrave:
> first having a *ToMany reference and then from there two or more *ToOne reference, when using UUID as identity type.

Yes, skipping the first *ToMany ... doesn't reproduce / works / presents us a workaround.  It looks like it's related to the *ToMany with both being the same type (both are of type Site).

Investigating ...

rbygrave:
This is a bit of a refactor that simplifies the ordering of
batch execution (BatchedBeanHolder ordering). In short we
batch by bean type and depth (rather than just type).

This means that there are some cases we are not optimal
(where we could batch by bean type in different depths)
but that is expected to be rare and problematic for
hierarchies of the same type (like this bug / test case)
hence I like the simplification that this fix brings in.

jonasPoehler:
Thank you very much for this fast fix! :+1: IMHO: This change really helps with understanding the behaviour!

rbygrave:
> IMHO: This change really helps with understanding the behaviour!

Yes, the prior logic was pretty ugly and complex - this code is happier :)

PR #1753, opened 2019-07-09
Issue with using the Formula-Properties (query generates wrong sql)

ThomasFOC:
Hello Rob,

we found an issue especially in `findIds()` and `findSingleAttributeList()`.

For example this query
```java
Ebean.find(Order.class)
        .select("totalItems")
        .where().eq("totalItems", 3)
        .findIds();
```
produces the following error:
```
javax.persistence.PersistenceException: Query threw SQLException:Feld "Z_BT0.TOTAL_ITEMS" nicht gefunden
Column "Z_BT0.TOTAL_ITEMS" not found; SQL statement:
select t0.id from o_order t0 where z_bt0.total_items = ? [42122-199] Bind values:[null] Query was:select t0.id from o_order t0 where z_bt0.total_items = ?
[...]
```
It seems that the join-clause of the Formula-Annotation of the `totalItems` is missing.

I've added some tests with Transient and Non-Transient formulas that should work in my opinion. it would be great if you can take a look at this.

cheers
Thomas

### DIFF
--- a/src/main/java/io/ebeaninternal/api/ManyWhereJoins.java
+++ b/src/main/java/io/ebeaninternal/api/ManyWhereJoins.java
@@ -7,7 +7,9 @@ import io.ebean.util.SplitName;
 import io.ebeaninternal.server.query.SqlJoinType;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.TreeMap;
 import java.util.TreeSet;
 
@@ -21,9 +23,7 @@ public class ManyWhereJoins implements Serializable {
 
   private final TreeMap<String, PropertyJoin> joins = new TreeMap<>();
 
-  private StringBuilder formulaProperties = new StringBuilder();
-
-  private boolean formulaWithJoin;
+  private List<String> formulaJoinProperties;
 
   private boolean aggregation;
 
@@ -125,27 +125,24 @@ public class ManyWhereJoins implements Serializable {
    * specifically for the findCount query.
    */
   public void addFormulaWithJoin(String propertyName) {
-    if (formulaWithJoin) {
-      formulaProperties.append(",");
-    } else {
-      formulaProperties = new StringBuilder();
-      formulaWithJoin = true;
+    if (formulaJoinProperties == null) {
+      formulaJoinProperties = new ArrayList<>();
     }
-    formulaProperties.append(propertyName);
+    formulaJoinProperties.add(propertyName);
   }
 
   /**
    * Return true if the query select includes a formula with join.
    */
   public boolean isFormulaWithJoin() {
-    return formulaWithJoin;
+    return formulaJoinProperties != null;
   }
 
   /**
    * Return the formula properties to build the select clause for a findCount query.
    */
-  public String getFormulaProperties() {
-    return formulaProperties.toString();
+  public List<String> getFormulaJoinProperties() {
+    return formulaJoinProperties;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


rPraml:
@rbygrave  I've added a possible fix, maybe you take a look at this fix, because I do not exactly know what happens in https://github.com/ebean-orm/ebean/blob/master/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java#L267

rbygrave:
Yes, happy with all that - well done.  I added some more asserts on the generated SQL and did minor tidy stuff.

> CQueryBuilder.java#L267

So the best explanation is that when for the findCount() cases we generate sql and we need those in the select because they come from the inline view like:
```sql
select count(*) from ( 
  select t0.id, z_bt0.total_items, z_bt0.total_amount      -- inline view columns here  
  from o_order t0 
  join (select order_id ....)
```
Now the `z_bt0.total_items` and `z_bt0.total_amount` are the formula properties we add to the select for this findCount() case.  We are doing that because we join to the inline view ... and use those columns from the inline view in the where clause.

For TestQueryJoinOnFormula.test_OrderFindCount_multiFormula:

```java
    int orders = Ebean.find(Order.class)
      .where()
      .eq("totalItems", 3)
      .gt("totalAmount", 10)
      .findCount();
```
We get the following SQL and we see the the where clause we use the columns from the inline view in the where clause. 
```sql
select count(*) from ( 
  select t0.id, z_bt0.total_items, z_bt0.total_amount      -- inline view columns
  from o_order t0 
  join (  select order_id, count(*) as total_items, 
                       sum(order_qty*unit_price) as total_amount 
           from o_order_detail 
           group by order_id
         ) z_bt0 on z_bt0.order_id = t0.id  
  where z_bt0.total_items = ?                  -- inline view columns used in where
  and z_bt0.total_amount > ?);                -- inline view columns used in where
--bind(3,10)
```

rPraml:
> Now the z_bt0.total_items and z_bt0.total_amount are the formula properties we add to the select for this findCount() case. We are doing that because we join to the inline view ... and use those columns from the inline view in the where clause.

Unfortunately, this triggers an issue on SqlServer (which is also in one of our application). See this test:

https://github.com/FOCONIS/ebean/blob/e56b63f65477a633908caa594077024728d7ae0f/src/test/java/org/tests/query/joins/TestQueryJoinOnFormula.java#L141
I get this exception
```
javax.persistence.PersistenceException: No column name was specified for column 2 of 'c'. 
Bind values:[3] 
Query was:
select count(*) from (
	select t0.identifier, coalesce(f2.child_age, 0) from parent_person t0 
		left join (
			select i2.parent_identifier, count(*) as child_count, sum(i2.age) as child_age
			from child_person i2 
			group by i2.parent_identifier
		) f2 on f2.parent_identifier = t0.identifier  
		where coalesce(f2.child_age, 0) = ?
	) as c
```

The reason is, that `coalesce(f2.child_age, 0)` has no column alias, and SqlServer is complaining about this.
(SqlServer17Platform has no columnAliasPrefix)
So, `query.setSelectId()` will solve the problem here for that particular test case.
I will do some investigations...

rPraml:
@rbygrave It seems that explicitly selecting the properties in the CQueryBuilder is not longer neccessary.
The changes in this PR (#1753) ensure, that the views are joined (which was prevoiously achived by adding the properties to the select, isn't?)

So I would suggest, always to select the ID now, as it fixes the SqlServer issue and breaks no other tests.

See this commit:
https://github.com/FOCONIS/ebean/commit/8370749ce9d5219ce7acdc18be102f7faf987608

rbygrave:
> explicitly selecting the properties in the CQueryBuilder is not longer neccessary.

Agreed. Pushed another commit that uses selectId().

PR #1754, opened 2019-07-12
When using @DbForeignKey(noconstraint = true) in views, you may not access the ID of a 'deleted' entity

rPraml:
When you explicitly select properties, you may not access to the origin id of a deleted reference

### DIFF
--- a/src/test/java/org/tests/lazyforeignkeys/ViewMainEntityRelation.java
+++ b/src/test/java/org/tests/lazyforeignkeys/ViewMainEntityRelation.java
@@ -1,0 +1,47 @@
+package org.tests.lazyforeignkeys;
+
+import java.util.UUID;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+
+import io.ebean.annotation.DbForeignKey;
+import io.ebean.annotation.View;
+
+@View(name = "vw_main_entity_relation")
+@Entity
+public class ViewMainEntityRelation {
+
+  @Id
+  private UUID id;
+
+  @ManyToOne
+  @JoinColumn(name = "id1")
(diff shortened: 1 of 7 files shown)


src/test/java/org/tests/lazyforeignkeys/ViewMainEntityRelation.java
```diff
@@ -0,0 +1,47 @@
...
+@View(name = "vw_main_entity_relation")
+@Entity
+public class ViewMainEntityRelation {
+
+  @Id
+  private UUID id;
+
+  @ManyToOne
+  @JoinColumn(name = "id1")
+  @DbForeignKey(noConstraint = true)
```
rbygrave: This doesn't make sense to me in that this is a `@View` entity.  Why put a `@DbForeignKey` annotation on an entity that is a `@View` ?

rbygrave:
Why are we using `@DbForeignKey(noconstraint = true)` in views in the first place?

src/test/java/org/tests/lazyforeignkeys/ViewMainEntityRelation.java
rPraml: The issue is reported by Alex and I think, this is not really view related and the main reason for doing this, is to enable the "softdelete" mechanism from #1527 as it is done in the `org.tests.model.nofk.*` test cases

@AlexWagner is this correct?

rbygrave: I can't merge this in based on what I think is an incorrect test - using `@DbForeignKey` annotation on a view isn't valid.

As I see it this PR needs to re-done.

AlexWagner: PR no longer uses a view

rbygrave:
Great, thanks @AlexWagner  and @rPraml

PR #1755, opened 2019-07-14
Cannot use enhanced classes, when they are in different maven packages with agent 11.40.1

rPraml:
We have an `AbstractBaseModel` in maven package A and `RegistryEntry` in package B

So both packages are compiled and enhanced in separate steps.

There were changes in the enhancer, so compilation of pkg B fails:
```
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_setFieldIntercept(int, Object)
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_setField(int, Object)
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getPropertyName(int)
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_newInstance()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getFieldIntercept(int)
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getField(int)
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_setEmbeddedLoaded()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getMarker()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_intercept()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_isEmbeddedNewOrDirty()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getPropertyNames()
The type RegistryEntry must implement the inherited abstract method EntityBean._ebean_getIntercept()
```

I assume that this change is the root cause: https://github.com/ebean-orm/ebean-agent/pull/113
AbstractBaseModel is enhanced, in pkg A, but the methods in `AbstractBaseModel implements EntityBean` are no longer visible for the compiler, so when compiling RegistryEntry in pkg B, it enforces the implementation of all methods in `EntityBean`

This workaround/fix has a default implementation for each method in EntityBean.

I'm not sure if this is an ebean or ebean-agent issue

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBean.java
+++ b/src/main/java/io/ebean/bean/EntityBean.java
@@ -16,12 +16,16 @@ public interface EntityBean extends Serializable {
   /**
    * Return all the property names in defined order.
    */
-  String[] _ebean_getPropertyNames();
+  default String[] _ebean_getPropertyNames() {
+    throw new NotEnhancedException();
+  }
 
   /**
    * Return the property name at the given position.
    */
-  String _ebean_getPropertyName(int pos);
+  default String _ebean_getPropertyName(int pos) {
+    throw new NotEnhancedException();
+  }
 
   /**
    * Return the enhancement marker value.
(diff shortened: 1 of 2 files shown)


rbygrave:
Which `javac` is this?

rPraml:
We use the eclipse compiler. It happens also in eclipse, if I add a precompiled and enhanced class to the eclipse build path. I didn't try it with javac.

PR #1756, opened 2019-07-15
backdoor to disable @Transactional during unit testing

ajcamilo:
```
HelpScopeTrans.setEnabled(false)
```

disables the creation of transactions.

I'm open to changes with the way to do this.

### DIFF
--- a/src/main/java/io/ebeaninternal/api/HelpScopeTrans.java
+++ b/src/main/java/io/ebeaninternal/api/HelpScopeTrans.java
@@ -7,22 +7,37 @@ import io.ebean.TxScope;
  * Helper object to make AOP generated code simpler.
  */
 public class HelpScopeTrans {
+  private static boolean enabled = true;
 
   /**
    * Entering an enhanced transactional method.
    */
   public static void enter(TxScope txScope) {
-    server().scopedTransactionEnter(txScope);
+    if (enabled) {
+      server().scopedTransactionEnter(txScope);
+    }
   }
 
   /**
    * Exiting an enhanced transactional method.
    */
   public static void exit(Object returnOrThrowable, int opCode) {
-    server().scopedTransactionExit(returnOrThrowable, opCode);
+    if (enabled) {
+      server().scopedTransactionExit(returnOrThrowable, opCode);
+    }
   }
 
   private static SpiEbeanServer server() {
     return (SpiEbeanServer) Ebean.getDefaultServer();
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


rbygrave:
I'll ponder some more but no I can't think of a better approach.

Note that we ought to expose this API via something somewhere under `io.ebean` package and not `io.ebeaninternal` package.  So once we are happy with the implementation here we should add that.

rbygrave:
Adding io.ebean.ForTests such that:

```java
public class ForTestsTest {

  @Test
  public void noTransactional_expect_noTransactionEnterExitCalled() {

    ForTests.noTransactional(this::checkForTransactional);
  }

  /**
   * Simple use BUT need to remember to call ForTests.enableTransactional(true); at the end!!
   */
  @Test
  public void enableTransactional() {

    ForTests.enableTransactional(false);

    checkForTransactional();

    ForTests.enableTransactional(true);
  }

  /**
   * Transactional method but no transaction start/exit is actually called 
   * as this is wrapped by ForTests.noTransactional()
   */
  @Transactional
  private void checkForTransactional() {

    final Transaction transaction = DB.currentTransaction();
    assertThat(transaction).isNull();
  }
}

```

ForTests:
```
package io.ebean;

import io.ebeaninternal.api.HelpScopeTrans;

/**
 * Helper methods for testing.
 */
public class ForTests {

  /**
   * Enable or disable <code>@Transactional</code> methods.
   * <p>
   * This is intended for testing purposes such that tests
   * on code with {@code @Transactional} methods don't actually
   * start or complete transactions.
   * </p>
   *
   * @param enable Set false to disable {@code @Transactional} methods
   */
  public static void enableTransactional(boolean enable) {
    HelpScopeTrans.setEnabled(enable);
  }

  /**
   * Run the closure with <code>@Transactional</code> methods
   * effectively disabled (they won't create/commit transactions).
   */
  public static void noTransactional(Runnable run) {
    try {
      enableTransactional(false);
      run.run();
    } finally {
      enableTransactional(true);
    }
  }
}
```

rPraml:
```

  /**
   * Simple use BUT need to remember to call ForTests.enableTransactional(true); at the end!!
   */
  @Test
  public void enableTransactional() {

    ForTests.enableTransactional(false);

    checkForTransactional();

    ForTests.enableTransactional(true);
  ```

Please add a try... finally statement in the documentation. (to avoid copying bad practice code) 
If the test fails, transactional stays disabed for other tests, too.

PR #1800, opened 2019-08-23
DRawSqlService might silently drop columns with duplicate names

focdanisch:
If the map already contains the column name, add schema and table names before the key to make it unique again


rbygrave:
Yes, I'm happy with this.  Nice work, merging ...

PR #1812, opened 2019-09-10
Allow for null items in @DbArray lists.

Ryszard-Trojnacki:
Current version of Ebean allows to store null items in @DbArray lists, but throws NPE while loading them.
Null items are valid for Json serialization/deserialization and are also valid for native Postgres arrays.

This pull requests removes NPE when loading @DbArray with null values from database.

I have also added tests for this case of use.


rbygrave:
Good work - yes, happy with this - merging.

PR #1840, opened 2019-10-10
exists() limits the max rows on the query, this should be done on a copy

focdanisch:
Hi Rob,

When I began to replace all the 'query.findCount() > 1' code with 'query.exists()', I realized that exists() modifies the query (sets the max rows size to 1). This should be done on a copy of the query so that the original query can be re-used after calling exists().

I simply changed the code so that exists() calls copy() before query modifications.

Best regards,
Manuel


focdanisch:
OK, the tests pass now. With my change, it is not possible any more to call getGeneratedSql() on the query, because a copy has been executed (same behaviour as findCount()).

PR #1859, opened 2019-11-10
Refactor rename method ProfileLocation.shortDescription() to location()

rbygrave:
location is strictly speaking the shortened version of the full location but it is the most useful and currently we don't use full location.

Renaming to `location()` makes sense.


rbygrave:
An API change for people reading the ProfileLocation part of the metrics - forgive me, bringing it into 12.1.5

Issue #1902, opened 2020-01-07
SingleAttributeQueries with child selection only does not work with autoTune=false (e.g. Views)


rPraml:
## Expected behavior
we have a query with `.select("").fetch("child", "attribute").setCountDistinct(true).findSingleAttributeList()` on a view.

## Actual behavior
the query produces wrong result on a view, but works on a table.
The problem was, that all properties of the root-bean were selected.

Possible fix:
https://github.com/FOCONIS/ebean/commit/849a768a62aeb327c16f94151f363fb6e1b1c6b2

## Test
sorry, no test yet - setting `.setAutoTune(false)` does not trigger the bug.
would require to build a test like `findSingleAttributeList_with_join_column` on a view.

Cheers
Roland

Cheers
Roland

rbygrave:
The possible fix looks good to me.

rPraml:
not yet fixed, see #1908

PR #1908, opened 2020-01-10
Testcase for #1902

rPraml:
Hello Rob,

I think we need to call `setDefaultSelectClause()` or an other fix.

Take a look at this test case. 
```
    Query<Customer> query = Ebean.find(Customer.class)
      .fetch("billingAddress", "city");
```
vs
```
    Query<VwCustomer> query = Ebean.find(VwCustomer.class)
      .fetch("billingAddress", "city");
```
The `VwCustomer` is a copy of `Customer`, but uses `@View` instead of `@Table` annotation.
(i use no real view here, but this should not matter)

the assertion `assertThat(cities).contains("Auckland").containsNull();` fails with:
```
java.lang.AssertionError: 
Expecting:
 <[NEW, NEW, ACTIVE, ACTIVE]>
to contain:
 <["Auckland"]>
but could not find:
 <["Auckland"]>
```

you see, that the produced query has all root properties selected:
```
java.lang.AssertionError: 
Expecting:
 <"select t0.status, t0.name, t0.smallnote, t0.anniversary, t0.cretime, t0.updtime, t0.version, t0.shipping_address_id, t1.city from o_customer t0 left join o_address t1 on t1.id = t0.billing_address_id ">
to contain:
 <"select t1.city from o_customer t0 left join o_address t1 on t1.id = t0.billing_address_id"> 
	at org.tests.query.other.TestQuerySingleAttribute.findSingleWithFetchOnView(TestQuerySingleAttribute.java:302)
```

### DIFF
--- a/src/test/java/org/tests/model/basic/VwCustomer.java
+++ b/src/test/java/org/tests/model/basic/VwCustomer.java
@@ -1,0 +1,228 @@
+package org.tests.model.basic;
+
+import io.ebean.annotation.Cache;
+import io.ebean.annotation.ChangeLog;
+import io.ebean.annotation.ChangeLogInsertMode;
+import io.ebean.annotation.DbComment;
+import io.ebean.annotation.DbEnumValue;
+import io.ebean.annotation.DocEmbedded;
+import io.ebean.annotation.DocStore;
+import io.ebean.annotation.JsonIgnore;
+import io.ebean.annotation.View;
+import io.ebean.annotation.Where;
+import org.tests.model.basic.finder.CustomerFinder;
+
+import javax.persistence.CascadeType;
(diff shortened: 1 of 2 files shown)


rbygrave:
> I think we need to call setDefaultSelectClause() or an other fix.

Ah ok.  The VwCustomer bean type is not autoTunable (because it is a view and not an entity).

rbygrave:
Right. The other fix option is to treat VIEW based entity types as autoTunable.  

I think this is the appropriate fix.  Conceptually there shouldn't be much difference between an entity based on a view and an entity based on a table - almost no difference (when considering that some views are updatable etc).  

So yeah, applying the fix where VIEW based entity types should be autoTunable true.

PR #1921, opened 2020-01-29
Split Postgres platform into Postgres10 & Postgres9 (DDL - generated by default as identity)

rbygrave:
- Use "generated by default as identity" for Postgres10 identity columns


rbygrave:
Also refer: https://www.2ndquadrant.com/en/blog/postgresql-10-identity-columns/

PR #1949, opened 2020-02-20
Test usingDatabase with a non default db

sebastianmrozek:
Move the test to separate class with a better name.
I did not think the test was sufficiently testing the purpose of this method, which I assume was provided to be able to use a different, non-default DB, when querying.
Commenting out the 
```
this.server = (SpiEbeanServer) database; 
```
assignment in io.ebeaninternal.server.querydefn.DefaultOrmQuery#usingDatabase did not make the original test fail.


rbygrave:
Good stuff, merging - thanks.

PR #1978, opened 2020-03-23
"insert into table default values" is invalid MySQL syntax

jonasPoehler:
Hi Rob,

while writing a testcase for another bugfix (coming up next) I found that MySQL has a problem when only inserting default values into a table. This might be kind of an edge case, because there are not many ways of letting a column fallback to the db default, but I fixed it anyway. 
Can you have look, if I did this right?

Best regards
Jonas


rbygrave:
Yup, looks great.  Well done.

PR #1981, opened 2020-03-25
Use temp variable to avoid NPE race

### LINKED ISSUE #1980: NullPointerException in EntityBeanIntercept
## Expected behavior
Clean execution.

## Actual behavior
Works more than 99.9% of the times, *rarely* a stacktrace with NPE. Approx once every >1'000 execution fails.
I don't think I've seen it more than one time for any unique test case. We have not upgraded production yet, so this is only in tests.

### Steps to reproduce

Not sure. I *suspect* a race condition, but we are completely unable to reproduce it reliably.

We do NOT have any pre-getter interceptors registered, but one IS attached (internal ebean hook).

Could it be because we do multiple saves on the same object instance?
Another possible reason it happens for us is that in tests we do 
`serverConfig.backgroundExecutorSchedulePoolSize = 1`, because we want to make sure all background tasks have been finished before moving on to the next test.

The code that blows up is trivial:
```kotlin
        if (bean.deleted) {
```

```console
Stacktrace
java.lang.NullPointerException
	at io.ebean.bean.EntityBeanIntercept.preGetterCallback(EntityBeanIntercept.java:917)
	at io.ebean.bean.EntityBeanIntercept.preGetter(EntityBeanIntercept.java:932)
	at OurDBO._ebean_get_deleted(OurDBO.kt:999)
	at OurDBO.getDeleted(OurDBO.kt:999)
	at OurBeanPersistListener.inserted(OurBeanPersistListener.kt:999)
	at io.ebeaninternal.server.core.PersistRequestBean.notifyLocalPersistListener(PersistRequestBean.java:565)
	at io.ebeaninternal.server.transaction.PostCommitProcessing.localPersistListenersNotify(PostCommitProcessing.java:160)
	at io.ebeaninternal.server.transaction.PostCommitProcessing.lambda$backgroundNotify$0(PostCommitProcessing.java:134)
	at io.ebeaninternal.server.core.DefaultBackgroundExecutor.lambda$execute$0(DefaultBackgroundExecutor.java:43)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)```

tobias-:
By using a temporary variable, avoid NullPointerException.

Resolves #1980

### DIFF
--- a/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -913,6 +913,7 @@ public final class EntityBeanIntercept implements Serializable {
   }
 
   private void preGetterCallback(int propertyIndex) {
+    PreGetterCallback preGetterCallback = this.preGetterCallback;
     if (preGetterCallback != null) {
       preGetterCallback.preGetterTrigger(propertyIndex);
     }


tobias-:
Crud. I'll fix the indentation

PR #1983, opened 2020-04-07
Add test for aggregating formula field with join

catatafishen:


rob-bygrave:
This PR is superseded by #3796, which implements the same feature (aggregation functions like `SUM()` over `@Formula` / `@Formula2` properties) in a modernised form:

- Updated to Jakarta Persistence (`javax` → `jakarta`)
- JUnit 4 → JUnit 5 with `@BeforeAll` / `@AfterAll` cleanup
- Uses `@Formula2` for the join-based `totalPrice` property instead of the raw `@Formula(join=...)`
- Fixes the underlying bug where `SUM(@Formula2property)` failed to auto-join the referenced table (`DynamicPropertyAggregationFormula` now exposes the required join paths via `formula2Joins()`)

Closing in favour of #3796 — thank you for the original test case that identified this gap!

PR #1986, opened 2020-04-09
@OrderColumn is ignored on relations referencing a abstract inheritance-model

jonasPoehler:
Hello Rob,

This PR shows a problem where the order-column is simply not inserted. The problem was, that the child-descriptor of the implementinig bean did not know of the order-column that was defined on a reference to the parent.
The fix is something probably up for debate. I am not entirly certain, that this is the right spot to fix this. I'm not sure, if this is not possibly too late and the DeployBeanDescriptor of the child was already read, but you should be able to assess this situation.

Best regards
Jonas


rbygrave:
> if this is not possibly too late and the DeployBeanDescriptor of the child was already read

I think this is ok.

PR #1988, opened 2020-04-09
Removing elements with @OrderColumn with a tree depth > 1 doesn't work

jonasPoehler:
This PR fixes a problem that from the depth > 1 a element is not being deleted from the database when removed from the collection. The other elements get an updated sort order, but the element is not deleted.
During trying to fix this, I also wrote another test for reordering elements on that same level, because my first attempt of fixing this effectively broke that case.
This also depends on this PR: https://github.com/ebean-orm/ebean/pull/1985, which fixes the insert of elements on that level.
The effective diff of this PR is this: https://github.com/FOCONIS/ebean/compare/bugfix/order_column_tree_depth...FOCONIS:bugfix/order_column_remove_and_reorder (sorry for the mess)

### DIFF
--- a/src/main/java/io/ebean/common/ModifyHolder.java
+++ b/src/main/java/io/ebean/common/ModifyHolder.java
@@ -1,5 +1,7 @@
 package io.ebean.common;
 
+import io.ebean.bean.EntityBean;
+
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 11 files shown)


src/main/java/io/ebean/common/ModifyHolder.java
```diff
@@ -54,6 +56,11 @@ private boolean undoDeletion(E bean) {
   void modifyAddition(E bean) {
     if (bean != null) {
       touched = true;
+
+      if (bean instanceof EntityBean) {
```
jonasPoehler, line 60: This is the place, where the question came up, why BeanCollection (and ModifyHolder) doesn't know anything about EntityBean: https://groups.google.com/forum/#!topic/ebean/wsiYTGYHpCg

PR #2041, opened 2020-08-04
Add STI Discrimination on Reference Associations

peterko2017:
This adds a discriminator predicate when joining against single-table inheritance associations.  

Without adding a discriminator on the join, certain use-cases where an Entity had multiple associations against entities using the same base entity for STI would not properly return the correct result set.


rbygrave:
Very nice, well done !!

PR #2061, opened 2020-09-23
Add failing testcase for wrong values retrieved from bean caching

jonasPoehler:
Hello Rob,

I seem to have stumbled upon an error in bean caching, which I don't really know how to fix properly.
I documented the problem in the testcase, but the problem seems to be, that the bean cached in the PersistenceContext is handed out on the first call to findOne() and if modified, it changes the value present in the cached object. A proposed fix would be to always return a new model (with the content present in the PC), but I am unsure where exactly to change this.
Could you assist with fixing this problem?

Thank you and best regards
Jonas

### DIFF
--- a/src/test/java/org/tests/cache/TestBeanCacheContactLazyLoad.java
+++ b/src/test/java/org/tests/cache/TestBeanCacheContactLazyLoad.java
@@ -1,0 +1,68 @@
+package org.tests.cache;
+
+import io.ebean.BaseTestCase;
+import io.ebean.DB;
+import io.ebeantest.LoggedSql;
+import org.junit.Test;
+import org.tests.model.basic.Contact;
+import org.tests.model.basic.Customer;
+
+import java.sql.Date;
+import java.time.LocalDate;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test class testing a wrong behaviour of the bean cache.
+ */
+public class TestBeanCacheContactLazyLoad extends BaseTestCase {
+
+  /**
+   * This test shows a wrong behaviour of the bean cache up to at least Ebean 12.4.*:
+   * <ul>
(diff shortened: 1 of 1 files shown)


rbygrave:
Ok, some terminology issues first:
- PersistenceContext (aka "L1 cache") ... I don't like calling it a cache because its PRIMARY role is to build consistent object graphs (de-duplicate instances)
- L2 Cache (a bunch of caches of beans, and query results).

Note that the L2 Bean cache doesn't hold bean instances (generally) (just like hibernates l2 cache) - refer to _CachedBeanData_ (CachedBeanDataFromBean, CachedBeanDataToBean). A hit on the L2 bean cache creates new instances.

But yes this problem is with the L2 bean cache (and not with the PersistenceContext).

----
- A Partial bean is loaded (properties email, lastname)   ... _OK_
- We mutate one of the loaded properties ... _OK_
- Lazy loading is invoked via contact.getPhone()  ... _OK_
- Ebean loads the L2 Bean cache after lazy loading ... _**PROBLEM**_

Generally it's ok for Ebean to load the L2 Bean cache after lazy loading. The issue though is that Ebean is not using all the "clean" data from the database to load the cache but instead includes the dirty property (mutated lastname).  This is the problem.

**Potential options:**
- Ebean doesn't load the L2 bean cache at all if lazy loading on a partial
- Ebean loads the "clean" data into L2 bean cache by (using old values of mutated properties) OR omits those properties altogether

rbygrave:
Note that the fix uses the second option.  When loading the bean into CacheBeanData, when the property is dirty/mutated use the original value.

jonasPoehler:
Wow, that was quick! Thank you for directly fixing this and for the detailed explanation on the topic. It shows that I was not looking in the correct spot, when debugging this problem. Your comment really clears things up for me 👍

PR #2082, opened 2020-10-15
Scalar type querybeans generics error

tobias-:
After my ScalarType change to querybeans, any ScalarType field with generics will strip the generics information. This will cause warnings in Java and compilation errors in Kotlin.

This PR includes a Kotlin test that tests it in addition to fixes it.


kotlin-querybean-generator/pom.xml
```diff
@@ -17,23 +22,126 @@
       <version>3.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.ebean</groupId>
+      <artifactId>ebean-querybean</artifactId>
+      <version>${project.version}</version>
```
rbygrave, line 28: I think this should be test scope.

PR #2112, opened 2020-11-24
2110 - Fix and test for executeBatch() on SqlUpdate

rbygrave:


ebean-core/src/test/java/org/tests/update/TestSqlUpdateBatch.java
```diff
@@ -0,0 +1,67 @@
...
+    try (Transaction txn = DB.beginTransaction()) {
+      txn.setBatchSize(100); // something bigger than 40
+      // Dummy update, that effectively does nothing, but ebean doesn't need to know this.
+      final SqlUpdate update = DB.sqlUpdate("update uuone set name = ? where 0=1");
+      for (int i = 0; i < 40; i++) {
+        update
+          .setParameter(1, String.valueOf(i))
+          .addBatch();
+      }
+      assertEquals(40, update.executeBatch().length);
```
rPraml, line 63: Hello Rob, our intention was, that the returned array of `update.executeBatch()` always matches the amount of `addBatch()` calls and does not depend on batch size settings.
 In most cases, we are not really interested in the row count, so it would be OK if `executeBatch` returns `void`
What do you think of a callback

```java
interface BatchCallback {
     rowsUpdated(int[] rowCounts)
}

final SqlUpdate update = DB.sqlUpdate("update uuone set name = ? where 0=1");
List<Integer> allCounts = new ArrayList;
update.setBatchExecutedCallback(rowCounts => allCounts.addAll(Arrays.asList(rowCounts)));
// do X addBatches
// allCounts should always contain X elements

```

rbygrave: >  update.executeBatch() always matches the amount of addBatch() calls and does not depend on batch size settings.

Well yes and I think that would be nice.  The issue with making that change is that the other bug/issue is totally dominated my "brain space" in the sense it is the bigger / tougher issue that I wanted to sort out first.  

We can look at this `executeBatch()` `addBatch` properly later like maybe in a week but in my mind that means it the suggested alternative is that `executeBatch()` does actually execute the batch (and that the batch isn't auto flushed earlier).  

> BatchCallback

My gut says no on that and says ... lets make executeBatch() actually execute the batch and stop the early flush that occurs (get it to ignore transaction batch size).

rbygrave: So approximately ... addBatch() needs to explicitly go deeper into the internals such that it "stops automatic batch execute on transaction batch size" from happening (and then see if that change is appropriate / ok).

PR #2113, opened 2020-11-24
Add support for Postgres lock types (no key, share, key share) with FOR UPDATE

rbygrave:
Gives us the ability to explicitly specify the "lock type" (strength) to use with Postgres `FOR UPDATE` clause including:

- `FOR UPDATE`
- `FOR NO KEY UPDATE`
- `FOR SHARE`
- `FOR KEY SHARE`

### Example

```java
Article article = DB.find(Article.class)
.setId(id)
.forUpdate(Query.LockType.NoKeyUpdate)
.findOne();

```

### DIFF
--- a/ebean-api/src/main/java/io/ebean/ExpressionList.java
+++ b/ebean-api/src/main/java/io/ebean/ExpressionList.java
@@ -175,6 +175,11 @@ public interface ExpressionList<T> {
   Query<T> forUpdate();
 
   /**
+   * Execute using "for update" with given lock type (currently Postgres only).
+   */
+  Query<T> forUpdate(Query.LockType lockType);
+
+  /**
    * Execute using "for update" clause with No Wait option.
    * <p>
    * This is typically a Postgres and Oracle only option at this stage.
@@ -183,12 +188,22 @@ public interface ExpressionList<T> {
   Query<T> forUpdateNoWait();
 
   /**
+   * Execute using "for update nowait" with given lock type (currently Postgres only).
+   */
+  Query<T> forUpdateNoWait(Query.LockType lockType);
+
+  /**
    * Execute using "for update" clause with Skip Locked option.
    * <p>
    * This is typically a Postgres and Oracle only option at this stage.
    * </p>
    */
   Query<T> forUpdateSkipLocked();
+
+  /**
+   * Execute using "for update skip locked" with given lock type (currently Postgres only).
+   */
+  Query<T> forUpdateSkipLocked(Query.LockType lockType);
 
   /**
    * Execute the query including soft deleted rows.
--- a/ebean-api/src/main/java/io/ebean/Query.java
+++ b/ebean-api/src/main/java/io/ebean/Query.java
@@ -180,6 +180,36 @@ import java.util.stream.Stream;
 public interface Query<T> {
 
   /**
+   * The lock type (strength) to use with query FOR UPDATE row locking.
+   */
+  enum LockType {
+    /**
+     * The default lock type - See PlatformConfig.lockWithKey option.
+     */
+    Default,
+
+    /**
+     * FOR UPDATE.
+     */
+    Update,
+
+    /**
+     * FOR NO KEY UPDATE.
+     */
+    NoKeyUpdate,
+
+    /**
+     * FOR SHARE.
+     */
+    Share,
+
+    /**
+     * FOR KEY SHARE.
+     */
+    KeyShare
+  }
+
+  /**
    * For update mode.
    */
   enum ForUpdate {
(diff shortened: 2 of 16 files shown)


arsentiikaptsan:
Awesome, thanks a lot!

jnehlmeier:
Just looked at it and thought that `.forUpdate(UPDATE)` looks weird and repetitive. Given that we want to configure the lock clause of the query, maybe rename it to `.withLock / .withLockClause(LockStrength.FOR_UPDATE)`? 

Possibly extending to `.withLock / .withLockClause(LockStrength.FOR_UPDATE, LockWaitBehavior.[WAIT|NO_WAIT|SKIP_LOCKED])` with `LockWaitBehavior.WAIT` just meaning to not append `NOWAIT | SKIP LOCKED` to the query.

rbygrave:
Hmmm maybe.

For all non-Postgres `forUpdate()` `forUpdateNoWait()` and `forUpdateSkipLocked()` are commonly supported and explicitly match the SQL so that does seem relatively nice and expected and why we have these methods like this (to match the SQL).

The "lock strength" is currently a Postgres only function and may well stay that for quite some time (this isn't well supported by other databases at all) so the thinking is that it is the lock strength that is the overloaded option of the "FOR UPDATE" clause.  So I'm only expecting Postgres folks to be using these explicit lock strength option.

These are all "FOR UPDATE" lock strength options (row locks). They are not another other type of locks and only apply to "FOR UPDATE".

Hence the thinking is we continue to have our existing `forUpdate()` `forUpdateNoWait()` and `forUpdateSkipLocked()` with wide database support ... and for PG folks add `forUpdate(LockType type)`, `forUpdateNoWait(LockType type)`, `forUpdateSkipLocked(LockType type)`



- `forUpdate()`   -- "FOR UPDATE" using the default lock strength. Well supported by almost all databases

- `forUpdate(NoKeyUpdate)`   -- Postgres only, explicit lock strength (maybe same or different to default PG lock strength)
- `forUpdate(Update)`    -- Postgres only, explicit lock strength (maybe same or different to default PG lock strength)

- `forUpdate(Share)`   -- Postgres only
- `forUpdate(KeyShare)`  -- Postgres only

Today, no one has reason to use `forUpdate(Update)` because it's Postgres only and matches the default lock strength.  If we change the default strength (or people change it via PlatformConfig.forUpdateNoKey) then it becomes useful.


> .forUpdate(UPDATE)

This is conceptually "FOR _KEY_ UPDATE" (as opposed to FOR _NO KEY_ UPDATE) but timing wise the NO KEY options were developed later for Postgres.

We could have called these KeyUpdate, NoKeyUpdate, KeyShare, NoKeyShare ... but then it doesn't match Postgres documentation so I thought it was better to match Postgres here.


Hmmm.

jnehlmeier:
But given that `Update` itself is already a lock strength, the method that takes a lock strength should not contain an existing lock strength in its name. So ideally it would be 

- `forUpdate()` -- the common "FOR UPDATE" SQL.
- `for(lockStrength = {UPDATE | NO KEY UPDATE | SHARE | KEY SHARE})` -- with `forUpdate` delegating to `for(UPDATE)`

However we can not use `for` as method name, thus we need a different method name in this case. That's why I thought about something like `withLock(lockStrength)` as method name. Something like `for_`, `for$`, `forLock` might also be fine.

Once the method names for both are settled one can duplicate the methods for the `no wait` and `skip locked` case.

rbygrave:
Still pondering the `.withLock / .withLockClause(LockStrength.FOR_UPDATE, LockWait.WAIT)` option.

However, given the existing `forUpdate()` methods I'm not unhappy with adding the `forUpdate(LockType type)` etc and if we decide to change this to something like `withLock()` then we need to deprecate the existing methods so ... looking to release with what we have here and review later.

In my thinking not many people are going to be explicitly using the explicit LockType which is maybe why I'm happy with what we have here even if `forUpdate(Query.LockType.UPDATE)` looks odd.

rbygrave:
Oh darn, I didn't actually see/read your comment until just now (I wrote my prior comment without having read your comment there).  

Now your point is sinking in and I think I get it now.  Preparing a PR with:

```java
  /**
   * Execute the query with the given lock type and WAIT.
   */
  Query<T> withLock(Query.LockType lockType);

  /**
   * Execute the query with the given lock type and lock wait.
   */
  Query<T> withLock(Query.LockType lockType, Query.LockWait lockWait);

```

In my mind I'm thinking of keeping the existing `forUpdate()`, `forUpdateNoWait()`, `forUpdateSkipLocked()`.  So we would end up with overlap in terms of 2 ways of doing the same thing, but I think that is ok.

rbygrave:
PR adding 2 `withLock()` methods - https://github.com/ebean-orm/ebean/pull/2116

jnehlmeier:
> Oh darn, I didn't actually see/read your comment until just now (I wrote my prior comment without having read your comment there).
> 
> Now your point is sinking in and I think I get it now. Preparing a PR with:
> 
> ```java
>   /**
>    * Execute the query with the given lock type and WAIT.
>    */
>   Query<T> withLock(Query.LockType lockType);
> 
>   /**
>    * Execute the query with the given lock type and lock wait.
>    */
>   Query<T> withLock(Query.LockType lockType, Query.LockWait lockWait);
> ```

Looks good.

> In my mind I'm thinking of keeping the existing `forUpdate()`, `forUpdateNoWait()`, `forUpdateSkipLocked()`. So we would end up with overlap in terms of 2 ways of doing the same thing, but I think that is ok.

Yes seems fine to keep the three `forUpdate` methods as a shortcut for the widely supported case.

PR #2116, opened 2020-11-25
Add Query.withLock(LockType) and withLock(LockType, LockWait)

rbygrave:


ebean-api/src/main/java/io/ebean/Query.java
```diff
@@ -1643,6 +1644,16 @@
...
+   */
+  Query<T> withLock(LockType lockType);
+
+  /**
+   * Execute the query with the given lock type and lock wait.
+   */
+  Query<T> withLock(LockType lockType, LockWait lockWait);
+
   /**
    * Execute using "for update" clause which results in the DB locking the record.
```
jnehlmeier, line 1658: Maybe add to javadoc "same as calling withLock(UPDATE, WAIT)" ? Same for the other forUpdate() methods.

PR #2130, opened 2020-12-14
Ignore special kotlin collection types

tobias-:
Kotlin prefers(emits compiler warnings if not) the built-in types
e.g. Set and MutableSet instead of the java.util.* variety. This
patch removes the imports for those, so the built-in varieties are
used.

### DIFF
--- a/kotlin-querybean-generator/src/main/java/io/ebean/querybean/generator/SimpleQueryBeanWriter.java
+++ b/kotlin-querybean-generator/src/main/java/io/ebean/querybean/generator/SimpleQueryBeanWriter.java
@@ -8,6 +8,9 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Writer;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


tobias-:
I'm not 100% sure this patch is unequivocally good.

The reason for this PR is that compiling with e.g. java.util.Set in Kotlin causes a compiler warning on the fields in question. I've thus far only got it as a side effect from https://github.com/ebean-orm/ebean/pull/2129 (jpa Transient fields included in QB), but I believe there are other cases where this will be the case, e.g. some combination of scalar type- or json-fields. I can therefore not supply you with a good test case that works after #2129 is merged.

rbygrave:
Ok.  

I think this means we should probably only merge this into a version with major or minor version bump (like version 12.7.0).  That is, not merge it into a 12.6.x patch release.

So hold this PR until we look to release `12.7.0` and then touch base at that point and confirm it's something we want to merge in at that point?

tobias-:
Sounds like a good plan. Will do. Please `@mention` me in the PR when/if you feel like discussing it.

rbygrave:
Ok @tobias-  I think we should merge this in for the 12.7.1 release.  No other comments suggesting we should not.

PR #2136, opened 2020-12-18
POM and BOM fixes

tobias-:
~* ebean-migration's version is in BOM.~
* Add ebean-postgis, ebean-redis and ebean-core-type version to BOM
~* Use old version of ebean-ddl-generator for tests to avoid
  Maven blowing up on the (not really) circular dependency.~
~* Fix compile error found by using single version of
  ebean-migration.~

### DIFF
--- a/ebean-api/pom.xml
+++ b/ebean-api/pom.xml
@@ -49,13 +49,13 @@
     <dependency>
       <groupId>io.ebean</groupId>
       <artifactId>persistence-api</artifactId>
-      <version>2.2.4</version>
+      <version>${ebean-persistence-api.version}</version>
     </dependency>
 
     <dependency>
       <groupId>io.ebean</groupId>
       <artifactId>ebean-annotation</artifactId>
-      <version>6.15</version>
+      <version>${ebean-annotation.version}</version>
     </dependency>
 
     <dependency>
--- a/ebean-bom/pom.xml
+++ b/ebean-bom/pom.xml
@@ -12,21 +12,17 @@
   <artifactId>ebean-bom</artifactId>
   <packaging>pom</packaging>
 
-  <properties>
-    <ebean-ddl-runner.version>1.0</ebean-ddl-runner.version>
-    <ebean-migration-auto.version>1.0</ebean-migration-auto.version>
-    <ebean-migration.version>12.4.0</ebean-migration.version>
-    <ebean-test-docker.version>4.1</ebean-test-docker.version>
-    <ebean-datasource.version>7.0</ebean-datasource.version>
-    <ebean-agent.version>12.6.2</ebean-agent.version>
-    <ebean-maven-plugin.version>12.6.2</ebean-maven-plugin.version>
-  </properties>
-
   <dependencyManagement>
 
     <dependencies>
 
       <!-- dependencies external to this ebean.git build -->
+
+      <dependency>
+        <groupId>io.ebean</groupId>
+        <artifactId>ebean-annotation</artifactId>
+        <version>${ebean-annotation.version}</version>
+      </dependency>
 
       <dependency>
         <groupId>io.ebean</groupId>
--- a/ebean-core/pom.xml
+++ b/ebean-core/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>io.ebean</groupId>
       <artifactId>ebean-ddl-runner</artifactId>
-      <version>1.0</version>
+      <version>${ebean-ddl-runner.version}</version>
     </dependency>
 
     <dependency>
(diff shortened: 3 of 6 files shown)


tobias-:
This is the result after fighting dependency hell for the better part of 3 days (because of an (by us) incorrectly pinned version of ebean-postgis).

The first goal was to just extend the BOM with the required dependencies that ebean cannot handle any (even minor) variation of, e.g. ebean-annotation and ebean-postgis.

It then snowballed into moving all those versions into the BOM and using the BOM in the rest of the ebean project. Which made me find a bug where two conflicting versions were required depending on the submodule. I *think* the BOM should still be good when released, but it's the one thing that's hard to test.

A question as well: Is there any reason that e.g. `12.6.4-SNAPSHOT` is hardcoded in the BOM instead of `${project.version}`?

rbygrave:
> A question as well: Is there any reason that e.g. 12.6.4-SNAPSHOT is hardcoded in the BOM instead of ${project.version}?

Yes. I prefer that as then anyone looking at the pom knows explicitly the version being used.  Hence not using ${project.version} ... and hence not that keen on a lot of this PR.

> incorrectly pinned version of ebean-postgis

I'm pretty sure you are not using composite poms to manage your [groups of] dependencies.

This PR also includes a code change to DbJson.readResource().

So a number of things here making it a PR I don't think we can merge.

tobias-:
> > A question as well: Is there any reason that e.g. 12.6.4-SNAPSHOT is hardcoded in the BOM instead of ${project.version}?
> 
> Yes. I prefer that as then anyone looking at the pom knows explicitly the version being used. Hence not using ${project.version} ... and hence not that keen on a lot of this PR.

Fair 'nuf. Let's see if I can either convince you of this or at least find something in it you want to merge.
> 
> > incorrectly pinned version of ebean-postgis
> 
> I'm pretty sure you are not using composite poms to manage your [groups of] dependencies.
Yes, since we're using gradle :wink: 

When we did use Maven, we had a `dependencyManagement` section in the root/parent pom (hence this is the first time I've used composite poms), but those artifacts were never consumed by other projects(other gits), only submodules in the same project.

> 
> This PR also includes a code change to DbJson.readResource().

Yes, this change was necessary as there is a version conflict in Ebean. `ebean-ddl-generation` depends on `ebean-migration` 12.2.0 and `ebean` depends on `ebean-migration` 12.4.0 which causes a compilation error in `ebean-test`. I will assume that `DbJson.readResource()` will not work in "production test code", at least if you use the BoM or have the dependencies in the correct order. If you have the dependencies in the wrong order your production code will have class path issues instead).

> 
> So a number of things here making it a PR I don't think we can merge.

Ok, while I want everything, I've changed the PR radically to just incorporate prioritizing the BoM versions and add the missing submodules to the BoM. I hope this is a good change as it actually finds a dependency bug in the existing code. Heads up: The code does not build, because I reverted the DbJson.readResource() fix which is required for `ebean-migration` 12.4.0.

rbygrave:
> using gradle 

To me one of the relative weaknesses of Gradle is the lack of "composite poms" to drive reuse of dependencies.  What do you do in Gradle terms to get composition of dependencies instead (that are ideally reusable across both maven and gradle projects)?  Additionally the other is the lack of maven tiles to drive composition based reuse of build plugins. Do you have a nice Gradle equivalent to maven tiles (as I didn't see that last time I looked for it).  

> Ok, while I want everything, I've changed the PR radically to just incorporate prioritizing the BoM versions and add the missing submodules to the BoM.

Ah well the PR currently additionally includes the dependencyManagement sections that I didn't want in the submodules.

tobias-:
> To me one of the relative weaknesses of Gradle is the lack of "composite poms" to drive reuse of dependencies. What do you do in Gradle terms to get composition of dependencies instead (that are ideally reusable across both maven and gradle projects)? Additionally the other is the lack of maven tiles to drive composition based reuse of build plugins. Do you have a nice Gradle equivalent to maven tiles (as I didn't see that last time I looked for it).
> 

With a caveat that I may not understand "composite poms" fully, but if we're talking about reusing/defining dependencies in one place, I see two ways, one is setting the dependencies via `configurations.all.resolutionStrategy.force` directive and one is using BoM (called "platform" in gradle) The latter can both use and produce BoM, but has a number of caveats around priority of the BoMs in case they have colliding dependencies, but the first way can be used to solve that. Imho, it's one of the more warty parts of Gradle atm, but it's getting some love, so hopefully it will soon be on par with Maven for that. A good place for platform knowledge is this link: https://docs.gradle.org/current/userguide/platforms.html

One of the nice things about Gradle platform is the ability to include a block like this in the `platform/build.gradle.kts`:
```
    dependencyManagement {
        dependencies {
            rootProject.subprojects.each {
                dependency(group: it.group, name: it.name, version: it.version)
            }
        }
    }
```

Ie, add all submodules to the platform (BoM) programmatically without having to specify the submodules again.

We also use a convention plugin (which seems to be a common way to do it) to do the common configuration for Gradle e.g. repositories, define which plugin to use and their versions, common plugins for setting up tests/lint/compilers, releases (a place where Maven outshines Gradle most imho, and which still isn't included in Gradle). 

> Ah well the PR currently additionally includes the dependencyManagement sections that I didn't want in the submodules.

Yes, and that part is the one that I was trying to use to convince you to set the dependencies once per git with :).

When you set the versions once per submodule, it's easy to get conflicting dependency versions (which you have with `ebean-migration`) and it's very hard to detect that. But I'll redo it again with variables then. Perseverance is the key to success :)

Btw, saw that you "cherry picked" the submodule version parts. Thanks for that!

tobias-:
I'll make another PR with just added ebean-annotation.

Issue #2063, opened 2020-10-07
Updating bean - L2 Bean cache update + l2 cache update on "Many Ids" when they have not change - can lead to race


rbygrave:
### Steps to reproduce

```java

	@Test
	public void testLazyLoad(){

		DB.script().run("/test-setup.sql");
		// Please execute the script in [test.sql] every time you execute the test to reset the data
		Config c1 = new QConfig().setUseCache(true).setId("config_1").findOne();
		List<ConfCp> confCpList = c1.getConfCpList();
		if(confCpList != null && !confCpList.isEmpty()){
			List<ConfCpCon> conList = confCpList.get(0).getConList();
			if (!conList.isEmpty()) {
				conList.get(0);
			}
		}
		c1.setConfigName("c1");

		Config c2 = new QConfig().setUseCache(true).setId("config_1").findOne();
		ConfCp confCp = c2.getConfCpList().get(0);

		ConfCpCon con = new ConfCpCon();
		con.setId("newCon");
		con.setEnabled("N");
		confCp.getConList().clear();
		confCp.getConList().add(con);

		c2.getConfCpList().clear();
		c2.getConfCpList().add(confCp);

		//DB.markAsDirty(c2);
		c2.save();

		c1.save();   // THIS UPDATE HERE ... will also push L2 cache update of ManyIds which have changed by c2.save()

		Config config = new QConfig().setUseCache(true).setId("config_1").findOne();
		List<ConfCp> confCpList1 = config.getConfCpList();
		if(confCpList1 != null && !confCpList1.isEmpty()){
			List<ConfCpCon> conList = confCpList1.get(0).getConList();
			if (conList != null) {
				for (ConfCpCon confCpCon : conList) {
					System.out.println("---- id: "+confCpCon.getId());
					// An exception will be thrown here
					// EntityNotFoundException: Bean not found during lazy load or refresh
					System.out.println(confCpCon.getId()+":"+confCpCon.getEnabled());
				}
			}
		}

	}
```

The issue can be seen in the logs as the last 2 `PUT` calls here should not really be made as the collection hasn't actually been mutated in terms of additions or removals.

```console
14:01:51.160 [main] DEBUG io.ebean.SUM - txn[1006] Updated [Config] [config_1]
14:01:51.161 [main] DEBUG io.ebean.TXN - txn[1006] Commit
[...]

PR #2144, opened 2021-01-11
Wrong beans in Persistcontext avoid proper lazy load

rPraml:
Hello @rbygrave,

I'm struggling with a rare NPE problem in our application. (Only certain click paths triggered the bug, in our app when lazy loading beans)
It took some time, to find a test case that reproduces the isse.
We are still using our ebean 11 fork, but the bug is still present in Ebean 12
First I did this: https://github.com/FOCONIS/ebean/commits/cache-mismatch and found at least one issue (#2063) that did some change, so that the test in this PR passes now, but I could not really explain why.

I changed the test slightly in this PR and it still fails.

Basically I am doing this three times and in the last try `cfg1.set1[0].name` is null

```java
    cfg1 = DB.find(M2MCacheMaster.class, 42);

    cfg1.getSet1().size();
    cfg1.getSet2().size();
    assertThat(cfg1.getSet1().iterator().next().getName()).isEqualTo("test");
```

Then I tried to comment out every line that is not neccessary to trigger the bug. See testM2MWithCacheMinimal

The problem seems, that the lazy load of the M2M bean does not work correctly.

My first try was to change the "DefaultBeanLoader". So the bean that is lazily loaded is in the PC (but this fixes only one of the two test cases)
My second try was not overwriting any bean in the PC (as it does not make really sense in my mind)
The put was called from "BeanDescriptor.createReference", so adding a "putIfAbsent" here, also will fix the problem
```java
      ebi.setReference(idPropertyIndex);
      if (Boolean.TRUE == readOnly) {
        ebi.setReadOnly(true);
      }
      if (pc != null) {
        Object ref = contextPutIfAbsent(pc, id, eb); // use putIfAbsent here
        if (ref != null) {
          System.out.println("DEBUG: got ref");
          return (T) ref;
        }
        ebi.setPersistenceContext(pc);
      }
      return (T) eb;
```
But then, the `createReference` method may return a bean, where the arguments `readOnly` and `disableLazyLoad` are not properly honored.

So it would be great if you find some time to take a look at this bug.

Cheers
Roland

/edit: This is my current fix
https://github.com/FOCONIS/ebean/pull/33/commits/88eee5a70a4132e936d7eb4d2a8b7c193c88267a
I think it makes sense, when creating a reference bean, that the one from the PC is taken

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
@@ -215,7 +215,7 @@ class DefaultBeanLoader {
       desc.contextPut(pc, id, bean);
       ebi.setPersistenceContext(pc);
     }
-
+    // desc.contextPut(pc, id, bean); // this will fix one of the two tests
     boolean draft = desc.isDraftInstance(bean);
 
     if (embeddedOwnerIndex == -1) {
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
@@ -296,6 +296,10 @@ public final class DefaultPersistenceContext implements PersistenceContext {
     }
 
     private void put(Object id, Object b) {
+      Object existing = map.get(id);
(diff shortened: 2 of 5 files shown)


rPraml:
accidently closed

rbygrave:
Note the fix I've gone with is slightly different in that it is more specific with where it does the check for the bean in the PersistenceContext.

PR #2146, opened 2021-01-19
Cache with lazy load doesnt work

NSzemenyei:
Hi @rbygrave ,

we found another NPE in our application, I could have extracted it in an ebean test. It fails in ebean 12 as well.

It is about @Cache and lazy loads:

One entity (ChildWithCache) has a @Cache and a property with lazy load.
Two classes have reference to this entity. If both of them load the @Cache-entity and one (non lazy load) property of it, then calling a Getter of the lazy load property leads to a null value. I think it will be loaded from the cache:

    ParentA tempA = DB.find(ParentA.class, 1L);
    tempA.getChild().getName(); //load name

    ParentB tempB = DB.find(ParentB.class, 1L);

    ChildWithCache temp = tempB.getChild();
    //if the next line is commented out, the test passes
    temp.getName(); //load name from cache --> ebean_intercept.loadedFromCache = true

    String tempLazyProp = temp.getAddress();
    assertThat(tempLazyProp).isEqualTo("Address");

Can you please take a look at this bug?

Kind regards
Noemi
FOCONIS AG (Germany)

### DIFF
--- a/ebean-core/src/test/java/org/tests/model/lazywithcache/ChildWithCache.java
+++ b/ebean-core/src/test/java/org/tests/model/lazywithcache/ChildWithCache.java
@@ -1,0 +1,52 @@
+package org.tests.model.lazywithcache;
+
+import javax.persistence.Basic;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.Id;
+
+import io.ebean.annotation.Cache;
+
+/**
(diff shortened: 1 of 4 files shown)


rPraml:
Hi @rbygrave FYI: we use this as temporary workaround.
https://github.com/FOCONIS/ebean/pull/35/commits/d518cdea158727322463f18371024151bb65af72

PR #2147, opened 2021-01-20
ADD: multiple to many outer joins cause wrong count in distinct count…

maTrojak:
hi @rbygrave, i found o problem with setCountDistinct queries where more than one @OneToMany, @ManyToMany edges (with data present) in the model cause a wrong result in the CountedValue.getCount() while calling findSingleAttributeList().


maTrojak:
@rbygrave: fix of the issue in foconis ebean 11: https://github.com/FOCONIS/ebean/pull/36/commits/9976c0733e706c14f6b8e5c85840ba5cdf76ee73

PR #2149, opened 2021-01-20
ADD: jdbc batch flush doesnt execute sql statements created in lifecycle methods (e.g. beans saved in PostInsert etc)

maTrojak:
hi @rbygrave, Issue where DB.saveAll wouldn't execute sql statements created in lifecycle methods. In the commit are test methods which reproduce the issue.


maTrojak:
Foconis fix is available in: https://github.com/FOCONIS/ebean/pull/34/commits

PR #2188, opened 2021-03-02
FIX: orderBy does not work when used on formula property 

rPraml:
or in conjunction with "exists" query

Hello @rbygrave 
We have found a severe bug, when using "orderBy" + "findIds"
- on a formula property: The property is not part of the select, so the join will not be added
- using it on an exists query (probably also other distinct queries), the order by will not be added to the select.

Please take a look at my solution. I'm not sure if the change in DefaultOrmQuery::createExtraJoinsToSupportManyWhereClause is really the right place (as method name suggests something different)

Cheers
Roland


rbygrave:
My first look says it's an ok fix.  I should get to have another look tonight.

PR #2191, opened 2021-03-09
ENH: Add Query findEach() with batch consumer

rbygrave:
This is a variation of findEach() that makes it easy to have a batch consumer processing a large result in batches.  For example, process in batches of 50 beans.

Note that the last batch consumed/processed will often have less than the batch size.


rbygrave:
FYI: I tested this out against production scale data and could take a proces from ~4 mins to just over 1 min.  So yeah cool feature - I love it and don't know why I didn't think of it before !!

PR #2194, opened 2021-03-14
Feature/json assert compare arrays

sebastian-mrozek:
The change allows finding matching elements in an array.

Could not rely on hashCode based sort as elements with additional fields (like 'id') might not sort the same way as without those (see 'assertContainsObjectsArrayShuffled' test where input json files reflect the above case)

Searching for a matching element (using 'json contains' logic) is not trivial either: there can be elements in the actual array that match multiple elements from expected set or when multiple elements from expected set match only one element in the actual array (see 'assertArrayElementsNotFound' test where input json files reflect the case)

I hope I haven't missed anything. The implementation looks a bit complex to my liking but it does work well.

The side effect of the approach is that size of arrays does not have to match, as long as the expected elements are found in the actual array. I guess that is desirable as it follows the 'contains' paradigm this utility conveys.


rbygrave:
Great stuff.

> that size of arrays does not have to match

I'm almost certain that I want them to exactly match in size.  I want this to start tighter rather than looser. I'd rather it fail tests being too tight/restrictive than for tests to pass by being too loose.  In the "assert db state" case we always expect exact size match.  

I'll merge this in and then adjust for the exact match I think.

sebastian-mrozek:
Yes, I thought if we just add array size check before searching for matching elements, this would do the job.
It could be implemented as an optional flag passed into the static `assertContains` method and the default could be strict..
```
enum JsonContainOption {
   ARRAY_SIZE_STRICT
}
```
or something like that?
I somehow feel it would be great to have an option but on the other hand can't really think of a strong case for having non-strict check.. If the array does not matter then it's probably easier to remove it from the assert file..

rbygrave:
Yeah.  I've started hacking away a bit ... maybe I'll get it sorted tonight
and merge.

PR #2198, opened 2021-03-15
Feature/json assert compare arrays

rbygrave:
Follow up to https://github.com/ebean-orm/ebean/pull/2194


rbygrave:
Adjusted the array matching with expectation that the array size is the same.  Adjusted the way that array matching works to iterate the expected array and match against the first match from actual.

PR #2212, opened 2021-03-31
Add bitwise operators to query beans

tobias-:
Bitwise operators are missing from `PShort`, `PInteger`, `PLong` and `PBigInteger`). This PR adds those missing methods to `PShort`, `PInteger` and `PLong`.

### DIFF
--- a/ebean-querybean/src/main/java/io/ebean/typequery/PInteger.java
+++ b/ebean-querybean/src/main/java/io/ebean/typequery/PInteger.java
@@ -24,4 +24,70 @@ public class PInteger<R> extends PBaseNumber<R,Integer> {
     super(name, root, prefix);
   }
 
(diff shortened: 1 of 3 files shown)


rbygrave:
Yeah nice - merging. Makes we wonder if we should add a PBaseIntegralNumber or something like that ... maybe ponder that later.

Thanks

tobias-:
@rbygrave Thanks! I considered some sort of integer base class as well, but with just one method for each operation (that takes `long flags`) type safety would suffer (e.g. it would allow the use of long values as arguments for a `PInteger`). I think it's actually better to keep it in different classes.

rbygrave:
Yes I agree.  I had a quick look after and came to the same conclusion :)

Cheers, Rob.

PR #2213, opened 2021-04-01
Initial experimental support for autoPersistUpdates / "transparent persistence"

rbygrave:
In short, when this is turned on at flush() get any dirty beans held by the PersistenceContext and persist them (always update for these beans)

### DIFF
--- a/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
+++ b/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
@@ -233,6 +233,12 @@ public class DatabaseConfig {
   private String historyTableSuffix = "_history";
 
   /**
+   * When true explicit transactions beans that have been made dirty will be
+   * automatically persisted via update on flush.
+   */
+  private boolean autoPersistUpdates;
+
+  /**
    * Use for transaction scoped batch mode.
    */
   private PersistBatch persistBatch = PersistBatch.NONE;
@@ -903,6 +909,20 @@ public class DatabaseConfig {
    */
   public void setTenantCatalogProvider(TenantCatalogProvider tenantCatalogProvider) {
     this.tenantCatalogProvider = tenantCatalogProvider;
+  }
+
+  /**
+   * Return true if dirty beans are automatically persisted.
+   */
+  public boolean isAutoPersistUpdates() {
+    return autoPersistUpdates;
+  }
+
+  /**
+   * Set to true if dirty beans are automatically persisted.
+   */
+  public void setAutoPersistUpdates(boolean autoPersistUpdates) {
+    this.autoPersistUpdates = autoPersistUpdates;
   }
 
   /**
@@ -2807,6 +2827,7 @@ public class DatabaseConfig {
     loadDocStoreSettings(p);
 
     defaultServer = p.getBoolean("defaultServer", defaultServer);
+    autoPersistUpdates = p.getBoolean("autoPersistUpdates", autoPersistUpdates);
     loadModuleInfo = p.getBoolean("loadModuleInfo", loadModuleInfo);
     maxCallStack = p.getInt("maxCallStack", maxCallStack);
     dumpMetricsOnShutdown = p.getBoolean("dumpMetricsOnShutdown", dumpMetricsOnShutdown);
@@ -3348,7 +3369,7 @@ public class DatabaseConfig {
     this.loadModuleInfo = loadModuleInfo;
   }
 
-  public enum UuidVersion {
+    public enum UuidVersion {
     VERSION4,
     VERSION1,
     VERSION1RND
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
@@ -888,6 +888,14 @@ public final class PersistRequestBean<T> extends PersistRequest implements BeanP
   }
 
   /**
+   * Remove deleted beans from the persistence context early.
+   */
+  public void removeFromPersistenceContext() {
+    idValue = beanDescriptor.getId(entityBean);
+    beanDescriptor.contextDeleted(transaction.getPersistenceContext(), idValue);
+  }
+
+  /**
    * Aggressive L1 and L2 cache cleanup for deletes.
    */
   private void postDelete() {
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
@@ -1038,6 +1038,10 @@ public final class PersistRequestBean<T> extends PersistRequest implements BeanP
     if (!publish) {
       beanDescriptor.setDraft(entityBean);
     }
+    if (transaction.isTransparentPersistence() && idValue != null) {
+      // with getGeneratedKeys off we will not have a idValue
+      beanDescriptor.contextPut(transaction.getPersistenceContext(), idValue, entityBean);
+    }
   }
 
   public boolean isReference() {
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
@@ -1038,7 +1038,7 @@ public final class PersistRequestBean<T> extends PersistRequest implements BeanP
     if (!publish) {
       beanDescriptor.setDraft(entityBean);
     }
-    if (transaction.isTransparentPersistence() && idValue != null) {
+    if (transaction.isAutoPersistUpdates() && idValue != null) {
       // with getGeneratedKeys off we will not have a idValue
       beanDescriptor.contextPut(transaction.getPersistenceContext(), idValue, entityBean);
     }
--- a/ebean-api/src/main/java/io/ebean/Transaction.java
+++ b/ebean-api/src/main/java/io/ebean/Transaction.java
@@ -60,6 +60,14 @@ public interface Transaction extends AutoCloseable {
   void register(TransactionCallback callback);
 
   /**
+   * EXPERIMENTAL - turn on transparent persistence and batchMode true.
+   * <p>
+   * With this turned on beans that are dirty in the persistence context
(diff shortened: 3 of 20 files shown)


jnehlmeier:
Surprising addition, given that JPA flush() behavior causes all kinds of annoying trouble. However it might be worth for people migrating from JPA to EBean, if that experimental mode behaves exactly as JPA flush(). That is:

- Newly created entities require a call to save() (similar to EntityManager.persist/merge()), however it must not be relevant when to call save(). Changes on the entity after calling save() need to be recognized and flushed to DB when calling flush().
- Entities retrieved from DB never require a call to save() to get changes persisted when calling flush().

Is that the case?

rbygrave:
The original discussion is at
https://groups.google.com/g/ebean/c/zGaaP7ubapk

I'll put in some answers to these questions later when I am on my computer.

Cheers, Rob.

ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
```diff
@@ -905,6 +911,20 @@ public void setTenantCatalogProvider(TenantCatalogProvider tenantCatalogProvider
     this.tenantCatalogProvider = tenantCatalogProvider;
   }
 
+  /**
+   * Return true if dirty beans are automatically persisted.
+   */
+  public boolean isAutoPersistUpdates() {
+    return autoPersistUpdates;
```
vladimirfx, line 918: I think when transparent persistence is enabled (non-soft) deletes must be deferred and runs after updates (ideally in JDBC batch).

Scenario:
1. Update changes FK (deffered)
2. Delete removes FK referenced entity from DB (not deffered)
3.  FK constraint fails on DB

Flush order:
1. Insterts (or run immediatly)
2. Updates
3. Deletes

vladimirfx (commented):
My 2 thoughts on implementation.

rbygrave:
Cut N Paste from the conversation leading to the PR:

-----
Hi!

Does it possible to implement (or emulate) JPA-like (JDO, etc) transparent persistence with Ebean?

We very exited by Ebean querying capabilities  but our business logic is comparably complex and depends on fact that on commit all changes are flushed to DB. We use repositories and just can't/won't inject it everywhere where entity is changed. In such case code becoming very procedural and MyBatis-like. 

I know there is many cases where is fine-grained control of persistence process is good. But not in DDD-like code with non anemic domain model in witch entities actually 'talks' with others and changes each other. 

P.S.: currently we use Eclipselink with its batch fetch and other extensions and considering migration to Ebean.

Regards, 
Vladimir
----

Ok, let me know if I misunderstood the questions but I think I understand.

> Does it possible to implement (or emulate) JPA-like (JDO, etc) transparent persistence with Ebean?

Specifically with JPA with beans still attached to the persistence context, on flush mutated/dirty beans are automatically persisted.  In comparison with Ebean where we expect to explicitly save() etc.

The question is, can Ebean simulate or do this such that we don't need to explicitly save() beans but instead auto persist mutated beans on flush/commit.  (I'll sort of answer below).


> on commit all changes are flushed to DB. We use repositories and just can't/won't inject it everywhere where entity is changed. In such case code becoming very procedural and MyBatis-like. 

Right, I can picture that.


> I know there is many cases where is fine-grained control of persistence process is good. But not in DDD-like code with non anemic domain model in witch entities actually 'talks' with others and changes each other. 

Ok, interesting. I don't think I have seen code quite like that. This is an interesting point and if possible I'd love to see a little example if you are able to share (obviously it might be code you are not allowed to share).


> P.S.: currently we use Eclipselink ...

Yes. I think we are seeing more of this.


Right, back to your original question.

> Does it possible to implement (or emulate) JPA-like (JDO, etc) transparent persistence with Ebean?

So does Ebean do this today?  No it does not.

Could we emulate it?  Maybe we could ...

Ebean has a PersistenceContext (io.ebean.bean.PersistenceContext, io.ebeaninternal.server.transaction.DefaultPersistenceContext). This PersistenceContext is held on the transaction (transaction scoped persistence context) and it holds all the beans that were loaded. These are the beans we would mutate (for updates). 

For the simple case of updating mutated beans:  It would be pretty much ... get the beans from the persistence context that are in dirty state, determine the order that they should be persisted (maybe the ordering of bean types does not matter per say), persist them - call database.save(dirtyBean).

What needs to be added, tested, thought about would be handling of insert and explicit delete's (orphan removal would be handled by database.save( dirtyBean )).  That is, I think it could be that we can still just do database.insert(bean) and database.delete(bean) as per normal - with batch persist on then these just go to the internal BatchControl held by the transaction so that should be fine.  There would be a question around if inserted beans should be added to the persistence context after a flush() [ideally they probably should for the case they are subsequently mutated and should be an update in a subsequent flush].

So emulating it might well be possible and not too tricky ... hmmm.

Note as a side point, that on SpiTransaction we have getPersistenceContext() and setPersistenceContext() - at this stage we have not had a good strong argument for exposing this in public API (via Transaction). If we got this all working then we could consider doing that to support "extended persistence context" - loading beans using one transaction, later transfering the persistence context to another transaction for doing more fetching and/or persisting.
  
Ok, probably worth writing some tests to simulate this and where the complexity is.  If we did manage to do this nicely then this likely would be an opt in option on a per transaction basis and globally (all transactions).

Note that when we have streaming queries findEach, findIterate, findStream ... these can use multiple persistence contexts (e.g. when iterating millions then we don't want all beans held in memory in the single persistence context) so we couldn't use this feature there but that should be expected. 


Next steps:

Create a test or 2 that "simulate" this and go from there.

-----

So playing around and we have this PR: https://github.com/ebean-orm/ebean/pull/2213

This PR/branch would be pretty close to what we would be looking for. It definitely looks like Ebean can support a "transparent persistence" mode. If you are able to have a look at that PR and look at the tests and see this looks close to what you are looking for.  


>  handling of insert 

With transaction.isTransparentPersistence() inserted beans are added to the PersistenceContext. Such that if the inserted bean is flushed and then later mutated followed by a later flush/commit - it is detected that this bean is a "dirty bean in the PC" and we get the update.  We have a test passing for this, looks good.  If getGeneratedKeys is turned off the beans 


> and explicit delete's

With transaction.isTransparentPersistence() - we need an adjustment to internals such that the bean is marked as deleted in the PersistenceContext relatively early [before flush looks for dirty beans in the PC]. We have a test passing for this, looks good.




Next steps:
- Ponder if there is some case that we are missing / not handling. Review cases of ordering of inserts vs updates when cascade persist is not used perhaps?
- Ponder if we desire some ordering of the dirty beans that are updated [In theory it is not necessary but I think it would be good to ensure some consistent ordering of the updates]
- Add a global property to DatabaseConfig to support turning this feature on globally for all transactions. This supports the case where we want it on by default for all transactions and then we explicitly turn it off when desired.
- Add an option for @Transactional to explicitly turn it on or off - something like  @Transactional(mode=DEFAULT | EXPLICIT | TRANSPARENT)?


Cheers, Rob.

rbygrave:
@jnehlmeier  - answers to your questions follow:

> Surprising addition, given that JPA flush() behavior causes all kinds of annoying trouble. 

Yes exactly. Not sure this is a good idea yet.

> However it might be worth for people migrating from JPA to EBean, 

Yes, that was the motivation for this PR.

> Newly created entities require a call to save() (similar to EntityManager.persist()), 

Yes.  em.persist(bean) -> database.insert(bean);   em.remove(bean) -> database.delete(bean);  
Noting that  em.merge(bean) is a bit interesting because we'd rather not do that but instead just use Ebean normally as we can with multiple transactions etc.


> ... Changes on the entity after calling save() need to be recognized and flushed to DB when calling flush().

That is one of the changes.  When database.insert(bean) we need to register those beans with the persistence context such that if they are mutated after a flush() the update is picked up.  There is a test for that.

>    Entities retrieved from DB never require a call to save() to get changes persisted when calling flush().

Yes. Effectively we don't need explicit use of update() and this is really the only benefit at this point.

ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
```diff
@@ -1030,6 +1038,10 @@ private void postInsert() {
     if (!publish) {
       beanDescriptor.setDraft(entityBean);
     }
+    if (transaction.isAutoPersistUpdates() && idValue != null) {
+      // with getGeneratedKeys off we will not have a idValue
```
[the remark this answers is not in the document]
rbygrave: The opposite.  No auto update when there are NO generated keys returned.  That is, for large bulk insert cases we would often turn off `getGeneratedKeys` on the transaction and in this case we can not then register those inserted beans into the PC as we don't have the id values.

This is good and expected. The ability to turn off getGeneratedKeys is an Ebean feature and not a feature provided by JPA at this point.

ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
rbygrave: Note that at this point this feature turns on JDBC batch - it is expected to run with JDBC batch turned on.

> deletes must be deferred

Probably not no. There are certainly cases where deletes must run before updates wrt OneToOne unique constraints etc. All the relative ordering is already handled in the existing cascade persist behaviour.  I don't see a change yet.

ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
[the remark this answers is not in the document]
vladimirfx: So maybe detect this collision at the configuration phase/transaction start and throw an exception? 
And certainly, state this behavior in docs for new flag.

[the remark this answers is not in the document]
rbygrave: > So maybe detect this collision at the configuration phase/transaction start and throw an exception?

What do you mean? This comment does not make sense to me.

ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
vladimirfx: I will make the PR with the test for the deferred delete use case.

rbygrave: ok cool.

PR #2230, opened 2021-05-12
Null check for soft-deleteable one-to-one exported assoc has extraneous soft-delete predicate

bayne:
Hello,

> We ran into an bug where soft-deleted entities were not being returned correctly after we updated to 12.8.3
>
> The generated query for some reason has an extraneous soft-delete predicate tacked on that is causing the result to be missing some expected rows.
> 
> select t0.id, t0.deleted, t0.organization_id, t2.id
> from esoft_del_y t0
> left join esoft_del_x t2 on t2.y_id = t0.id and t2.deleted = false
> left join esoft_del_z t1 on t1.id = t0.organization_id and t1.deleted = false and t2.deleted = false
> where t1.uuid = ? and t2.id is null and t0.deleted = false
> 
> Note the t2.deleted = false on the second join when it's expected that t2 will be all null.
> 
> I was able to reproduce the bug as a unit test here: https://github.com/bayne/ebean/commit/ad52711fe172e70d5ab2f649f163399f2dc4ce62
> 
> If I named the field anything other than "organization", the test would pass.
> 
> Thanks,
> 
> Brian

https://groups.google.com/g/ebean/c/Ms-s9pl_mQA

I tracked down the extraneous soft-delete predicate to `SqlTreeNodeExtraJoin` and one-to-one exported associations. It appears the join for the one-to-one exported is already in the `ctx` by the time it reaches https://github.com/bayne/ebean/blob/42fc41975907274f44bb58bff35f1b1f573f8054/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeExtraJoin.java#L144. This resulted in the soft-delete predicate being added again.

Let me know what you think.

Thanks!

Brian

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeExtraJoin.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeExtraJoin.java
@@ -5,6 +5,7 @@ import io.ebean.bean.EntityBean;
 import io.ebean.core.type.ScalarType;
 import io.ebean.util.SplitName;
 import io.ebeaninternal.api.SpiQuery;
+import io.ebeaninternal.server.deploy.BeanPropertyAssocOne;
 import io.ebeaninternal.server.deploy.DbReadContext;
 import io.ebeaninternal.server.deploy.DbSqlContext;
 import io.ebeaninternal.server.deploy.TableJoin;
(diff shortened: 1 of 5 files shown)


rbygrave:
I think you have got this exactly correct. Well done !!!  Apologies for taking a while to get to look at this but yes, well done.

Issue #2145, opened 2021-01-19
SoftDelete predicate missing on join to ManyToOne supporting predicate expression


rbygrave:
I want to find child by parent bean which using @SoftDelete.

This is example.

```java
    @Test
    public void testParentJoinWithSoftDelete() {
        EBasicSoftDelete bean = new EBasicSoftDelete();
        bean.setName("Joined with deleted false");
        DB.save(bean);
        final EBasicSDChild child = new EBasicSDChild(bean, "Find me from parent name", 1L);
        DB.save(child);
        
        Query<EBasicSDChild> query = DB.find(EBasicSDChild.class).where().eq("owner.name",
            "Joined with deleted false").query();
        
        List<EBasicSDChild> list = query.findList();
        // Make sure that query includes the soft delete join predicates
        assertSql(query).contains("t1.deleted = false");
        assertThat(list).hasSize(1);
        
        // Cleanup created entity
        query.delete();
        DB.deleteAllPermanent(singletonList(child));
        DB.deleteAllPermanent(singletonList(bean));
    }
```

I expect 

```
select t0.id, t0.child_name, t0.amount, t0.version, t0.deleted, t0.owner_id 
from ebasic_sdchild t0 
join ebasic_soft_delete t1 on t1.id = t0.owner_id and t1.deleted = false 
where t1.name = ? and t0.deleted = false
```

but, actually

```
select t0.id, t0.child_name, t0.amount, t0.version, t0.deleted, t0.owner_id 
from ebasic_sdchild t0 
join ebasic_soft_delete t1 on t1.id = t0.owner_id  
where t1.name = ? and t0.deleted = false
```
Is this expected behavior?  

Thanks.

----
No, this is a bug.

rbygrave:
So this is specifically on joins that are added to support query expressions - "extra joins" as opposed to joins added to support fetching.

PR #2237, opened 2021-05-24
add test testSelectOneToOneDisjunction

bkempe:
Test showing failure of join table in select, OneToOne and disjunction query


rbygrave:
Background: 
- This issue was introduced in 12.9.0 #2230
- #2230 Was added to fix #2145

PR #2250, opened 2021-06-10
Change the default of jsonDirtyByDefault to be false

rbygrave:
This means that by default DbJson beans that do not implement ModifyAwareType are not considered dirty and will NOT be included in update statements. For updating they are expected to be set with a different instance rather than mutated.


rbygrave:
This is now obsolete.

PR #2251, opened 2021-06-14
Test for transparent persistence of to-many associations

vladimirfx:
While actively using new experimental transparent persistence future (thanks Rob!) we found the first issue - changes of @*ToMany assosiations are not propagated to DB.

In this PR I've modified one of (IMO) relevant tests to reproduce the issue.

### DIFF
--- a/ebean-core/src/test/java/org/tests/transparentpersist/TestTransparentPersist.java
+++ b/ebean-core/src/test/java/org/tests/transparentpersist/TestTransparentPersist.java
@@ -8,6 +8,7 @@ import org.junit.Test;
 import org.tests.model.basic.Customer;
 import org.tests.model.basic.EBasicVer;
 import org.tests.model.basic.Order;
+import org.tests.model.basic.OrderShipment;
 
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


ebean-core/src/test/java/org/tests/transparentpersist/TestTransparentPersist.java
```diff
@@ -194,7 +197,11 @@ public void updateReferenceOnlyWithPersistCascade_Insert_andUpdateForeignKey() {
       Customer c1 = new Customer();
       c1.setName("newCust CascadePersist");
       foundOrder.setCustomer(c1);
-
+      foundOrder.getShipments().remove(0);
+      OrderShipment osh2 = new OrderShipment();
+      foundOrder.addShipment(osh2);
+      // if this is called insert is performed but no delete
```
vladimirfx, line 203: It seems that delete is also performed but not logged. See shipments collection size check - it is passed.

rbygrave: > // if this is called insert is performed but no delete

There was no `orphanRemoval = true` on the order -> shipments relationship.  As such I think it is the correct behaviour to not delete the orphan (the bean that was removed from the collection) and instead that would need to be explicitly done (via say `DB.delete(foundOrder.getShipments().remove(0))`.

I have instead made a change to the Order model adding the `orphanRemoval = true` to that relationship.

PR #2256, opened 2021-06-21
Improved @Where support, especially for Many-to-many tables

rPraml:
Hello Rob, this PR adds the following features/fixes to `@Where`

- New: `@Where` can use `${mta}` (=Many-Table-Alias) on m2m relations. Example: `@Where(clause = "${mta}.flags & 1 != 0")`
- New/Fix: If `@JoinTable` in a m2m relation references to a table, that is managed by ebean (i.e an entity or view exists), ebean will not try to create the m2m-table and will not fail with "Property $PROP has duplicate ManyToMany intersection table $TABLE Please use @JoinTable to define unique table to use"
- New: You can use the placeholder `${dbTableName}` for the current table name. This allows you to share one entity across multiple tables
- Fix: The `@Where` clause was often ignored in queries. See the modified test cases.

Cheers
Roland

### DIFF
--- a/ebean-core/pom.xml
+++ b/ebean-core/pom.xml
@@ -72,7 +72,7 @@
     <dependency>
       <groupId>io.ebean</groupId>
       <artifactId>ebean-ddl-generator</artifactId>
-      <version>12.9.0</version>
+      <version>12.9.3a</version>
       <scope>test</scope>
     </dependency>
 
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorManager.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorManager.java
@@ -445,6 +445,12 @@ public class BeanDescriptorManager implements BeanDescriptorMap, SpiBeanTypeMana
     return tableToDescMap.get(tableName.toLowerCase());
   }
 
+  @Override
+  public boolean isTableManaged(String tableName) {
+    return tableToDescMap.get(tableName.toLowerCase()) != null
+        || tableToViewDescMap.get(tableName.toLowerCase()) != null;
+  }
+
   /**
    * Invalidate entity beans based on views via their dependent tables.
    */
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyAssocMany.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyAssocMany.java
@@ -439,6 +439,7 @@ public class BeanPropertyAssocMany<T> extends BeanPropertyAssoc<T> implements ST
   @Override
   public String getAssocIsEmpty(SpiExpressionRequest request, String path) {
     boolean softDelete = targetDescriptor.isSoftDelete();
+    boolean needsX2Table = softDelete || getExtraWhere() != null;
     StringBuilder sb = new StringBuilder(50);
     SpiQuery<?> query = request.getQueryRequest().getQuery();
     if (hasJoinTable()) {
@@ -446,7 +447,7 @@ public class BeanPropertyAssocMany<T> extends BeanPropertyAssoc<T> implements ST
     } else {
       sb.append(targetDescriptor.getBaseTable(query.getTemporalMode()));
     }
-    if (softDelete && hasJoinTable()) {
+    if (needsX2Table && hasJoinTable()) {
       sb.append(" x join ");
       sb.append(targetDescriptor.getBaseTable(query.getTemporalMode()));
       sb.append(" x2 on ");
@@ -461,6 +462,16 @@ public class BeanPropertyAssocMany<T> extends BeanPropertyAssoc<T> implements ST
       }
       exportedProperties[i].appendWhere(sb, "x.", path);
     }
+    if (getExtraWhere() != null) {
+      sb.append(" and ");
+      if (hasJoinTable()) {
+        sb.append(getExtraWhere().replace("${ta}", "x2").replace("${mta}", "x"));
+      } else {
+        sb.append(getExtraWhere().replace("${ta}", "x"));
+      }
+    }
+
+    
     if (softDelete) {
       String alias = hasJoinTable() ? "x2" : "x";
       sb.append(" and ").append(targetDescriptor.getSoftDeletePredicate(alias));
@@ -1060,5 +1071,17 @@ public class BeanPropertyAssocMany<T> extends BeanPropertyAssoc<T> implements ST
    */
   public void bindElementValue(SqlUpdate insert, Object value) {
     targetDescriptor.bindElementValue(insert, value);
+  }
+  
+  /**
+   * Returns true, if we must create a m2m join table.
+   */
+  public boolean createJoinTable() {
+    if (hasJoinTable() && getMappedBy() == null) {
+      // only create on other 'owning' side
+      return !descriptor.isTableManaged(intersectionJoin.getTable());
+    } else {
+      return false;
+    }
   }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationParser.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationParser.java
@@ -129,4 +129,11 @@ public abstract class AnnotationParser extends AnnotationBase {
     }
     return columnNames;
   }
+
+  /**
+   * Process any formula from &#64;Formula or &#64;Where.
+   */
+  protected String processFormula(String source) {
+    return source == null ? null : source.replace("${dbTableName}", descriptor.getBaseTable());
+  }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultDbSqlContext.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultDbSqlContext.java
@@ -111,7 +111,7 @@ class DefaultDbSqlContext implements DbSqlContext {
   }
 
   @Override
-  public void addJoin(String type, String table, TableJoinColumn[] cols, String a1, String a2) {
+  public void addJoin(String type, String table, TableJoinColumn[] cols, String a1, String a2, String extraWhere) {
 
     if (tableJoins == null) {
       tableJoins = new HashSet<>();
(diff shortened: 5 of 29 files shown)


ebean-core/pom.xml
```diff
@@ -72,7 +72,7 @@
     <dependency>
       <groupId>io.ebean</groupId>
       <artifactId>ebean-ddl-generator</artifactId>
-      <version>12.9.0</version>
+      <version>12.9.3a</version>
```
rPraml, line 75: There are cyclic dependencies. I build it with skipTests and  then executed the maven command
```
mvn install:install-file -DgroupId=io.ebean -DartifactId=ebean-ddl-generator -Dpackaging=jar -Dversion=12.9.3a -Dfile=ebean-ddl-generator/target/ebean-ddl-generator-12.9.3-SNAPSHOT.jar
```
to put it in my local repo

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorManager.java
```diff
@@ -445,6 +445,12 @@ public void cacheNotify(TransactionEventTable.TableIUD tableIUD, CacheChangeSet
     return tableToDescMap.get(tableName.toLowerCase());
   }
 
+  @Override
+  public boolean isTableManaged(String tableName) {
+    return tableToDescMap.get(tableName.toLowerCase()) != null
+        || tableToViewDescMap.get(tableName.toLowerCase()) != null;
+  }
```
rPraml, line 452: isTableManaged means, a view or an entity with that name exists. This allows you to access the table in a m2m relation. 
See MnyEdge / MnyNode

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyAssocMany.java
```diff
@@ -461,6 +462,16 @@ public String getAssocIsEmpty(SpiExpressionRequest request, String path) {
...
       exportedProperties[i].appendWhere(sb, "x.", path);
     }
+    if (getExtraWhere() != null) {
+      sb.append(" and ");
+      if (hasJoinTable()) {
+        sb.append(getExtraWhere().replace("${ta}", "x2").replace("${mta}", "x"));
+      } else {
+        sb.append(getExtraWhere().replace("${ta}", "x"));
+      }
+    }
```
rPraml, line 472: We have to add extraWhere on the x / x2 table

ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationParser.java
```diff
@@ -129,4 +129,11 @@ protected void setColumnName(DeployBeanProperty prop, String name) {
     }
     return columnNames;
   }
+
+  /**
+   * Process any formula from &#64;Formula or &#64;Where.
+   */
+  protected String processFormula(String source) {
```
rPraml, line 136: Every argument passed by `@Where` / `@Formula` is currently routed over this method.
My future plans are, to implement a "DeployParser" alternative, so that you can register your own functions somewhere in the serverConfig: If you write something like  `ebean_BitAnd(field1, field2)` this can invoke the registered method and return a platform dependent result.

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultDbSqlContext.java
```diff
@@ -164,6 +164,14 @@ public void addJoin(String type, String table, TableJoinColumn[] cols, String a1
     if (addAsOfOnClause) {
       sb.append(" and ").append(historySupport.getAsOfPredicate(a2));
     }
+    
+    if (extraWhere != null && !extraWhere.isEmpty()) {
+      sb.append(" and ");
+      // we will also need a many-table alias here
+      sb.append(extraWhere.replace(tableAliasPlaceHolder, a2).replace("${mta}", a1));
```
rPraml, line 171: @rbygrave Why is `tableAliasPlaceHolder` specified in the databasePlatform?
Should we specify `${mta}` also in the db-platform or should we refactor databasePlatform and remove tableAliasPlaceHolder. As long as I see, it is always `${ta}` and never changed...

ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeManyRoot.java
```diff
@@ -47,16 +47,6 @@ public void appendDistinctOn(DbSqlContext ctx, boolean subQuery) {
     ctx.popTableAlias();
   }
 
-  /**
-   * append extraWhere to the join.
-   */
-  @Override
-  protected SqlJoinType appendFromAsJoin(DbSqlContext ctx, SqlJoinType joinType) {
-    SqlJoinType join = super.appendFromAsJoin(ctx, joinType);
-    super.appendExtraWhere(ctx);
```
rPraml, line 56: The super call is not needed anymore

ebean-core/src/test/java/org/tests/model/m2m/MnyNode.java
```diff
@@ -0,0 +1,129 @@
...
+    inverseJoinColumns = @JoinColumn(name = "from_id", referencedColumnName = "id"))
+  List<MnyNode> allReverseRelations;
+
+  @ManyToMany
+  @JoinTable(name = "mny_edge",
+    joinColumns = @JoinColumn(name = "from_id", referencedColumnName = "id"),
+    inverseJoinColumns = @JoinColumn(name = "to_id", referencedColumnName = "id"))
+  @Where(clause = "${mta}.flags & 1 != 0")
+  @Where(clause = "BITAND(${mta}.flags, 1) != 0", platforms = Platform.H2)
+  List<MnyNode> bit1Relations;
```
rPraml, line 40: This entity demonstrates how to use an existing entity as ManyToMany table and add certain where filters.

ebean-core/src/test/java/org/tests/query/TestImplicitJoinOnParentRelationship.java
```diff
@@ -24,11 +24,11 @@ public void test() {
     query.findList();
 
     if (isPostgres()) {
-      String expectedSql = "select distinct on (t0.id) t0.id, t0.name from o_customer t0 join o_order u1 on u1.kcustomer_id = t0.id join o_order_detail u2 on u2.order_id = u1.id join o_product u3 on u3.id = u2.product_id where u3.name = ?";
+      String expectedSql = "select distinct on (t0.id) t0.id, t0.name from o_customer t0 join o_order u1 on u1.kcustomer_id = t0.id and u1.order_date is not null join o_order_detail u2 on u2.order_id = u1.id and u2.id > 0 join o_product u3 on u3.id = u2.product_id where u3.name = ?";
```
rPraml, line 27: The query path from customer is: `orders.details.product.name`
Declaration of `orders`
```
  @OneToMany(mappedBy = "customer")
  @Where(clause = "${ta}.order_date is not null")
  List<Order> orders;
```
Declaration of `details`
```
  @Where(clause = "${ta}.id > 0")
  @OneToMany(cascade = CascadeType.ALL, mappedBy = "order")
  @OrderBy("id asc, orderQty asc, cretime desc")
  @DocEmbedded
  List<OrderDetail> details;
```
So in my mind it is correct to add the extra-where conditions here and the old code was probably wrong.

ebean-core/src/test/java/org/tests/query/TestQueryIsNull.java
```diff
@@ -17,90 +18,90 @@
...
   @Test
   public void isEmpty_when_OneToMany_expect_notExistsSubquery() {
     ResetBasicData.reset();
 
-    Query<Order> query = Ebean.find(Order.class).where().isEmpty("details").query();
+    Query<Order> query = DB.find(Order.class).where().isEmpty("details").query();
     query.findList();
 
-    assertTrue(query.getGeneratedSql().contains(" where not exists (select 1 from o_order_detail x where x.order_id = t0.id)"));
+    assertThat(query.getGeneratedSql()).contains(" where not exists (select 1 from o_order_detail x where x.order_id = t0.id and x.id > 0)");
```
rPraml, line 64: @rbygrave what do you think? We must add the extra where here and the old code was wrong IMHO. 
The fix may introduce breaking changes for other people

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultDbSqlContext.java
rbygrave: That is probably the case.  I need to review that.

ebean-core/src/test/java/org/tests/model/m2m/MnyNode.java
rbygrave: excellent

ebean-core/pom.xml
rbygrave: Ah, I think it can be just a prior version of ebean-ddl-generator.  I don't think we want 3a here.

rbygrave:
Pondering if this should go in 12.9.3 or hold for 12.9.4

ebean-core/src/test/java/org/tests/query/TestQueryIsNull.java
rbygrave: Yes, it looks like the old code was wrong.  It looks we should bump to 12.10.x with this change.

ebean-core/src/test/java/org/tests/query/TestImplicitJoinOnParentRelationship.java
rbygrave: > it is correct to add the extra-where conditions here and the old code was probably wrong.

I agree.

ebean-core/pom.xml
rbygrave: Putting this back to 12.9.2 (previously released ebean-ddl-generator).  At some point these tests that need ddl generation (most of them) need to move to break the cycle.

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultDbSqlContext.java
rbygrave: > Why is tableAliasPlaceHolder specified in the databasePlatform?

It is not actually.  Just defined locally as a static final.  We can do the same with `${mta}`

Issue #1198, opened 2017-11-02
Remove support for PropertyChangeListener from entity beans


rbygrave:
Entity beans have PropertyChangeListener support.  I announced this on the forum with no feedback to suggest anyone needs or want this so will remove this feature.

rPraml:
>  I announced this on the forum with no feedback to suggest anyone needs or want this so will remove this feature.
Oh... I did not see the announcement. I have some cases where I still use this. I think it's a good point to migrate my existing code.

buckfeng:
This change seemed to break the [example-springboot](https://github.com/ebean-orm-examples/example-springboot.git)

The following exception was thrown at line `return EbeanServerFactory.create(config)`

Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'ebeanFactoryBean': FactoryBean threw exception on object creation; nested exception is java.lang.NoSuchMethodError: io.ebean.bean.EntityBeanIntercept.preSetter(ZILjava/lang/Object;Ljava/lang/Object;)Ljava/beans/PropertyChangeEvent;
	at org.springframework.beans.factory.support.FactoryBeanRegistrySupport.doGetObjectFromFactoryBean(FactoryBeanRegistrySupport.java:175)
	at org.springframework.beans.factory.support.FactoryBeanRegistrySupport.getObjectFromFactoryBean(FactoryBeanRegistrySupport.java:103)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getObjectForBeanInstance(AbstractBeanFactory.java:1634)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:317)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.beans.factory.config.DependencyDescriptor.resolveCandidate(DependencyDescriptor.java:208)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1138)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1066)
	at org.springframework.beans.factory.support.ConstructorResolver.resolveAutowiredArgument(ConstructorResolver.java:835)
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:741)
	... 58 common frames omitted
[...]

PR #2258, opened 2021-06-28
Ask for Feedback: alternative isDirty detection for JSON objects

rPraml:
Hello @rbygrave,

I've an idea how to solve the dirty detection of JSON objects and need some feedback:

When a JSON value is read in DataReader.read (`T ret = objectReader.readValue(json, deserType);`)
I build a weakReference that tracks `ret` and acts as a Map key.
The map
```
private Map<TrackReference<T>, String> originalJson = new ConcurrentHashMap<>();
```
will store the original json for each weakReference-queue.
So when a value is returned by the read method, the ScalarType will remember the object id and the json.
If the same object will passed to the `isDirty` method, we can get the original json from the map and check if the json content is changed.

<strike>This code change should cover all of our use cases.</strike>
/edit: Unfortunaltely I loose the ability to fetch the valuePairs of changed properties in BeanPersistRequest.
For the long-term solution I only see the possibility to read an object twice.

Cheers
Roland

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
@@ -20,11 +20,18 @@ import javax.persistence.PersistenceException;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Supports @DbJson properties using Jackson ObjectMapper.
@@ -68,12 +75,12 @@ class ScalarTypeJsonObjectMapper {
       super(Set.class, jsonManager, field, dbType, docType);
     }
 
-    @Override
-    @SuppressWarnings("unchecked")
-    public Set read(DataReader reader) throws SQLException {
-      Set value = super.read(reader);
-      return value == null ? null : new ModifyAwareSet(value);
-    }
+//    @Override
+//    @SuppressWarnings("unchecked")
+//    public Set read(DataReader reader) throws SQLException {
+//      Set value = super.read(reader);
+//      return value == null ? null : new ModifyAwareSet(value);
+//    }
   }
 
   /**
@@ -86,12 +93,12 @@ class ScalarTypeJsonObjectMapper {
       super(List.class, jsonManager, field, dbType, docType);
     }
 
-    @Override
-    @SuppressWarnings("unchecked")
-    public List read(DataReader reader) throws SQLException {
-      List value = super.read(reader);
-      return value == null ? null : new ModifyAwareList(value);
-    }
+//    @Override
+//    @SuppressWarnings("unchecked")
+//    public List read(DataReader reader) throws SQLException {
+//      List value = super.read(reader);
+//      return value == null ? null : new ModifyAwareList(value);
+//    }
   }
 
   /**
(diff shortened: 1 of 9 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
```diff
@@ -149,11 +156,23 @@ public boolean isMutable() {
     }
 
     /**
-     * Return true if the value should be considered dirty (and included in an update).
+     * Return true if the value should be considered dirty (and included in an
+     * update).
      */
     @Override
     public boolean isDirty(Object value) {
-      return dirtyHandler.isDirty(value);
```
rPraml, line 156: The dirtyHandler is obsolete now, 
we could also check, if we change the interface of the 
```java
DirtyHandler {
    boolean isDirty(Object value);

    void track(Object value, String json);
}
```
and  move the functionality to the dirtyHandler.

I am also thinking, if we should extend `ModifyAwareType` or an other interface with `set/getTrackingJson()`
Then we can use the object itself to store the "tracking value" instead of the weakReference-construct

ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
```diff
@@ -163,11 +182,52 @@ public T read(DataReader reader) throws SQLException {
...
+      @Override
+      public boolean equals(Object other) {
+        return other instanceof TrackReference && this.get() == ((TrackReference<?>)other).get();
+      }
+    }
+    
+    private void track(T obj, String json) {
+      if (obj != null) {
+        TrackReference<T> ref = new TrackReference<T>(obj, queue);
+        originalJson.put(ref, json);
```
rPraml, line 224: I do not use a WeakHashMap here, as the map uses equals/hashcode of the json object and I would not find the objects again in the map.... (While writing this line, I think this is no problem)
TODO: check if weakHashMap can work as well here...

rPraml: WeakHashMap will not work here, if the same object is deserialized several times, the following may be true:
`object1.equals(object2)`  and `object1 != object2` 
if `object1` is gc'ed then you may not find `object2` in the map any more.

rbygrave:
This could be a case of just doing the simple thing which is https://github.com/ebean-orm/ebean/pull/2250

With 2250 merged for a JSON bean to be dirty it has to be a new instance that has to be set. We never look for "deep change detection".  I suspect if we merge 2250 then we might not need much if any further change.

rPraml:
Hello Rob, 
Thanks for the feedback.
We use Ebean with a lot of @DbJson fields. These are often complex datastructures, which may be modified in different places in the code. Objects may be passed to other methods, which do not have access to the root bean. For example:
```java
@Entity
class DbBean {
    @DbJson
    private JsonObj jo;
}

class JsonObj {
    private String name;
    private List<String> log;
}

void someFunc() {
   bean = DB.find(DbBean.class);
   doIt(bean.getJo());
   DB.save(bean);
} 

void doit(JsonObj o) {
   o.getLog().add("foo");
}
```
So it would mean, we have to do a lot of code change.

But the more I sleep on it, the more I like the idea in #2250.

I would need some API to intercept getter and setter calls (a bit similar like #1198)
1. It should be easy to implement ModifyAwareType. All I need is to hook into the EBI preSetter calls
2. If there would be also a "postGetter" callback, a primitive approach could be to copy the whole Json object to "originalValue"
3. It should be also possible to implement also a "copyOnDemand" behaviour. So if all json objects in a json tree implements "ModifyAwareType", they could propagate the change up to the Db-Bean.

So do you think if it is possible to get back something like #1198 (in a lightweight version)
I think to allow overriding the EntityBeanIntercept by an annotation.

Cheers
Roland

rbygrave:
> I like the idea in #2250.

Yes, but it would not help in the case where the code is like your example above.

> All I need is to hook into the EBI preSetter calls

We probably need to talk, I don't see it. Dirty detection of plain un-enhanced JsonObj needs either code added to all the types in JsonObj, or AOP or a way to compare the raw json form (like md5 of the string).   I don't see PropertyChangeListener on entity properties helping.

rbygrave:
One way to look at it is the relative distance between ScalarTypeJsonObjectMapper.GenericObject with its read() and bind() methods and BeanProperty.

- On the read path ... how to transport a md5 value of the String back so that it can be set on the entityBeanIntercept
- On the write/bind path ... convert to json String early at dirty detection time, and then when dirty use that already serialised json content to bind

rbygrave:
> So if all json objects in a json tree implements "ModifyAwareType", they ...

If they implement ModifyAwareType then that is all that is needed - it works today like that. The issue as I see it is that we desire dirty detection _without_ requiring people to do that (implement ModifyAwareType).

rbygrave:
We can close this PR now right?

rPraml:
#2274 makes this PR obsolete

PR #2259, opened 2021-06-30
Experimental: JSON-Mutable support

rPraml:
Hello Rob,
I'm still struggling with the "dirty detection" of JSONs

I have 3 different approaches now
1. My old solution, where I do a deepCopy in ScalarType https://github.com/ebean-orm/ebean/pull/1595 
2. The Idea that stores the origin-json String for each object in a Map: https://github.com/ebean-orm/ebean/pull/2258
3. This solution, that uses a special "readMutable" method, which produces a "MutableValue" that returns a new instance of each "get" call.

We use a lot of JSON data and we would need some kind of dirty detection to avoid some unneccessary data updates.
We also need the "originalValue" of any data object, as we often do an object comparison of the user data (similar to changelog)

So the first approach works best in our environment (it is also the solution we are using since some years)
Unfortunately, this doubles the memory footprint, as every json object is held in the bean value and also in the "originalValue" and MAY have also a performance impact, as the deepCopy mehtod serializes/deserializes the object again

The second approach has a good dirty recognizion, but does not provide access to the "EBI.originalValues". It is also a bit unconventional, as it uses "WeakReferences" 

The third approach (this PR) tries to store only the original JSON string in the "EBI.originalValues". In detail, the JSON string is encapsulated in a MutableValue with the scalarType, so that a call to `get()` will desereialize the JSON again.
This has a lower memory footprint and should also have a low performance impact, as deserializion only the data on demand.
Unfortunately, the current draft seems a bit complex, as I need a lot of special cases.

Maybe you can give me some advice, how to continue with this task. My preferred way would be solution 1, as I know, that it is working.

Acceptance criterias that should be fulfilled:
- Stable dirty detection on `@DbJson` annotated values (could be theoretically fulfilled by implementing ModifyAwareType, but needs a lot of additional code in our JSON-models)
- Provide access to the old/original value 

cheers
Roland

### DIFF
--- a/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -37,7 +37,7 @@ public final class EntityBeanIntercept implements Serializable {
    */
   private static final byte FLAG_LOADED_PROP = 1;
   private static final byte FLAG_CHANGED_PROP = 2;
-  private static final byte FLAG_CHANGEDLOADED_PROP = 3;
+
   /**
    * Flags indicating if a property is a dirty embedded bean. Used to distinguish
    * between an embedded bean being completely overwritten and one of its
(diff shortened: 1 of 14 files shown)


rbygrave:
We can close this PR now right?

rPraml:
#2274 makes this PR obsolete

PR #2260, opened 2021-06-30
Fix additional autotune runtimeExceptions due to toString() calls instead of asString()

casmir:
Autotune still throws RuntimeExceptions due to existing
 ORMQueryDetail.toString() calls, replaced them with ORMQueryDetail.asString()


rbygrave:
Great, thank you.

PR #2266, opened 2021-07-13
FIX: SubQuery with fetch-path does not work correctly

rPraml:
Hello Rob,

yesterday we discovered a bug when a subQuery uses a fetch path to the ID of a child entity:

**How to reproduce**
```java
subQuery = DB.find(M1.class).fetch("child","id");
query = DB.find(M2.class).where().isIn("id", subQuery);
```

**Workaround**
```java
subQuery = DB.find(M1.class).fetch("child","id");
query = DB.find(M2.class).where().isIn("id", subQuery.findSingleAttributeList());
```

(of course, the workaround may have a big performance impact, but the result would be the same)

The problem was, that the "subQuery" case selected the root model and the query looks like

select ... from t0.m2 where (t0.id) in (select **t0.m1prop1, t0.m1prop2, t1.id**, t1.id from t0.m1 join t1 ...)

While H2 accepts the wrong query, without throwing an error, other platforms may throw an error, becase the datatypes are not compareable

the "findSingleAttributeList" handles this case correctly and returend only t1.id.
So I made some test cases and fixed this in our Ebean 11 and ported it also to Ebean 12.

Cheers
Roland


ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
```diff
@@ -311,7 +311,7 @@ private SqlTreeNode buildNode(String prefix, STreePropertyAssoc prop, STreeType
 
     } else {
       // do not read Id on child beans (e.g. when used with fetch())
-      boolean withId = isNotSingleAttribute();
+      boolean withId = isNotSingleAttribute() && !subQuery;
```
rPraml, line 314: Note: in Ebean 11 I had two cases to fix: https://github.com/FOCONIS/ebean/commit/403d1e80987f2609b40d02e88a4f1344079780e9#diff-5cf825762bcb8ca9a200d86e0882dcd1d740715fcaf39872cf4e98595edd3177R336

both cases covered by a test case. Tests are passed in Ebean 12 without the second modification

ebean-core/src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
```diff
@@ -616,7 +616,7 @@ public SpiQuerySecondary convertJoins() {
    * Limit the number of fetch joins to Many properties, mark as query joins as needed.
    */
   private void markQueryJoins() {
-    detail.markQueryJoins(beanDescriptor, lazyLoadManyPath, isAllowOneManyFetch(), type != Type.ATTRIBUTE);
+    detail.markQueryJoins(beanDescriptor, lazyLoadManyPath, isAllowOneManyFetch(), type != Type.ATTRIBUTE && type != Type.SUBQUERY);
```
rPraml, line 619: Checkme: Should we also check for other types here (e.g. `Type.ID_LIST`)?

@rbygrave There are several ways how Ebean detects if we have a subQuery:
See https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java#L105
you probably know better what's right here

ebean-core/src/test/java/org/tests/query/TestSubQuery.java
```diff
@@ -47,6 +49,75 @@ public void test_IsIn() {
...
+        .query().setDistinct(true);
+
+    // execute the subQuery as copy (generatedSQL must be part of original query)
+    Query<OrderDetail> debugSq = sq.copy();
+    debugSq.findSingleAttribute();
+
+    Query<Order> query = DB.find(Order.class).select("status").where().isIn("shipDate", sq).query();
+    query.findSingleAttribute();
+    assertThat(query.getGeneratedSql())
+        .isEqualTo("select t0.status from o_order t0 where  (t0.ship_date) in (" + debugSq.getGeneratedSql() + ")");
```
rPraml, line 95: @rbygrave FYI: additional space between 
```
t0 where  (t0.ship_date)
```

rPraml:
Unfortunately, this change requires to select properties in subqueries explicitly: See https://github.com/FOCONIS/ebean/commit/285092f431732afd645d5bf9ddf91755aaccce9e

rPraml:
Rob FYI: This PR requires some updates, before it is ready to merge

ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
rPraml: Comes from this change: https://github.com/FOCONIS/ebean/pull/26. PR & Test will follow when this is merged

ebean-core/src/test/java/org/tests/query/TestQueryExists.java
```diff
@@ -89,26 +89,26 @@ public void testExists_onOneToMany() {
   public void testExists() {
     ResetBasicData.reset();
 
-    Query<Order> subQuery = Ebean.find(Order.class).alias("sq").select("id").where().raw("sq.kcustomer_id = qt.id").query();
+    Query<Order> subQuery = DB.find(Order.class).alias("sq").select("id").where().raw("sq.kcustomer_id = qt.id").query();
 
-    Query<Customer> query = Ebean.find(Customer.class).alias("qt").where().exists(subQuery).query();
+    Query<Customer> query = DB.find(Customer.class).alias("qt").where().exists(subQuery).query();
```
rPraml, line 94: removed the `.select("id")` from all exists subqueries and refactored `Ebean` to `DB`

ebean-core/src/test/java/org/tests/query/cancel/SqlQueryCancelTest.java
```diff
@@ -113,8 +113,8 @@ public void cancelOrmQueryAtBegin() throws SQLException {
     doCancelOrmAtBegin(Query::findOne);
     doCancelOrmAtBegin(q -> q.setMaxRows(1000).findPagedList().getList()); // untested
     doCancelOrmAtBegin(Query::findSet);
-    doCancelOrmAtBegin(Query::findSingleAttribute);
-    doCancelOrmAtBegin(Query::findSingleAttributeList);
+    doCancelOrmAtBegin(q -> q.select("name").findSingleAttribute());
+    doCancelOrmAtBegin(q -> q.select("name").findSingleAttributeList());
```
rPraml, line 117: specifying an attribute is mandatory now, otherwise you'll get an assertionError `selectSql was null`

rPraml (commented):
@rbygrave this PR shuld be ready now

PR #2267, opened 2021-07-16
Retrieve (C|B)LOB on SqlRow construction

jonasPoehler:
Hi Rob,

this PR fixes a possible problem, when selecting a LOB with RawSqlBuilder. The actual data wrapped by e.g. java.sql.Clob is not fetched until explicitely accessed and for that to work, we need an open connection. Since this is not always the case (see testcase) and it depends on the implementation of the JDBC driver to either eagerly or lazily fetch the Clob. H2 for instance seems to directly fetch the data meaning the tests would pass even without this fix, but other drivers don't behave like that.
**Note:** This somehow breaks the current behaviour. Currently, a call to `row.get("foo")` on a Clob field returns a java.sql.Clob object implemented by the JDBC driver. With the provided implementation a BLOB would return a byte-array and a CLOB a String. Since this also is a "silent" change without breaking compile-time API, this might need some special communicating or version update.

All the best
Jonas


rbygrave:
I have added a "reason-for-version-bump" tag that we can now use going forward.  As we are bumping up to 12.10.0 with this next release this is a good time to merge this in I think.

PR #2274, opened 2021-07-23
Add support for JSON bean mutation detection via HASH and SOURCE of JSON string content

rbygrave:
- Only converts to JSON once (at dirty detection time)

On load: 
- For HASH mode, only compute CRC32 checksum once, checksum stored on EntityBeanIntercept for mutation detection
- For SOURCE mode, store original json content on EntityBeanIntercept for mutation detection

SOURCE mode supports regenerating the 'oldValue' for change log and persist listeners. It does this by rebuilding the 'oldValue' from the original json source that was stored on EntityBeanIntercept.

Both SOURCE and HASH support BeanState.isDirty() checking.


rbygrave:
@rPraml ... maybe have a look at this and see what you think.

Cheers, Rob.

rPraml:
Hello Rob, I did a very quick (= 5 minutes) look at this. (Will take a detailed look at this next week)
Dirty detection should work this way, but I wonder if it makes sense, to store the MD5 hash instead of the whole json string.
(Tradeoff between RAM and CP'U time)

Note, we have the second use case, that we need the originalValue / valuePair for tracking the changes in a changelog.
If the whole string would be stored, it would/should be possible to reconstruct the originalValue when BeanPersistRequest.getUpdatedValue / beanState.getChangedValues is called.

Cheers
Roland

rbygrave:
Yes @rPraml  we can probably work out how to make that an option. With this current PR I think we can try to move some of the logic into the ScalarType (from BeanProperty) so I think there are a couple of adjustments to this.

rbygrave:
Ok @rPraml I think this is pretty close. Did you want to add any more test cases?

rPraml:
Hello @rbygrave I think the current solution covers all of our use cases and is ready to merge from my point of view.
Maybe you can check if  #2248 / #2250 is obsolete now.
Thanks for your support

rbygrave:
>  if #2248 / #2250 is obsolete now.

Yes, those are obsolete now @rPraml

PR #2276, opened 2021-07-28
Proof of concept to support also bean change detection / changelog on mutable properties

rPraml:
Hello Rob, I've reviewed your changes.
Dirty detection should work fine based on MD5, but I am missing the feature to get the changed and original values (as it would be required for a changelog)

I've written a proof of concept on top of your PR that adds this functionality.
Instead of holding the MD5 sum as String, I hold a `MutableJson` interface, that can decide, if a value is dirty or not. It can also recreate the original value for `ValuePairs`

I will add some review-comments. Maybe you can take a look at this.


ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -94,7 +94,7 @@
   /**
    * Holds MD5 hash of json loaded jackson beans.
    */
-  private String[] mutableHash;
+  private MutableJson[] mutableHash;
```
rPraml, line 97: Naming things: "mutableInfo" may be a better name.

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -381,7 +381,11 @@ public void setLoaded() {
     this.origValues = null;
     for (int i = 0; i < flags.length; i++) {
       flags[i] &= ~(FLAG_CHANGED_PROP + FLAG_ORIG_VALUE_SET);
+      if (mutableHash != null && mutableHash[i] != null) {
+        mutableHash[i].update(owner._ebean_getField(i));
```
rPraml, line 385: when we reset the bean (after a save) we need to update the "mutableHash" with the current value. (needs conversion TO json)

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -649,7 +653,7 @@ public void setNewBeanForUpdate() {
   public void addDirtyPropertyNames(Set<String> props, String prefix) {
     int len = getPropertyLength();
     for (int i = 0; i < len; i++) {
-      if ((flags[i] & FLAG_CHANGED_PROP) != 0) {
+      if ((flags[i] & FLAG_CHANGED_PROP) != 0 || isChangedByHash(i)) {
```
rPraml, line 656: CHECKME: should we set the `FLAG_CHANGED_PROP` when we detect a change by hash (would save some json conversions)

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -695,7 +699,12 @@ public boolean hasDirtyProperty(Set<String> propertyNames) {
   public void addDirtyPropertyValues(Map<String, ValuePair> dirtyValues, String prefix) {
     int len = getPropertyLength();
     for (int i = 0; i < len; i++) {
-      if ((flags[i] & FLAG_CHANGED_PROP) != 0) {
+      if (isChangedByHash(i)) {
+        String propName = (prefix == null ? getProperty(i) : prefix + getProperty(i));
+        Object newVal = owner._ebean_getField(i);
+        Object oldVal = mutableHash[i].get();
```
rPraml, line 705: CHECKME: should we cache the newly instantiated object in "originalValue"

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
```diff
@@ -25,11 +26,11 @@ public BeanPropertyJsonMapper(BeanDescriptor<?> desc, DeployBeanProperty deployP
...
     final String json = scalarType.jsonMapper(value);
-    final String newHash = Md5.hash(json);
-    final String oldHash = ebi.mutableHash(propertyIndex);
-    if (!Objects.equals(newHash, oldHash)) {
+    final MutableJson oldHash = ebi.mutableHash(propertyIndex);
+    if (oldHash == null || !oldHash.isEqualToJson(json)) {
       ebi.mutableContent(propertyIndex, json); // so we only convert to json once
-      ebi.mutableHash(propertyIndex, newHash); // for dirty detection next time
+      //ebi.mutableHash(propertyIndex, scalarType.jsonMutable(json)); // for dirty detection next time
+      //must be done AFTER persistControllers are called.
```
rPraml, line 33: See EBI.setLoaded

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
```diff
@@ -43,8 +44,8 @@ public Object readSet(DataReader reader, EntityBean bean) throws SQLException {
         setValue(bean, value);
         String json = reader.popJson();
         if (json != null) {
-          final String hash = Md5.hash(json);
-          bean._ebean_getIntercept().mutableHash(propertyIndex, hash);
+          final String hash = scalarType.format(value);
+          bean._ebean_getIntercept().mutableHash(propertyIndex, scalarType.jsonMutable(hash));
```
rPraml, line 48: This seems to be suboptimal. We can use "json" directly here
```suggestion
          bean._ebean_getIntercept().mutableHash(propertyIndex, scalarType.jsonMutable(json));
```

ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
```diff
@@ -66,6 +70,66 @@ public boolean isJsonMapper() {
...
+        return Md5.hash(json).equals(md5);
+      }
+
+      @Override
+      public Object get() {
+        return null; // cannot create object from json
+      }
+      @Override
+      public void update(Object obj) {
+        md5 = Md5.hash(format(obj));
```
rPraml, line 96: I tried to make that class static, to have a lower memory footprint. But I need the "update" method.

ebean-core/src/test/java/org/tests/json/TestDbJson_Jackson3.java
```diff
@@ -69,12 +95,54 @@ public void updateIncludesJsonColumn_when_loadedAndNotDirtyAware() {
...
     LoggedSql.start();
     DB.save(found);
 
-    final List<String> sql = LoggedSql.stop();
+    List<String> sql = LoggedSql.stop();
     assertThat(sql).hasSize(1);
     // plain_bean=?, no longer included with MD5 dirty detection
     assertThat(sql.get(0)).contains("update ebasic_json_list set name=?, bean_list=?, version=? where id=?");
+    
+    assertThat(EBasicJsonListPersistController.updatedValues.entrySet())
```
rPraml, line 119: I've added some tests that should work

rPraml (commented):
This is not yet perfect, but I want to share my ideas with you

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -239,6 +244,17 @@ public boolean isPartial() {
    * if any embedded beans are either new or dirty (and hence need saving).
    */
   public boolean isDirty() {
+    if (dirty) {
+      return true;
+    }
+    if (mutableHash != null) {
+      for (int i = 0; i < mutableHash.length; i++) {
+        if (mutableHash[i] != null && !mutableHash[i].isEqualToObject(owner._ebean_getField(i))) {
```
rPraml, line 252: if there are mutableHashes, I iterate over all of them and check if the hash thinks the value is dirty.
If so, the whole bean is considered as dirty. This is required for `DB.getBeanState(...).isDirty()`

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -379,8 +395,9 @@ public void setLoaded() {
     this.owner._ebean_setEmbeddedLoaded();
     this.lazyLoadProperty = -1;
     this.origValues = null;
+    this.mutableContent = null;
```
rPraml, line 398: discard mutableContent after save, as it is not longer reqired

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
```diff
@@ -25,11 +24,9 @@ public BeanPropertyJsonMapper(BeanDescriptor<?> desc, DeployBeanProperty deployP
   boolean isDirtyValue(Object value, EntityBeanIntercept ebi) {
     // dirty detection based on md5 hash of json content
     final String json = scalarType.jsonMapper(value);
-    final String newHash = Md5.hash(json);
-    final String oldHash = ebi.mutableHash(propertyIndex);
-    if (!Objects.equals(newHash, oldHash)) {
+    final MutableHash oldHash = ebi.mutableHash(propertyIndex);
+    if (oldHash == null || !oldHash.isEqualToJson(json)) {
       ebi.mutableContent(propertyIndex, json); // so we only convert to json once
-      ebi.mutableHash(propertyIndex, newHash); // for dirty detection next time
```
rPraml, line 32: moved to BindablePropertyJsonUpdate, so that it is done after the persist-controllers are invoked.

ebean-core/src/main/java/io/ebeaninternal/server/type/DataBind.java
```diff
@@ -46,12 +46,15 @@ public DataBind(DataTimeZone dataTimeZone, PreparedStatement pstmt, Connection c
 
   @Override
   public void pushJson(String json) {
+    assert this.json == null; // we can only push one value
```
rPraml, line 49: this may result in unexpected behaviour, if multiple beans are saved and a json from a previous bind is left on the "stack".

rPraml:
Hello @rbygrave , I did some updates, 
I add the MutableHash interface (which probably should be renamed to MutableInfo or something similar)

I moved the code to the scalarType, so the ScalarType can decide, if it will generare a MutableHash based on MD5 (which has effectively the same functionality as your code) or an extended MutableHash based on the originalJson which can restore the original JSON on demand.
This is required for use cases where "ValuePairs" (e.g. beanState.getChangedProps / changeLog) are needed.
I paid attention, that no unneccessary json-serialize/deserialize calls are made (if someone changes the originalValue, then it is his own fault)

I have still one unresolved problem. While plain json beans work  fine, I have problems with collections. What do you think, if we add a property to @DbJson/@DbJsonB to control the behaviour. I think we will need 2 or 3 options:
DEFAULT - uses MD5 change detection for PlainBeans and ModifyAware for List/Set/Map of beans
MD5 - uses MD5 for all type of beans
EXTENDED - uses the extended dirty/change detection mainly implemented in this PR (also for List/Set/Map)

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
rbygrave: This means that the hash is computed twice which isn't ideal.

ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonObjectMapper.java
```diff
@@ -66,6 +122,19 @@ public boolean isJsonMapper() {
...
       return formatValue(value);
     }
+    
+
+
+
+    
+    @Override
+    public MutableHash createMutableHash(String json) {
+      if (false) { // TODO should we make that configurable?
```
rbygrave, line 132: I think we could specify via the `@DbJson` annotation if the source json content should be kept or not.

`@DbJson(keepSource=true)`
MyBean myBean;

... then this method would move to the BeanProperty (well BeanPropertyJsonMapper)

ebean-api/src/main/java/io/ebean/bean/MutableHash.java
```diff
@@ -0,0 +1,28 @@
...
+   * Compares the given object to an internal value. This is an optional method, but required for proper changelog/beanState support.
+   * The implementation can serialize the object and compare it against the original json.
+   */
+  default boolean isEqualToObject(Object obj) {
+    return true;
+  }
+  /**
+   * Creates a new instance from the internal json string. This is an optional method.
+   */
+  default Object get() {
```
rbygrave, line 24: I think this is pretty much `getOldValue()`

rbygrave:
Excellent work @rPraml. I've merge that and done a few changes since. In short, I desired that we only did the checksum calculation once so we ended up with the MutableValueNext thing - but the result looks good.

> collections. ... DEFAULT , MD5, EXTENDED

I think we can do this via passing DeployBeanProperty .isKeepSource() to ScalarTypeJsonObjectMapper.createTypeFor() ... but I might not have this right in that I added `dirtyDetection()` and `keepSource()` ... which doesn't kind of support 3 values DEFAULT / MD5 / EXTENDED

```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.FIELD)
public @interface DbJson {
  ...

  boolean dirtyDetection() default true;

  boolean keepSource() default false;
```

rbygrave:
### 4 options maybe:

NONE - No dirty detection, only include in update when explicitly set
EXTENDED/SOURCE - Stores the original json, for dirty/change detection and oldValue support
HASH - Uses checksum for dirty detection, no oldValue support
DEFAULT - uses HASH for PlainBeans and ModifyAware for List/Set/Map of beans

PR #2278, opened 2021-07-29
Do not override empty json collections, when database value is null

jonasPoehler:
Currently, when a Bean initializes a DbJson Collection property with an empty Collection in the constructor, Ebean overrides that with `null`, if the column in the database is null. Usually the value might be null, when the column has been added through migration and no default has been given. This leads to problems where a seemingly non-null property becomes null, when the bean is loaded from the database. We use annotations like @Nonnull for IDE nullability-checks and kinda rely on that to add null-checks where needed. It probably is a similar (maybe worse) case for Kotlin users, where a compile-time non-null property becomes null and there simply are no null-checks then.
This PR changes the behaviour, so that if the column is null and the constructor provides a default, the default is kept, so you don't run into unexpected NPEs.


rbygrave:
Cool. This is going to conflict with the big PR for json property mutation detection (so I'll not merge this until after that).  Plus I think there might be a slightly better fix that we can look at.

So all good but I'll look to merge that other PR in first and go from there.

Cheers, Rob.

PR #2280, opened 2021-07-30
Shows why we should really ignore ModifyAwareType for all Jackson beans

rbygrave:
We should NOT merge this.  Just demostrates the issues we have with ModifyAwareType on Jackson beans.

Pretty much only SOURCE and HASH (and NONE) modes work as we would expect. Relying on the ModifyAwareType is confusing or buggy.

So instead for ModifyAwareType beans lets just default to HASH based (and not do this). Arguably we could use HASH for List, Set and Map for all the Jackson ObjectMapper cases just have SOURCE, HASH or NONE (and ignore  ModifyAwareType).

@rPraml


rPraml:
@rbygrave  I would agree, **not** to merge that in.
I think what I've shown here is an edge case and if it is documented, it would be OK to use only SOURCE/HASH detection as it should be 100% relyable now.

rbygrave:
Good stuff - discarding this @rPraml

Issue #2261, opened 2021-07-01
Refactor BindHash: Using MD5 to avoid collisions


rPraml:
I noticed, that Ebean still uses hashCode for bindHash computation,

This has a risk of collisions. especially if the prime factor 92821 plays a role (which was introduced in #869)

## Expected behavior

same queries with different bind values must not hit the cache.

## Actual behavior

query with different bind values may hit the cache, but should not.

### Steps to reproduce

The query `q.between("prop", 0, 92821)` and `q.between("prop", 1, 0)` will have the same hash
because of: 
https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/expression/BetweenExpression.java#L52

### Possible solution

I've fixed this in Ebean 11: https://github.com/FOCONIS/ebean/commit/1344f790577c687359db589ec856baa9b2c0c70d
Maybe you can take a look at it. Especially if you think, if the algorithm should be configurable (https://github.com/FOCONIS/ebean/commit/1344f790577c687359db589ec856baa9b2c0c70d#diff-fdfd7023c25af12fa4d0bc1104850a905c18eaffdb90c1577111e2162f39fc60R299) or if we should use MD5 as default
If you want, I can port this change to 12

rbygrave:
> If you want, I can port this change to 12

Yes please. This does look like a nice improvement.

PR #2288, opened 2021-08-06
Change query metric hash from MD5 of sql + name + loc to Checksum of sql (+ rename to sqlHash)

rbygrave:
- Change query metric hash from MD5 of sql + name + loc to Checksum of sql
- Rename hash to sqlHash to better reflect what it is (CRC32 checksum o…


rPraml:
Hello @rbygrave,
It seems, that you replace all MD5 computations with CRC32 checks. (I think your intention is performance)
But is it really the correct way? There may exist uncritical areas (like metrics) where a hash collision does not really harm, or other areas, like migration scripts, where a collision would be noticed immediately.

But there are also some more critical places, where a hash collision may produce wrong results. Especially in the QueryBindHash #2261. (Personally I didn't like that from the beginning to rely only on the Object.hashCode())

So in my oppinion, on places that can produce data corruption or wrong cache hits, we need a stable enough hash computation. I'm  afraid, that a 32 bit value is enough here. (with ~10000 values in a cache, the probability rises up to 1%)
https://stackoverflow.com/questions/14210298/probability-of-collision-when-using-a-32-bit-hash
Note: you have a `long` in your code, but CRC32 does only produce 32 bits
64 bit would probably be enough, but the more bits, the better ;) 

So what are your plans about checksums/hashes

Cheers
Roland

rbygrave:
>  (like metrics) 

That was this change and it was not being used as a "key" per say, the only user of this hash is my ebean-insight backend (nothing in ebean itself).


> like migration scripts, where a collision would be noticed immediately.

It has always been CRC32 checksum there.  We do this _exactly_ the same as flyway.  The only thing I want to do is move that calculation forward before the expression evaluation as per the logged ticket in ebean-migration.


> QueryBindHash

That is a different thing. I do believe you are preparing a PR for that maybe as per the issue you logged.


> on places that can produce data corruption or wrong cache hits, we need a stable enough hash computation.

Yes. This is all about QueryBindHash which didn't use Md5 at all. This change didn't impact that at all. Yes we can look to improve QueryBindHash.

rbygrave:
Hi All, @rPraml ,

I'm pretty much reverting this change in that, for a query metrics reporting perspective this PR/change is the wrong fix.  I'm preparing a PR that almost reverts all of this change.

The essence of the original issue that this tried to solve is that the metric hash was not stable with respect to code changes making the source line change. This PR took the wrong approach to fix that issue. The correct fix for the metrics hash was to adjust the location part so that it didn't line the source line - "location" moving forward is package + type + method name (and not including the source file and line number).

So preparing that PR now and also note that this puts it back be being a metric plan hash (not a sqlHash, and we drop locationHash).

My belief is that the only thing really using the query plan metrics is ebean insight (basically me). ebean insight desires the use of the "metric hash" but it isn't a requirement for other consumers to use the "metric hash" - for example, they could instead just use the SQL (and somewhat ignore the location and name aspect of the metrics).

PR #2295, opened 2021-08-09
Use MD5 for QueryBindHash

rPraml:
As mentioned in #2261 - the current hashCode implementation has a high risk of collisions.
This PR use MD5 sum for QueryBindHash computation, which should minimize the risk.

It is prepared to be configurable, so it shoud be easy to switch from MD5 to any other MessageDigest, like SHA1 or SHA256
or even back to the old HashCode algorithm.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
@@ -54,12 +54,11 @@ public class BindParams implements Serializable {
     positionedParameters.clear();
   }
 
-  public int queryBindHash() {
-    int hc = namedParameters.hashCode();
+  public void queryBindHash(BindHash hash) {
+    hash.update(positionedParameters.size());
     for (Param positionedParameter : positionedParameters) {
-      hc = hc * 92821 + positionedParameter.hashCode();
+       positionedParameter.queryBindHash(hash);
     }
-    return hc;
   }
 
   /**
@@ -420,6 +419,10 @@ public class BindParams implements Serializable {
       hc = hc * 92821 + (type);
       hc = hc * 92821 + (inValue == null ? 0 : inValue.hashCode());
       return hc;
+    }
+
+    void queryBindHash(BindHash hash) {
+      hash.update(isInParam).update(isOutParam).update(type).update(inValue);
     }
 
     @Override
--- a/ebean-core/src/main/java/io/ebeaninternal/server/expression/BitwiseExpression.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/expression/BitwiseExpression.java
@@ -1,5 +1,6 @@
 package io.ebeaninternal.server.expression;
 
+import io.ebeaninternal.api.BindHash;
 import io.ebeaninternal.api.SpiExpression;
 import io.ebeaninternal.api.SpiExpressionRequest;
 
@@ -39,8 +40,8 @@ class BitwiseExpression extends AbstractExpression {
   }
 
   @Override
-  public int queryBindHash() {
-    return Long.hashCode(flags);
+  public void queryBindHash(BindHash hash) {
+    hash.update(flags).update(match);
   }
 
   @Override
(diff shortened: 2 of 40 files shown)


ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
```diff
@@ -54,12 +54,11 @@ public void reset() {
     positionedParameters.clear();
   }
 
-  public int queryBindHash() {
-    int hc = namedParameters.hashCode();
+  public void queryBindHash(BindHash hash) {
+    hash.update(positionedParameters.size());
```
rPraml, line 58: all `queryBindHash` are changed. For collections, I changed the format also to append the size of the collection, otherwise you may run into collisions if you may have queries like these: `in("prop1", 1,2).in("prop2", 3)` `in("prop1", 1).in("prop2", 2, 3)` which have the same hash (also if MD5 is used)

ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
```diff
@@ -422,6 +421,10 @@ public int hashCode() {
       return hc;
     }
 
+    void queryBindHash(BindHash hash) {
+      hash.update(isInParam).update(isOutParam).update(type).update(inValue);
```
rPraml, line 425: CHECKME: I've added queryBindHash here. I think `equals` (some lines below) is not correctly implemented here, as it relies only on hashCode

ebean-core/src/main/java/io/ebeaninternal/server/expression/BitwiseExpression.java
```diff
@@ -39,8 +40,8 @@ public void queryPlanHash(StringBuilder builder) {
   }
 
   @Override
-  public int queryBindHash() {
-    return Long.hashCode(flags);
+  public void queryBindHash(BindHash hash) {
+    hash.update(flags).update(match);
```
rPraml, line 44: There was an error in the old implementation. `match` was not added to queryBindHash

ebean-core/src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
```diff
@@ -1298,8 +1300,10 @@ public int queryBindHash() {
   public HashQuery queryHash() {
     // calculateQueryPlanHash is called just after potential AutoTune tuning
     // so queryPlanHash is calculated well before this method is called
-    int hc = queryBindHash();
-    return new HashQuery(queryPlanKey, hc);
+    BindHash hash = bindHashAlgorithm == null ? new HashCodeBindHash() : new MdBindHash(bindHashAlgorithm);
```
rPraml, line 1303: @rbygrave  
- Should I make the whole thing configurable via ServerConfig or can/should we use MD5 as default?
- Should I keep the old (dangerous) HashCodeBindHash implementation?

ebean-core/src/main/java/io/ebeaninternal/server/querydefn/MdBindHash.java
```diff
@@ -0,0 +1,138 @@
...
+    md.update(boolValue ? (byte) 1 : (byte) 0);
+    return this;
+  }
+
+  @Override
+  public BindHash update(Object value) {
+    if (value == null) {
+      md.update((byte) 0);
+
+      // do some special handling for known object types
```
rPraml, line 72: I think it makes sense, to use some special handling for certain classes
(these are the datatypes we mainly use in our application)
for all others, fall back to hashCode. This class can be easily extended to add other datatypes

@rbygrave maybe I make it in a way configurable, that you can specify the className in the ServerConfig or load it with ServiceLoader. What do you think?

ebean-core/src/test/java/org/tests/cache/TestQueryCache.java
```diff
@@ -299,4 +298,35 @@ public void findIds() {
     assertThat(sql).hasSize(1);
   }
 
+  @Test
+  public void findCountDifferentQueriesBit() {
+    DB.getDefault().getPluginApi().getServerCacheManager().clearAll();
```
rPraml, line 303: I've added this test, which discovered the initial problem, the missing `match` in `BitwiseExpression` and the I noticed, that it would be easy to produce hashCollisions if the magic number `92821` is used.

I also refactored that class to use `DB` instead of `Ebean`

rPraml (commented):
Hello Rob, this is my draft how to use MD5. My main question is, if/how should I make it configurable to select between different implementations.

Cheers
Roland

PR #2298, opened 2021-08-10
Foconis md5 for querybindhash

rbygrave:


ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
```diff
@@ -424,7 +420,15 @@ public int hashCode() {
...
-      return o != null && (o == this || (o instanceof Param) && hashCode() == o.hashCode());
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      Param param = (Param) o;
+      return encryptionKey == param.encryptionKey && isInParam == param.isInParam && isOutParam == param.isOutParam
+        && type == param.type && Objects.equals(inValue, param.inValue);
+    }
+
+    void queryBindHash(BindHash hash) {
+      hash.update(isInParam).update(isOutParam).update(type).update(inValue);
```
rPraml, line 431: has `encryptionKey` also part of the bindHash?

ebean-core/src/main/java/io/ebeaninternal/server/querydefn/HashCodeBindHash.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * HashCode builder that uses Object.hashCode for computing bind-hashes.
+ * This is a fast and lightweight implementation, but may produce collisions.
+ *
+ * @author Roland Praml, FOCONIS AG
+ */
+public class HashCodeBindHash implements BindHash {
+
+  private final List<Object> values = new ArrayList<>();
```
rPraml, line 17: Good idea, this would make MD5 complete obsolete.
BTW: we do not need a separate "hashCode" computation, we can use just `values.hashCode()`

ebean-core/src/main/java/io/ebeaninternal/api/BindHash.java
```diff
@@ -0,0 +1,37 @@
...
+   * Update with object value.
+   */
+  BindHash update(Object value);
+
+  /**
+   * finishes the hash. May be used to compute internal state. After finish, no
+   * update method must be called
+   */
+  void finish();
+
```
rPraml, line 36: As HashCodeBindHash is just a holder for all values, we could think to change the `public interface BindHash`
to 
```java
public class BindHash {
   private final List<Object> values = new ArrayList<>();
   public void update(Object value) { values.add(value); }
   public int hashCode() { return values.hashCode; }
   public boolean equals(Object o) { ... }
}
```
Is `BindHash` still the right word. 

BTW: I like the idea, to store the object values, and remove the MdBindHash as it is impossible now to get a collision. 

But do we have to pay attention, so that we do not run in an OOM problem when very long strings/big byte arrays ... are used?
(We had the problem with the serverCache, thats why we made 
https://github.com/FOCONIS/ebean/commit/44193f36e5d1c9c67ac50ada9193b06e8cda7425 - pr will come)

ebean-core/src/main/java/io/ebeaninternal/api/BindParams.java
rbygrave: Looking again ...

rbygrave: No. Removing it.

ebean-core/src/main/java/io/ebeaninternal/api/BindHash.java
rbygrave: > problem with the serverCache, thats why we made

Yeah, nice !!

PR #2301, opened 2021-08-10
Updating ebean-ddl-generator version in ebean-core

ThomasFOC:
Hello together, I think the version was behind the currently needed. Am I right? Without this "fix" the current maven compilation fails locally.


rbygrave:
Great, thank you !!

PR #2304, opened 2021-08-11
Support for Sequences in MariaDB

ThomasFOC:
### Foreword
Note that with this PR, apps using MariaDB and with `GeneratedValue(strategy=GenerationType.SEQUENCE).` will start using DB sequences (before this PR with MariaDB it would always use identity).

The expectation is that this generally won't happen in that generally apps will use "all sequences" or "all identity" and setting the generation strategy explicitly on a per entity basis is extremely rare.  People hitting this would then remove that generation strategy to then default to the platform choice which for MariaDB is still `Identity`.

If people hit this issue or think this is a problem please let us know. 
Thanks, Rob.

------------------
Hello rob,
since Version 10.3 MariaDB supports sequences:
https://mariadb.com/kb/en/sequence-overview/

This allows us to implement the `SequenceGenerator`.

Due to the still valid version 10.2 of MariaDB this maybe not good, but due to the end of life this may not be critical:
https://en.wikipedia.org/wiki/MariaDB#Versioning


ebean-api/src/main/java/io/ebean/config/dbplatform/mariadb/MariaDbPlatform.java
```diff
@@ -11,6 +16,21 @@
...
     super();
     this.platform = Platform.MARIADB;
+    this.sequenceBatchMode = false;
     this.historySupport = new MariaDbHistorySupport();
   }
+  
+  @Override
+  protected void configureIdType(IdType idType) {
+    if (idType == IdType.SEQUENCE) {
+      this.dbIdentity.setSupportsSequence(true);
```
rbygrave, line 26: Hmmm, I think we should do this on the constructor and not here.

ebean-core/src/test/java/org/tests/idkeys/TestGeneratedKeys.java
```diff
@@ -49,30 +48,43 @@ public void testSequence() throws SQLException {
...
+          rs.next();
+          return rs.getLong(1);
+
+        case MARIADB :
+          rs = stm.executeQuery("select previous value for " + sequence);
+          rs.next();
+          return rs.getLong(1);
+
+        case SQLSERVER :
+          rs = stm.executeQuery(
```
rbygrave, line 71: We can get some reuse here.

ebean-core/src/test/java/org/tests/idkeys/db/GenKeySequence.java
```diff
@@ -4,12 +4,14 @@
...
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
+import javax.persistence.SequenceGenerator;
 
 @Entity
 public class GenKeySequence {
   public final static String SEQUENCE_NAME = "SEQ";
 
   @Id
+  @SequenceGenerator(name = "SEQ_NAME", sequenceName = GenKeySequence.SEQUENCE_NAME)
```
rbygrave, line 14: Hmmm, I think this should be removed / we should not add `@SequenceGenerator` here.

ThomasFOC: In Ebean 11 `@SequenceGenerator(name = "SEQ_NAME", sequenceName = GenKeySequence.SEQUENCE_NAME)` is annotated on the Entity itself. This is no longer possible.

* rith this annotation the sequence is named `SEQ` in the migration scripts.
* without this annotation the sequence is named `SEQ_NAME` in the migration script (and such the test fails currently).
* renaming the static variable `SEQUENCE_NAME = "SEQ_NAME"` may also fix this test.

I decided to reuse the solution of Ebean 11 by annotating the field, because I thought this was done on purpose ;)

Should I remove the annotation and rename `SEQUENCE_NAME = "SEQ_NAME"` ?

ebean-api/src/main/java/io/ebean/config/dbplatform/mariadb/MariaDbPlatform.java
rPraml: Our idea was, not to enable sequences by default. It has to be explicitly enabled in config:

If someone switches to a newer ebean versions, all migrations would use "SEQUENCE" from then on - and this makes it incompatible with MariaDB <10.3

ebean-core/src/test/java/org/tests/idkeys/db/GenKeySequence.java
rbygrave: It isn't clear why this expects and needs the additional `@SequenceGenerator` when `@GeneratedValue` has those attributes.  So we need to understand why that is / is there a behaviour change here etc

ebean-api/src/main/java/io/ebean/config/dbplatform/mariadb/MariaDbPlatform.java
rPraml: I feared that SEQUENCE is used if someone has used `GeneratedValue(strategy=GenerationType.SEQUENCE)`.
Old behaviour: AUTO_INCREMENT is used for mariadb, SEQUENCE will be used for other platforms
New behaviour: SEQUENCE will be used also for mariadb. The user has to update/migrate existing tables to SEQUENCE

ebean-core/src/test/java/org/tests/idkeys/db/GenKeySequence.java
ThomasFOC: We removed this change from this pull request, as it does not have anything to do with sequences in MariaDB (except that we wanted to test sequences).

We provided a different pull request to address this situation:
https://github.com/ebean-orm/ebean/pull/2308

ebean-core/src/test/java/org/tests/idkeys/TestGeneratedKeys.java
ThomasFOC: I've used the 🧹 to tidy up the code. MariaDB seems not to support this approach.

PR #2305, opened 2021-08-11
multi tenancy aware sequences

ThomasFOC:
Hey Rob,
the `TestSequenceMultiTenant` shows the problem with multi tenancy. To address this problem we implemented a delegator in the `BeanDescriptorManager`.

There is an older attempt of @rPraml where he uses `TenantDataSourceProvider` and `CurrentTenantProvider` and extends them to the `SequenceIdGenerator`:
https://github.com/FOCONIS/ebean/commit/2de1f215bbb164eef21b3f5cbe9f3d8475a78353

Do you have any other suggestions or is this simple attempt sufficient?


ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorManager.java
```diff
@@ -1294,7 +1297,39 @@ private void checkMappedByManyToMany(DeployBeanPropertyAssocMany<?> prop) {
...
+    return new PlatformIdGenerator() {
+
+      private Map<DataSource, PlatformIdGenerator> map = Collections.synchronizedMap(new WeakHashMap<>());
+
+      private PlatformIdGenerator create() {
+        return databasePlatform.createSequenceIdGenerator(backgroundExecutor, dataSourceSupplier.getDataSource(), stepSize, seqName);
+      }
+
+      private PlatformIdGenerator get() {
+        return map.computeIfAbsent(dataSourceSupplier.getDataSource(), k -> create());
```
rPraml, line 1309: FI: This works good for a DS-supplier that supplies different dataSources for different tenants (As it is done in TenantMode.DB/DB_WITH_MASTER - which is our use case @foconis) - for other tenant modes like SCHEMA or CATALOG it may cause other problems. Maybe in a schema/catalog environment this should somehow configurable, where the Sequence lives.

rbygrave:
Hmmm, my initial thought is that this looks problematic.

ThomasFOC:
> Hmmm, my initial thought is that this looks problematic.

Hello @rbygrave ,
we tried to provide a test case for sequences in `TenantMode.SCHEMA`:
https://github.com/FOCONIS/ebean/commit/109958a1339f76f8dc336fc2ba29d7ad6ea78c45

We have tried to generate the schemas, when there is a `UserContext`-switch and provide the new schema.

An exception is raised in `io.ebean.config.dbplatform.SequenceIdGenerator.getMoreIds(int)` in line 158. To be more precise the exception comes from `io.ebeaninternal.server.core.MultiTenantDbSchemaSupplier.SchemaDataSource.getConnection()` where the schema is set for the connection.

This seems correct for us and may address the issue @rPraml sees.

But to dive deeper into this topic the test has to run over the problem that no schema seems to be generated.

After multiple hours of debugging and searching we are at our wits' end. Do you have an idea how to test `TenantMode.SCHEMA` with multiple schemas?

ebean-test/src/test/java/org/tests/idkeys/TestSequenceMultiTenant.java
```diff
@@ -66,32 +68,100 @@ public DataSource dataSource(Object tenantId) {
...
+
+    UserContext.set("5711", "2");
+    assertEquals(1L, db.nextId(GenKeySequence.class));
+
+    UserContext.set("4711", "1");
+    assertEquals(3L, db.nextId(GenKeySequence.class));
   }
 
-  private static DataSource createDataSource(Object tenantId) {
+  private Database setupSchema() {
```
ThomasFOC, line 118: With the support of @rPraml we were able to write a failing test for tenantmode schema. THe main question seems to be, were the sequence schould live.

rPraml:
Statusupdate: We know, that sequences in a multi-tenant environment are problematic as the sequence may be stored in the wrong database. We do NOT use them. This PR is not part of our fork.

@rbygrave I think the problem still exists... If you decide to close this PR, then it should be mentioned as limitation in multi-tenant setups with one DB/schema per tenant.

rPraml:
Hi @rbygrave  we will not pursue this topic any further as we do not need multi-tenant sequences.
You can decide, to close this issue (and maybe add a note to doc about known limitations on sequences + multi-tenant support)

rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3805

PR #2306, opened 2021-08-11
Use SoftReferences in Cache

jonasPoehler:
Over the last time, we started using the Ebean-cache more and more in order to gain some performance. But with that performance comes the problem of high heap usage up to the point where the JVM can't free any more space and we ran into an OOM situation. With this enhancement, the JVM is able to free cache elements from memory, if they are not used by any other code than the server cache. A subsequent cache lookup will result in a cache miss and thus an a fetch from database as if the cache was invalidated or not filled in the first place.
The one drawback to this change is, that the GarbageCollection does not care, which element to clear first, so it might collect an element just recently fetched from the database, while it leaves another one, that is much older.
I hope that this change is acceptable, nevertheless I'm open to suggestions on how to improve this change.


jnehlmeier:
While using SoftReferences makes sense for a Cache, the default implementation of the cache in ebean still uses an unbounded ConcurrentHashMap. While it is fast, it is unbounded. So even with SoftReferences you might not be protected against a slow JVM, because with already high heap usage (caused by non-ebean code) you might re-fill the cache fast enough to constantly trigger lots of GC calls and slowing down the JVM. 

Thus it is usually better to use a bounded cache with good eviction policy. An excellent choice would be https://github.com/ben-manes/caffeine as a default implementation. It supports time based, size based and reference based (weak/soft keys and/or values) eviction and is pretty fast given it is a bounded cache: https://github.com/ben-manes/caffeine/wiki/Benchmarks

jnehlmeier:
> While using SoftReferences makes sense for a Cache, the default implementation of the cache in ebean still uses an unbounded ConcurrentHashMap. While it is fast, it is unbounded. So even with SoftReferences you might not be protected against a slow JVM, because with already high heap usage (caused by non-ebean code) you might re-fill the cache fast enough to constantly trigger lots of GC calls and slowing down the JVM.

What I mean is that even though ebean has some cache settings like TTL, max size, etc. which makes it look like a bounded cache, it only evicts periodically using a fixed interval, by default 60 seconds. So at worst you could eat up all memory within this 60 seconds as the underlying ConcurrentHashMap is unbounded. I guess that is what you are experiencing.

Caches like Caffeine can evict periodically but also do it occasionally during read/write operations. So it is less likely that you run out of memory even with high write rates. Obviously this costs some performance, but you end up with a more controlled memory footprint.

rPraml:
@jnehlmeier thanks for your feedback.
I would agree that the clean up of the soft references may slow down due GC.
(but a slow JVM is still better than a crashing one)
Any yes, the current implementation is not quite optimal in terms of eviction.

A bounded cache still does not resolve problems IMHO. You can eat up the memory with also a few heavy objects.
The cache would have to determine the effective memory footprint of the cached data, which isn't trivial, especially
for local caches. There are some [ObjectSizeCalculators](https://github.com/JetBrains/jdk8u_nashorn/blob/master/src/jdk/nashorn/internal/ir/debug/ObjectSizeCalculator.java) but they will not work very good for objects that have references to global objects (like a Bean knows its EbeanServer)

I analyzed the types of cached data: There are two different types of cache:

- `DefaultServerCache` (which can be replaced by a Redis/Hazelcast/Ignite cache. Objects have to be serialized)
- `DefaultServerQueryCache` (which is always local and uses Redis etc. only for clear() notification)

The `DefaultServerCache` only uses keys and values that are `Externalizable` (like primitives,  `CacheBeanData` or `CachedManyIds`), so it would be possible to estimate the memory footprint of the cache value. (We make excessive usage of ID caches and we cache ID lists with sizes up to 10000 entries)

The `DefaultServerQueryCache` uses non serializable objects like `HashQuery` for the cache key¹ and QueryCacheEntry for the values. QueryCacheEntry can contain anything, like a `BeanList` which can have a memory footprint of several megabytes.

¹ @rbygrave As already mentioned [here](https://github.com/ebean-orm/ebean/pull/2298#discussion_r685945193) we can run in memory problems, when the query cache is used for large bind values. But I would say, this is the programmer's fault, if queryCache is enabled for that kind of queries

----
What I want to say is: @jonasPoehler 's change will improve the current behaviour. Things can always be done better, but this would mean that more parts of the cache implementation will have to be rewritten.

@rbygrave **Important:** Due code review I may have found a severe bug when a cache plugin is used in a multi-tenant environment.
We noticed years ago, that the cache was not tenatAware. Thats why the `TenantAwareKey` was introduced with this [commit](https://github.com/ebean-orm/ebean/commit/88c6ec358c86d33ba1573515f9b37c6855bd26ef). As far as I understand, this is only used for Ebean's default implementation, but I do not see any tenant handling in Redis/Hazelcast/Ignite.
The easiest fix could be, to remove the teant-handling from `DefaultServerCache` and add a `TenantAwareServerCacheDelegate` in the DefaultCacheHolder that delegates all calls to the actual implementation with a modified cache key.

@rbygrave / @jonasPoehler There is some [demonstration code](https://github.com/FOCONIS/ebean/blob/master/src/test/java/io/ebeaninternal/server/cache/DefaultServerCacheTest.java#L64). Shoud this be added in a (disabled) test mehthod?

jnehlmeier:
> @jnehlmeier thanks for your feedback.
> I would agree that the clean up of the soft references may slow down due GC.
> (but a slow JVM is still better than a crashing one)

From my experience if the JVM is in a state that it needs to continuously GC then you don't want customers to work on it because it is too slow to do anything meaningful. Especially because SoftReferences will always survive one GC if they have been accessed since the last GC. So cache hits actually hurt in that situation.

I am not against this change and I am pretty sure it will solve some problems for you, but it is not the silver bullet. You might still see problems once your JVM runs long enough with reasonable amount of workload.


> Any yes, the current implementation is not quite optimal in terms of eviction.
> 
> A bounded cache still does not resolve problems IMHO. You can eat up the memory with also a few heavy objects.
> The cache would have to determine the effective memory footprint of the cached data, which isn't trivial, especially
> for local caches. There are some [ObjectSizeCalculators](https://github.com/JetBrains/jdk8u_nashorn/blob/master/src/jdk/nashorn/internal/ir/debug/ObjectSizeCalculator.java) but they will not work very good for objects that have references to global objects (like a Bean knows its EbeanServer)

Yes sure, if you have a large value, you still can eat up memory with bounded caches.  In case your cache entries have hugely different memory footprints Caffeine actually has an interface `Weigher` that assigns a weight to each cache entry and a maximum cache weight can be configured. While you can use memory size estimators to calculate the weight you could also use a simpler measure, depending on the data you put into the cache. In your ID list example you could use the list length and some roughly estimated (hardcoded) factor for each list entry. At least it gives you more control in this area so you have more predictable memory consumption.

jonasPoehler:
First of all, thanks for the feedback. What the discussion shows, is that is unclear, what a good and efficient cache for Ebean should look like. The current implementation supports clearing old entries through the use of a periodically executed task. This does not protect from OOMs, because the task clears entries based on how old the are, rather than if the JVM can hold any more or not. While our proposed implementation is not very selective, it allows the application to use memory more efficiently by caching more data, while it can (only being limited by the removal of old entries) and reducing the amount of cache use when the general heap allocation is higher. Of course this comes, as pointed out by @jnehlmeier with some overhead, worst case the GC and Ebean playing basically Ping Pong with each other. 
To give you a practical example: We are running a user-facing applications, where we can observe that the heap allocation over a day follows the user count (with a peak around noon). With a cache considering heap availability (like this one with SoftReferences), you can fill up the available memory with cached data over the entire day. Of course this results in worse performance during lunch time, but the rest of the day, you can cache a lot more data, resulting in an overall better performing application. That way you are using most of the available memory resources, while still remaining at overall good quality of performance. After all, if you have a system that for 90% of the time is only in use by 50%, you are probably overpaying for resources.
The more I think about it, the more it gets clear to me, that it is probably not Ebean's responsibility to provide a good, memory and resource friendly cache implementations. As pointed out earlier, others have done so already. Nevertheless, an improvement or offer of improvement by Ebean would be great. At this point I see these options:

- Keep the current cache implementation, document its behaviour and tell everyone looking for something else, to implement ServerCacheFactory and ServerCache themselves.
- Integrate a solution like ours into the default behaviour of Ebean
- Implement and distribute an implementation of ServerCache using an appropriate cache manager (like caffeine) similar to e.g. ebean-redis for a more configurable in-memory cache solution

What do you think?

rbygrave:
Firstly, great discussion - excellent points.

For me, I think this PR is a positive so we can merge it in. As pointed out we can do better wrt more strict bounding of the cache (along the lines of every 100 puts check against max size etc) and we probably need to look at that soon'ish.

> implement ServerCacheFactory and ServerCache themselves. 

I think it will be rare that people do this themselves - way less than 1%.  That is, I think the defaults must be good - better than what we have.  As Roland pointed out, we kind of need a good implementation for the local query cache regardless due to it's nature.

> Implement and distribute an implementation of ServerCache using an appropriate cache manager (like caffeine)

Certainly an option. For me I think the first step is to probably improve the existing with more strict bounding of the cache. For example, every 100 puts check against max size (and compare/contrast with caffeine implementation for example) as adding this aspect might actually be pretty easy and get us 99% there.  Always good to have options though.

PR #2308, opened 2021-08-12
fixing test entity GenKeySequence

ThomasFOC:
Hello Rob,

to reduce the complexity of the other [pull request](https://github.com/ebean-orm/ebean/pull/2304) we cutted out this minor test fix for further discussions.
___
The test `org.tests.idkeys.TestGeneratedKeys.testSequence()` does fail with the ebean property `ebean.h2.idType=SEQUENCE`.

This is due the naming of the sequence in the migration scripts. The sequence is

```sql
create sequence SEQ_NAME increment by 1;
```
 but to read the current value of the sequence in the test the used sql is

```sql
select currval('SEQ')
```

Therefore the test fails.
___
As for our understanding `@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "SEQ_NAME")` means that there should be a `@SequenceGenerator` with the name `SEQ_NAME`. As the documentation of both annotations states, this is used to access the name of the generated sequence and override the (maybe) autogenerated name of the "persistence provider".

If such a generator is not present, Ebean itself handles the generator by naming the sequence like the generator.

Another possibility for Ebean would be to raise an exception if `generator` in `@GeneratedValue` is populated but no `@SequenceGenerator` with `generatedValue.generator().equals(sequenceGenerator.name())` is present. Users may have to change their code.

The parsing of both annotations is done in [`io.ebeaninternal.server.deploy.parse.AnnotationFields.readGenValue(GeneratedValue, Id, DeployBeanProperty)`](https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/AnnotationFields.java#L525).
___
To simply summarize this change: We wanted to fix a test and ended up in a rabbit hole :sweat_smile:


rbygrave:
Yeah, to me I think this is the incorrect fix so I need to look at it.  

That is, the static field says "SEQ" which doesn't match "SEQ_NAME" so it looks like one of those is incorrect. I don't think the best fix to add another annotation to override what looks wrong to me.  Maybe the field "SEQ" value is incorrect I'm not sure yet but adding `@SequenceGenerator` seems to me like it is masking some other mistake in the test.

rbygrave:
Said another way, this test might suggest to someone that `@SequenceGenerator` is required and that isn't the case.  There really ought to be 2 tests - one with `@SequenceGenerator` and one with out. This change effectively removes that case of using sequences without a `@SequenceGenerator` and we don't want that.

ThomasFOC:
We've added another entity `GenKeySeqB` which is annotated with both annotations. `GenKeySeqA` is the old entity.
We added some Javadoc and a second test.

PR #2309, opened 2021-08-13
FIX: mutationDetection = SOURCE works also for special json types

rPraml:
Hello Rob, I discovered a bug in the new mutationDetection.

While  
```java
@DbSource(mutationDetection = SOURCE)
List<PlainBean> plainBeanList;
```
works well, it does not work for 
```java
@DbSource(mutationDetection = SOURCE)
List<String> stringList;
```
because here Ebean uses `ScalarTypeJsonList`

I tried to write a test for all `@DbJson` annotations. The same problem would occur with `@DbArray` (which we did not use)
We currently use a lot of `@DbJson` annotations and we rely on a proper dirtyValue (oldValue<->newValue) recognition.

cheers
Roland

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonList.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonList.java
@@ -24,24 +24,24 @@ public class ScalarTypeJsonList {
   /**
    * Return the appropriate ScalarType based requested dbType and if Postgres.
    */
-  public static ScalarType<?> typeFor(boolean postgres, int dbType, DocPropertyType docType, boolean nullable) {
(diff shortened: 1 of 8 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeJsonList.java
```diff
@@ -68,14 +68,24 @@ public JsonB(DocPropertyType docType, boolean nullable) {
...
+    private final boolean keepSource;
 
-    public Base(int dbType, DocPropertyType docType, boolean nullable) {
+    public Base(int dbType, DocPropertyType docType, boolean nullable, boolean keepSource) {
       super(List.class, dbType, docType, nullable);
+      this.keepSource = keepSource;
+    }
+    
+    @Override
+    public boolean isJsonMapper() {
```
rPraml, line 79: checkme: Should we refactor that getter to `isKeepSource` `pushSource` / `popSource`

PR #2316, opened 2021-08-16
Fix: setting json values to null was throwing a NPE when persisting

jonasPoehler:
This is a regression introduced with #2309. When the value on a json property (Set or List) is set to `null` and `formatValue` is called, there was no null-check for a call to `value.isEmpty()`. My solution for this, is to delete the if-case, because `EJson.write` returns the exact same thing and it should decide on how to serialize an empty value. Also this was inconsistent to e.g. handling of a Map.


rbygrave:
Good stuff.

Issue #2233, opened 2021-05-20
Lazy loading problem 


rbygrave:
### Steps to reproduce

There was a problem with my program business. I found that the program got incorrect configuration items. It seems that because the init method of BeanList was triggered, the delayed loading changed the collection that shouldn't be changed. I don't know how to describe this problem, but I completed a test case that can reproduce this problem. Please see:

```java
===AppConfig ===start===AppConfig===
@Entity
@Cache
@Table(name = "app_config")
public class AppConfig {

@Id
@Column(name = "id")
private Integer id;

@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
@JoinColumn(name = "id", referencedColumnName = "id")
private List<AppConfigControl> items;

public Integer getId() {
return id;
}

public void setId(Integer id) {
this.id = id;
}

public List<AppConfigControl> getItems() {
return items;
}

public void setItems(List<AppConfigControl> items) {
this.items = items;
}
}
===AppConfig ===end===AppConfig===

===AppConfigControl ===start===AppConfigControl===
@Entity
@Table(name = "app_config_control")
public class AppConfigControl {

@Id
private Integer id;

private String name;

@ManyToOne
@JoinColumn(name = "config_id", referencedColumnName = "id")
private AppConfig appConfig;

public Integer getId() {
return id;
}

public void setId(Integer id) {
this.id = id;
}

public String getName() {
return name;
}

public void setName(String name) {
this.name = name;
}

public AppConfig getAppConfig() {
return appConfig;
}

public void setAppConfig(AppConfig appConfig) {
this.appConfig = appConfig;
}
}
===AppConfigControl ===end===AppConfigControl===


===Relationship ===start===Relationship===
@Entity
@Cache(enableBeanCache = true)
@Table(name="app_config_re")
public class Relationship {

@Id
private Integer id;

@ManyToOne
@JoinColumn(name="config_id",referencedColumnName="id")
private AppConfig appConfig;

public Integer getId() {
return id;
}

public void setId(Integer id) {
this.id = id;
}

public AppConfig getAppConfig() {
return appConfig;
}

public void setAppConfig(AppConfig appConfig) {
this.appConfig = appConfig;
}
}
===Relationship ===end===Relationship===

===test_case ===start===test_case===
@Test
public void testRe() {
AppConfig globalAppConfig = new AppConfig();
globalAppConfig.setId(1);
globalAppConfig.setItems(new ArrayList<>());
[...]

PR #2321, opened 2021-08-17
Improve and fix EbeanVersion checks

jonasPoehler:
Hi Rob,

this PR extends EbeanVersion to also check the required ebean-agent version against the actually running version of the agent. We acutally experienced a couple of problems with mismatched agent versions and I have also seen some concern around runtime enhancement on the forum targeting "unexplainable" errors on the forum. This version now only logs with error, if the agent version doesn't match. Since a mismatched version doesn't have to result in chaos, this seems more sensible than failing hard.

While implementing this, I realized, that the original version of EbeanVersion did not work as well, because the pom.properties file could not be retrieved. I'm unsure if this broke, when the project structure was changed, or if it never worked. For the agent version check I have now introduced a pom.properties file which with the help of maven's resource-filtering is being filled in with the correct versions and stats. For this I had to move the version properties from ebean-bom into the parent pom, so that ebean-api can also access them. Since datasource already was up there and thus duplicated, I moved them all, so that in case of an update only one line in the parent pom has to be changed.

Hope that this is a helpful improvement to Ebean
Cheers
Jonas

### DIFF
--- a/ebean-api/pom.xml
+++ b/ebean-api/pom.xml
@@ -107,6 +107,24 @@
   </dependencies>
 
   <build>
+    <resources>
+      <!-- We only want to apply filtering on the pom.properties file -->
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+        <includes>
+          <include>**/pom.properties</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>false</filtering>
+        <excludes>
+          <exclude>**/pom.properties</exclude>
+        </excludes>
+      </resource>
+    </resources>
+
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/ebean-api/src/main/java/io/ebean/EbeanVersion.java
+++ b/ebean-api/src/main/java/io/ebean/EbeanVersion.java
@@ -17,6 +17,8 @@ public class EbeanVersion {
   private static final Logger logger = LoggerFactory.getLogger(EbeanVersion.class);
 
   private static String version = "unknown";
+  private static String requiredAgentVersion = "unknown";
+  private static String currentAgentVersion = null;
 
   static {
     try {
@@ -24,11 +26,27 @@ public class EbeanVersion {
       try (InputStream in = DB.class.getResourceAsStream("/META-INF/maven/io.ebean/ebean/pom.properties")) {
         if (in != null) {
           prop.load(in);
-          in.close();
           version = prop.getProperty("version");
+          requiredAgentVersion = prop.getProperty("agent.version");
         }
       }
-      logger.info("ebean version: {}", version);
+      try {
+        Class<?> cls = Class.forName("io.ebean.enhance.Transformer");
+        currentAgentVersion = (String) cls.getMethod("getVersion").invoke(null);
+      } catch (Exception e) {
+        logger.trace("could not get run-time-agent version", e);
+      }
+      logger.info("ebean version: {} (depends on agent {})", version, requiredAgentVersion);
+      if (currentAgentVersion != null) {
+        if (!currentAgentVersion.equals(requiredAgentVersion)) {
+          logger.error("Runtime enhancement detected. Agent version mismatch");
+          logger.error("      Current:  {}", currentAgentVersion);
+          logger.error("      Required: {}", requiredAgentVersion);
+          logger.error("THIS CAN RESULT IN UNEXPECTED BEHAVIOUR!");
+        } else {
+          logger.info("Runtime enhancement detected. Agent version matches with required version.");
+        }
+      }
     } catch (IOException e) {
       logger.warn("Could not determine ebean version: {}", e.getMessage());
     }
--- a/ebean-api/src/main/resources/META-INF/maven/io.ebean/ebean/pom.properties
+++ b/ebean-api/src/main/resources/META-INF/maven/io.ebean/ebean/pom.properties
@@ -1,0 +1,4 @@
+version=${project.version}
+groupId=${project.groupId}
+artifactId=${project.artifactId}
+agent.version=${ebean-agent.version}
--- a/ebean-bom/pom.xml
+++ b/ebean-bom/pom.xml
@@ -12,16 +12,6 @@
   <artifactId>ebean-bom</artifactId>
   <packaging>pom</packaging>
 
-  <properties>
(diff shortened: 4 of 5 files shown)


rbygrave:
> We acutally experienced a couple of problems with mismatched agent versions and I have also seen some concern around runtime enhancement on the forum targeting "unexplainable" errors on the forum. 

Really? This is definitely not expected.   

> this PR extends EbeanVersion to also check the required ebean-agent version against the actually running version of the agent.

So no, the ebean-agent is expected to be backwards compatible and in fact **MUST** be backwards compatible.  If we have any problem no being backwards compatible then is needs addressing.

This PR can not be merged.

jonasPoehler:
> So no, the ebean-agent is expected to be backwards compatible and in fact **MUST** be backwards compatible. If we have any problem no being backwards compatible then is needs addressing.
I might have to kinda disagree on that. While the agent from a standpoint of functionality must be backwards compatible, bugs are not. Consider the following: We have a team of ~20 developers working with Ebean with everybody centrally set the ebean-agent-path for their JVM inside the IDE. When an Ebean update including a fix in the agent is commited onto the project, every developer has to change the agent version. Since the IDE-plugins are not very selective on whiche agent to use and there might be different versions of agent in different projects, the settings are done manually which is ok for us, since we don't update every other week with a new agent version. This usually is not a big problem, but if someone doesn't get the memo to update the version, there might be behaviour caused by a bug not fixed or missing new functionality in the agent in their IDE. These are errors, which at first glance are not easy to understand and some kind of warning in the log on startup might be helpful for that. The same usually goes for newer Java versions. The bytecode usually changes slightly from on Java version to the next and a compatibility update for the agent is being pushed. In that case you might even get class incompatibility errors while in the other cases you might just get wrong code behaviour.
I was having a bug which needed fixing in the agent (PR probably coming tomorrow) and without the fix in the agent, changes to a reference were simply not persisted. This was "only" leading to references not being saved if changed, so it is a silent error, which has to be uncovered with a test. Maybe with that PR the idea behind this one might get clearer.
I would maybe propose to make the log messages less invasive: maybe only log as warning, but IMHO there always is a good reason for a new agent version and a mismatch could be a possible reason for wrong behaviour. A warning in the log just might make troubleshooting easier for a developer. I will come back to this in a day.

jonasPoehler:
FYI: I have now opened two Pull-Requests
https://github.com/ebean-orm/ebean/pull/2325
https://github.com/ebean-orm/ebean-agent/pull/169
Here we have the case, that the test in Ebean simply fails without the fix in the agent and on first glance you don't really see the difference, when you don't think about this.

jonasPoehler:
Could we maybe come back to this? Do you still not see that this might be good addition? If not I would suggest only fixing the EbeanVersion, which currently always prints out `INFO  io.ebean - ebean version: unknown` making it useless at the moment (could as well remove it).
Another option is to reduce the log level since we are basically only pointing to a potential problem and not to a definitive one.
What is your current opinion towards this?

rPraml:
> It can't be equality - it has to be a minimum version. That is, everyone using IntelliJ with the plugin uses an agent that is arguably older (but has no effective difference). You are not experiencing that because you are not using IntelliJ right?

@jonasPoehler uses IntelliJ (probably the only one in our company) 
@rbygrave I think I understand, what you want to say. We should not enforce the latest version, but we should enforce the agent version where the last fix was done (i.e. minimum compatible version).
I checked the latest commits of ebean-agent and many of them did not add real code changes. (just version bumps - which makes the agent not incompatible to existing ebean versions)

BUT there were a few changes, where an agent update is necessary to get the proper fix.

For example https://github.com/ebean-orm/ebean/pull/2325  works only, if agent >= 12.11.2
and https://github.com/ebean-orm/ebean/issues/2233 enforces agent >= 12.8.4

Not everyone reads each line in the changelog. If #2325 is mentioned in the changelog of ebean 12.11.2, you may rely, that it is fixed (but it is only, if you also update the agent to 12.11.2)

IMHO it is definitively an improvement, if we warn the developer, if an incorrect/incompatible agent is used. To compare the versions we could use the  MigrationVersion.
```java
// this property must be set to the minimum agent version, which is still compatible with ebean
// List of fixes, that reqire a certain ebean version
// fix #2233 enforces 12.8.4
// fix #2325 enforces 12.11.2
private static MigrationVersion MINIMUM_AGENT_VERSION = MigrationVersion.parse("12.11.2");
....
if (MigrationVersion.parse(currentAgentVersion).compareTo(MINIMUM_AGENT_VERSION)) {
   // print out a warning, that ebean is too new for this agent. 
}
```
I even thought about saving the agent version in the enhanced classes. (Similar to _EBEAN_MARKER)
If you have a precompiled jar, let's say enhanced with an old ebean-agent, that is incompatible with ebean 12, ebean coud detect this on start-up that certain beans are enhanced with the wrong agent. 
Of course this can only work for future agent versions.

cheers
Roland

rbygrave:
> but we should enforce the agent version where the last fix was done (i.e. minimum compatible version).

Yes, minimum compatible version.

> Not everyone reads each line in the changelog.

Yes, I suspect most people only read the changelog after the fact is something doesn't work.

> IMHO it is definitively an improvement, if we warn the developer, if an incorrect/incompatible agent is used. 

Agreed. It is a matter of determining the way to do that.

My initial thought was to have a plain int to represent the version in a text file resource as part of ebean-agent (say META-INF/ebean-agent-effective-version.txt). This would need to be manually updated when the ebean-agent makes a bug fixing change. I'll call this the "effective ebean-agent version". [NB: This resource needs to be open in module-path]

e.g.  say we had version `12.3.4`  that might be `120304`  (so padding is used for minor and patch versions - 0 padding to 100).

Then Ebean 12.3.4 and onwards would _require a minimum_ "effective ebean-agent version" of 120304.

If we then get a ebean-agent bug at 12.7.2 we manually bump the "effective ebean-agent version" to 120702. Ebean 12.7.2 and onwards would then  _require a minimum_ "effective ebean-agent version" of 120702.

That is the thought I had.

rbygrave:
> To compare the versions we could use the MigrationVersion.

I don't think so. I don't think we want ebean-core to have a hard dependency on ebean-migration just in order to do this.

rbygrave:
Forget all that about a ebean-agent-effective-version.txt ... replacing with a PR that just checked semantic version of ebean-agent major and minor versions.

https://github.com/ebean-orm/ebean/pull/2402

rbygrave:
I think this PR can be closed now as it has been superseded by #2402

rbygrave:
I'll close this.

PR #2329, opened 2021-08-24
@OrderdColum did not work properly with cached beans

jonasPoehler:
This PR fixes a problem, where a bean with a OneToMany relation with @OrderColumn retrieved from cache didn't have a sortOrder set and when the list was modified, the change was not detected and the saving of the bean skipped. With the provided test, we can see, that beans from cache have to behave the same as ones without cache.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
@@ -294,8 +294,13 @@ final class BeanDescriptorCacheHelp<T> {
 
     List<Object> idList = entry.getIdList();
     bc.checkEmptyLazyLoad();
+    int i = 0;
     for (Object id : idList) {
-      many.add(bc, targetDescriptor.createReference(readOnly, id, persistenceContext));
+      final EntityBean ref = targetDescriptor.createReference(readOnly, id, persistenceContext);
+      if (many.hasOrderColumn()) {
+        ref._ebean_getIntercept().setSortOrder(++i);
+      }
+      many.add(bc, ref);
     }
     return true;
   }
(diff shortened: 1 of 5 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
```diff
@@ -294,8 +294,13 @@ boolean manyPropLoad(BeanPropertyAssocMany<?> many, BeanCollection<?> bc, Object
 
     List<Object> idList = entry.getIdList();
     bc.checkEmptyLazyLoad();
+    int i = 0;
     for (Object id : idList) {
-      many.add(bc, targetDescriptor.createReference(readOnly, id, persistenceContext));
+      final EntityBean ref = targetDescriptor.createReference(readOnly, id, persistenceContext);
+      if (many.hasOrderColumn()) {
+        ref._ebean_getIntercept().setSortOrder(++i);
```
rbygrave, line 301: Ha, are we 1 based on sortOrder - interesting. Note to just double check this afterwards.

rbygrave:
Excellent work @jonasPoehler  !!

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorCacheHelp.java
jonasPoehler: Yeah we are! Confused me a bit when we started to use @OrderColumn, but was also more helpful in other places. Note: The official [documentation](https://docs.jboss.org/hibernate/jpa/2.2/api/javax/persistence/OrderColumn.html) states that the first element starts with 0 so this might not be what we want for people adopting Ebean and moving from another JPA implementation.

PR #2331, opened 2021-08-24
Fix for NPE in mutableNext in EBI

jonasPoehler:
This change fixes an error, where the mutableValueNext was null and thus a NPE occured on call to `content()`, when two Json properties were changed on a bean with one empty and the other `null`.


rbygrave:
Nice !!

PR #2332, opened 2021-08-27
Improve MAVEN build Performance

ChenZhangg:
[Parallel builds in Maven 3](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3) Maven 3.x has the capability to perform parallel builds.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ env:
 install: true
 
 script:
-  - mvn clean test
+  - mvn -T 1C clean test
 
 after_failure:
   - ./.travis/print_surefire_reports


rPraml:
Note, that the [ebean-maven-plugin](https://github.com/ebean-orm-tools/ebean-maven-plugin) is not recognized to be threadSafe:
```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in ebean core:
[WARNING] io.ebean:ebean-maven-plugin:12.11.2
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
```
I think it should be no problem to run parallel builds, but maven will print the warning above, until the plugin will be marked `@threadSafe`.

PR #2337, opened 2021-08-31
Query bean PByteArray type for byte[]

Ryszard-Trojnacki:
Added query beans mapping for column type byte[].


rbygrave:
Excellent work, well done.

PR #2339, opened 2021-09-01
no effective code change: Add missing @Override annotation 

rPraml:
also removed duplicate methods in interfaces


ebean-core/src/main/java/io/ebeaninternal/api/SpiDtoQuery.java
```diff
@@ -36,11 +36,6 @@
    */
   boolean isRelaxedMode();
 
-  /**
-   * Return the label for the query.
-   */
-  String getLabel();
```
rPraml, line 42: method already definde by super interface

PR #2371, opened 2021-09-15
Allow formula defined properties to be used in where and orderBy

jonasPoehler:
Hello Rob,

this change allows a better use of formula properties in where- or orderBy-clauses. Currently this would only work, if the Formula-annotation would either not contain a join-clause, or the join-clause is not using table-aliases to define a join-alias used in the select. With this change the join and the select are better tracked together so that filters and sort-order can work properly.
This also adds a couple of tests. I hope this is the correct fix, otherwise I'm looking forward to feedback.

Cheers
Jonas


rbygrave:
Dang, impressive PR !!  

> I hope this is the correct fix

Yes, this looks good to me. I don't see a simpler option - very well done !!

PR #2372, opened 2021-09-15
Undo generated properties on OptimisticLockException

jonasPoehler:
This PR makes sure, that a generated property like a @Version-property gets reset when an OptimisticLockException is thrown. This is a particular problem in the following case:

1. A loads bean with version 1
2. B loads bean with version 1
3. A updates bean with version 1 (gets incremented to 2)
4. B tries to update bean with version 1, but gets OptimisticLockException (version on bean in memory stays incremented, now 2 -- wrong)
5. B now can save again, since the version matches now, and A's changes are overwritten.

With this change the version would be rolled back in step 4, making sure, that simply saving a second time doesn't change the behaviour.


rbygrave:
Nice one, I like it !!

PR #2373, opened 2021-09-15
Allow post-processing in BeanFindController

jonasPoehler:
Hi Rob,

With this change we would like to introduce post-processing of beans in the BeanFindController. I tried to show our use-case as best as I can with the provided testcase and especially the TestBeanFindController. We have a bean "softly referencing" different beans with their own respective tables. On load of the beans we want to fetch the referenced beans and store them in a transient property and that is what I'm trying to show with the test. We cannot load the referenced beans in their individual getters, because that way we are unable to fetch multiple beans in one go opposed to the new `postProcessMany` method.
I hope that I was able to bring our intent across and that you would want to include this extension. Nevertheless I'm unsure, if maybe there would be a better way of doing this, so I'm open to your suggestions.

All the best
Jonas


rbygrave:
>  include this extension

Yes, looks good and not adding too much complexity.

PR #2378, opened 2021-09-17
#2377 - Wrong order in batch save - Fix to transaction depth

rbygrave:
This extends the fix and tests in PR https://github.com/ebean-orm/ebean/pull/2377 ... with an additional fix to transaction depth.

The guts of this fix is in DefaultPersister.saveAssocOne() with the change to use the new transaction.depthDecrement() method. For our failing test cases this changes the depth from -1, 0 to be 0, 1.  

Our failing tests failed as **-1** will always order before **0**.  The tests got the -1 ordering via the `saveAssocOne()`, so our fix effectively shifts the depth computation by +1.

We can see the impact of this change in the batch ordering via the logging of:
 io.ebean.SUM - txn[1001] BatchControl flush ...

With this additional change testMultiSave4 looks like

```
BatchControl flush [Relation3:0 i:1, Relation2:1 i:1, Relation1:100 i:1]
```

... the resulting ordering is 0, 1 and 100.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
@@ -29,6 +29,8 @@ import java.util.List;
  */
 public final class BatchControl {
 
+  private static final Object DUMMY = new Object();
+
   /**
    * Used to sort queue entries by depth.
    */
@@ -50,7 +52,7 @@ public final class BatchControl {
    * Set of beans in this batch. This is used to ensure that a single bean instance is not included
    * in the batch twice (two separate insert requests etc).
    */
-  private final IdentityHashMap<Object, String> persistedBeans = new IdentityHashMap<>();
+  private final IdentityHashMap<Object, Object> persistedBeans = new IdentityHashMap<>();
 
   /**
    * Helper to determine statement ordering based on depth (and type).
@@ -179,22 +181,15 @@ public final class BatchControl {
    * Add the request to the batch and return true if we should flush.
    */
   private boolean addToBatch(PersistRequestBean<?> request) {
-    int depth = transaction.depth();
-    BeanDescriptor<?> desc = request.descriptor();
-    // batching by bean type AND depth
-    String key = desc.rootName() + ":" + depth;
-    
-    String alreadyInBatch = persistedBeans.put(request.entityBean(), key);
+    Object alreadyInBatch = persistedBeans.put(request.entityBean(), DUMMY);
     if (alreadyInBatch != null) {
       // special case where the same bean instance has already been
       // added to the batch (doesn't really occur with non-batching
       // as the bean gets changed from dirty to loaded earlier)
-      BatchedBeanHolder beanHolder = getBeanHolder(request, alreadyInBatch);
-      int ordering = depthOrder.orderingFor(depth);
-      return beanHolder.getOrder() > ordering;
+      return false;
     }
 
-    BatchedBeanHolder beanHolder = getBeanHolder(request, key);
+    BatchedBeanHolder beanHolder = getBeanHolder(request);
     int bufferSize = beanHolder.append(request);
 
     bufferMax = Math.max(bufferMax, bufferSize);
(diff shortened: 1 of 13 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -1123,7 +1125,7 @@ private void saveAssocOne(PersistRequestBean<?> request) {
           && !prop.isReference(detailBean)
           && !request.isParent(detailBean)) {
           SpiTransaction t = request.transaction();
-          t.depth(-1);
+          t.depthDecrement();
```
rbygrave, line 1128: The magic here - effectively we only decrement the depth when it is already greater than 0. This has the impact that it now will not make the depth negative (and hence order it before other persists requests at depth zero).

When we want to do this we now need to ensure we reset the depth back to zero after top level inserts and updates occur.

ebean-core/src/main/java/io/ebeaninternal/server/transaction/JdbcTransaction.java
```diff
@@ -484,34 +484,28 @@ public final boolean isSaveAssocManyIntersection(String intersectionTable, Strin
...
-  /**
-   * Return the current depth.
-   */
   @Override
   public final int depth() {
     return depth;
   }
 
+  @Override
+  public final void depthDecrement() {
```
rbygrave, line 498: this effectively means it won't take the depth negative.  So relative to other independent persist requests also at depth zero will execute in the order that they are persisted.  

Fixes testMultiSave4() and testMultiSave4_usingRelation4()

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -417,7 +418,7 @@ public void update(EntityBean entityBean, Transaction t) {
       } else {
         update(req);
       }
-
+      req.resetDepth();
```
rbygrave, line 421: Reset the depth after each top level insert and update.  This means that all our top level requests at depth zero are treated in the order that they are saved().

rbygrave:
FYI @rPraml

rbygrave:
With this additional change testMultiSave4 looks like

```console
BatchControl flush [Relation3:0 i:1, Relation2:1 i:1, Relation1:100 i:1]
```

ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
```diff
@@ -181,15 +179,22 @@ public int executeOrQueue(PersistRequestBean<?> request, boolean batch) throws B
...
+    
+    String alreadyInBatch = persistedBeans.put(request.entityBean(), key);
     if (alreadyInBatch != null) {
       // special case where the same bean instance has already been
       // added to the batch (doesn't really occur with non-batching
       // as the bean gets changed from dirty to loaded earlier)
-      return false;
+      BatchedBeanHolder beanHolder = getBeanHolder(request, alreadyInBatch);
+      int ordering = depthOrder.orderingFor(depth);
+      return beanHolder.getOrder() > ordering;
```
rPraml, line 194: Can you check, if this really makes sense what I've done here ;)

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
rPraml: Q: Is it OK to increment the depth two lines below, if we did not decrement it?
```java
int decremented = t.depthDecrement(); // returns 0 or 1
saveRecurse(...)
t.depth(decremented )
```
May also other places where depth is modified be affected by this kind of bug (e.g. `deleteAssocMany`, `deleteList` ...)
t.depth(decremented);

rbygrave: > Q: Is it OK to increment the depth two lines below, if we did not decrement it?

Yes I think so.  I actually started off with code like what you have (I called decremented delta) but actually that just left things at depth 0 and was ok but I felt it was better to instead ~ "move everything down 1" in terms of depth.

>  deleteAssocMany, deleteList ...

I am not sure there is the same ordering issue with deletes yet. I'd prefer to have a failing test first.

ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
rbygrave: Yes, I think it looks ok but I probably need to put a break point here and see some examples so I'll do that.

rbygrave: Ah so yes it turns out we can revert this back (we don't need this change) now that we have the other change to "fix the depth" per say.  So I'll revert this part of the change back.

PR #2402, opened 2021-09-28
Modify EbeanVersion to check for minimum ebean-agent version

rbygrave:
In EbeanVersion we manually maintain a minimum version of ebean-agent that is expected.  Log an error when an older ebean-agent is being used.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/EbeanVersion.java
+++ b/ebean-api/src/main/java/io/ebean/EbeanVersion.java
@@ -16,21 +16,71 @@ public class EbeanVersion {
 
   private static final Logger log = LoggerFactory.getLogger("io.ebean");
 
+  /**
+   * Maintain the minimum ebean-agent version manually based on required ebean-agent bug fixes.
+   */
+  private static final int MIN_AGENT_MAJOR_VERSION = 12;
+  private static final int MIN_AGENT_MINOR_VERSION = 12;
   private static String version = "unknown";
   static {
+    readVersion();
+    checkAgentVersion();
+  }
+
+  private static void readVersion() {
     try {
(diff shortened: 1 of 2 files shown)


jonasPoehler:
I think that this is good solution for the discussion we were having in #2321. Maybe just two things that came to my mind:
- I'm not sure, that the test will last long, because it will fail as soon as the next "breaking" agent-update comes along. A possible solution could be allowing instantiation of EbeanVersion with major- and minor-version as parameters in order to be able to "mock" these version numbers in tests. That way the static methods could delegate to a private EbeanVersion instance while we could test the basic behaviour
- Maybe add a comment in the pom.xml at the position of the agent-/maven-plugin-version regarding the EbeanVersion class, so that nobody could "forget" to update the required versions when e.g. updating the agent in a PR.

rbygrave:
> will fail as soon as the next "breaking" agent-update comes along

Well, there isn't a short-cut in that it's manual knowledge when ebean-agent has an "interesting" or "breaking" or "required" change. When that occurs **I must manually** update the minimum required version here in EbeanVersion.  As such those tests should break and I must also update the tests - this is all good and expected (and yes it relies on knowledge of ebean-agent).

That all said, we have very few changes to ebean-agent, so I don't expect this to be a drama at all.  Also note that we always release ebean-agent + the various plugins in advance of releasing ebean so that also isn't a problem.
  
> so that nobody could "forget" to update the required versions

Yeah, lets see. I'm somewhat "hyper aware" of ebean-agent changes but I suspect this should be top of ebean-agent README to others to note.

rbygrave:
New IntelliJ plugin released so merging this now.

Issue #1802, opened 2019-08-23
Change MySQL DDL to not use helper stored procedures by default - was auto-generated script containing create procedures fails


cyrilfr:
I experience [this issue](https://github.com/playframework/play-ebean/issues/166) with the Ebean SBT plugin while running the auto-generated script.

The script is actually produced by the plugin calling this method:

    public static String generateEvolutionScript(EbeanServer server) {
        try {
            SpiEbeanServer spiServer = (SpiEbeanServer) server;
            CurrentModel ddl = new CurrentModel(spiServer);

            String ups = ddl.getCreateDdl();
            String downs = ddl.getDropAllDdl();

            if (ups == null || ups.trim().isEmpty()) {
                return null;
            }

            return
                "# --- Created by Ebean DDL\r\n" +
                "# To stop Ebean DDL generation, remove this comment and start using Evolutions\r\n" +
                "\r\n" +
                "# --- !Ups\r\n" +
                "\r\n" +
                ups +
                "\r\n" +
                "# --- !Downs\r\n" +
                "\r\n" +
                downs;
        } catch (IOException ex) {
            throw new RuntimeException(ex);
        }
    }

from the [EbeanDynamicEvolutions.java](https://github.com/playframework/play-ebean/blob/master/play-ebean/src/main/java/play/db/ebean/EbeanDynamicEvolutions.java) file.

There is a syntax error when I try to run the SQL script manually:

> Syntax error near 'delimiter' at line 3

rbygrave:
Was this not fixed by https://github.com/ebean-orm/ebean-migration/issues/69 ?

rbygrave:
@almothafar Can you add a comment please?

almothafar:
What is your DBMS @cyrilfr? 
For me, I worked with PG and MSSQL and workarounds working fine for me.

Workarounds mentioned already in https://github.com/playframework/play-ebean/issues/166 

If you are using SQL Server, with the versions of 11.41.4 or later for ebean-orm and 11.19.1 or later for ebean-migration (and BTW play-ebean does not using these versions, it is sit on older versions), you should have GO and no delimiters! and still, you need to do manual workarounds or run the script manually.

cyrilfr:
@rbygrave I've this issue with the last version of the plugin and the last version of ebean.
@almothafar I've tried with MySQL 5.7.27. phpMyAdmin detect the syntax error. 
[...]

PR #2403, opened 2021-09-29
Reintroduce stored procedures to drop columns on MySql

jonasPoehler:
Hello Rob,

this PR would re-introduce using stored procedures to drop columns in migration scripts on MySql database platforms as originally introduced in #1451. With #1802 the original code was reverted for MySql, because some migration runners other than Ebean's could not interpret the init-script. I have modified the orignal change so that Ebean's default behaviour stays as it is (generating simple `alter table ... drop ...` calls) but now there is an option available to generate the init-script as well as use the stored procedure to drop the columns while before dropping all constraints.
I guess that for quite a lot of people using MySql (or MariaDB) with Ebean and its migration runner, this would be a major improvement and less hassle when dropping columns.
Additionally I added a test, which checks for the correct procedure calls in generated SQL and I didn't have to change anything for the other migration tests.
If this would get merged, I would also like to provide documentation for the website, so that there is a place where you can actually stumble upon this feature or reference to in case of a question e.g. on the forum. I already sketched out a paragraph for the documentation [here](https://github.com/ebean-orm/website-source/compare/master...FOCONIS:update_migration_docs#diff-cec9554fc6df8b1b53fe7e2f21d926473db3c8db38c41f9c21123112bb6ebe77R312), but I would open a PR once this one is finalized.

Looking forward to your feedback and all the best!
Jonas

### DIFF
--- a/ebean-api/src/main/java/io/ebean/config/PlatformConfig.java
+++ b/ebean-api/src/main/java/io/ebean/config/PlatformConfig.java
@@ -66,6 +66,8 @@ public class PlatformConfig {
 
   private boolean caseSensitiveCollation = true;
 
+  private boolean useMigrationStoredProcedures = false;
+
   /**
    * Modify the default mapping of standard types such as default precision for DECIMAL etc.
    */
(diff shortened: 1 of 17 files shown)


jonasPoehler:
Thank you @rbygrave for taking time to review and merge this. The changes look fine to me, I guess some of the problems were relicts from the previous implementation and so it was good to have them cleaned up 👍🏿.

rbygrave:
Yes, apologies that it took so long to get to that.  I had a review partially done but not submitted for a few months there - oh well, sorted now :)

Issue #2021, opened 2020-06-17
Use smarter persistence context for findEach/iterate/stream queries - improve performance of un-tuned findEach/iterate/stream queries that invoke lazy loading


rbygrave:
These changes should improve the performance of untuned findEach/iterate/stream queries that invoke lazy loading.

### Changes:
- Enable use existing persistence context (e.g. beans loaded by prior queries in the same transaction)

- Use a smarter persistence context with a "per 1000 beans added" reset limit. This improves our use of batch lazy loading during iteration 

### Current limitations
The current implementation of has a few limitations which this change addresses. The findEach etc queries create their own persistence context effectively per "top level bean". The short lived persistence contexts allows these queries to support iterating of many thousands or millions of rows without having to how all the loaded beans in memory (ie. as we no longer have all beans referenced in a single persistence context).

PR #2407, opened 2021-10-07
Change the PersistenceContext default resetLimit for iteration queries to 100 (from 1000)

rbygrave:
FYI @rPraml 

Changes the resetLimit from 1000 to 100 for iteration queries (findEach, findIterate, findStream).

Then with `TestPersistenceContext.findWithGcTest()` we see

```
Total instances: 500, instances left in memory: 99
Total instances: 1000, instances left in memory: 99
Total instances: 1500, instances left in memory: 99
Total instances: 2000, instances left in memory: 99
Total instances: 2500, instances left in memory: 99
Total instances: 3000, instances left in memory: 99
Total instances: 3500, instances left in memory: 99
Total instances: 4000, instances left in memory: 99
Total instances: 4500, instances left in memory: 99
Total instances: 5000, instances left in memory: 99
Total instances: 5500, instances left in memory: 99
Total instances: 6000, instances left in memory: 99
Total instances: 6500, instances left in memory: 99
Total instances: 7000, instances left in memory: 99
Total instances: 7500, instances left in memory: 99
Total instances: 8000, instances left in memory: 99
Total instances: 8500, instances left in memory: 99
Total instances: 9000, instances left in memory: 99
Total instances: 9500, instances left in memory: 99
Total instances: 10000, instances left in memory: 99
```

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
@@ -273,10 +273,10 @@ public final class DefaultPersistenceContext implements SpiPersistenceContext {
     }
 
     /**
-     * Return true if grown above the reset limit size of 1000.
+     * Return true if grown above the reset limit size of 100.
      */
     private boolean resetLimit() {
-      return map.size() > 1000;
+      return map.size() >= 100;
     }
 
     @Override
--- a/ebean-test/src/test/java/io/ebeaninternal/server/transaction/DefaultPersistenceContextTest.java
+++ b/ebean-test/src/test/java/io/ebeaninternal/server/transaction/DefaultPersistenceContextTest.java
@@ -199,23 +199,23 @@ public class DefaultPersistenceContextTest {
     final PersistenceContext pcIterate = initialPc.forIterate();
     assertFalse(pcIterate.resetLimit());
 
-    // added 900 NEW contact beans
-    addContacts(pcIterate, 2000, 900);
-    assertThat(pcIterate.size(Contact.class)).isEqualTo(1910);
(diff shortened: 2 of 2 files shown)


rPraml:
Hello Rob, thanks for the fix. The memory leak fix seems to be easier than I thought.
Changing the value to 100 suits us.

But why is this done at all? As far as I understand, you put a lot of effort in findIterate (#2021 and so on) so that it periodically starts a new persistence context.

What do you generally think if we have only one persistence context, that uses weak references, so that it has not the ability to hold beans in memory.

IMHO that would make sense, that the PC holds only the beans in memory, that are strong referenced somewhere else. It would also free up memory if you clear lists (e.g. customer.getOrders().clear() will free all orders in the PC)

I made a proof of concept in this branch https://github.com/FOCONIS/ebean/commits/experimental/use-weak-ref

I do not know every detail (e.g. if PC is used for prefetch/preload and so on) - I also know, that using references may add additional costs (runtime/memory) - but on the other hand, the change will remove complexity and the PC will cache as many beans as the garbage collection allows.

If you think, WeakReferences are the wrong way here, you may take a look at that change https://github.com/FOCONIS/ebean/commit/060c385a8884a147458b5343fa4a603132fdcae0 it will remove code used only in one test.

Cheers
Roland

rbygrave:
> But why is this done at all? 

Largely because it means people can then use findEach(), findStream() etc even for small results - choose it for the programming model/style rather than "large resultSet" reasons. In doing so they can get the same behaviour wrt persistence context as if they used findList.  

That is, people can use findEach(), findStream() without thinking or understanding much and the previous persistence context per top level bean isn't really appropriate/good for that case.  I felt we need the default behaviour for small results via findEach/findStream to be just as good as findList.

rbygrave:
Discard this now.

PR #2411, opened 2021-10-11
Weak reference based persistence context for streaming queries

rbygrave:
The concept is to use this for streaming queries rather than periodically creating a new persistence context


rbygrave:
Thanks @rPraml , I think  we have ended up with a good result here.

PR #2438, opened 2021-11-16
FIX: typeGetAll did not work with repeated annotations on one class

NSzemenyei:
Hello  @rbygrave ,

@jonasPoehler and I wrote a test and made a fix for typeGetAll() with repeated annotations.

Kind regards,
Noemi


rbygrave:
Nice work.

PR #2440, opened 2021-11-16
FIX: Json deserialize references

NSzemenyei:
Hello @rbygrave ,

@jonasPoehler and I found a bug in deserializing of references.
We also have a proposal for the solution.

Kind regards
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
@@ -3038,6 +3038,10 @@ public class BeanDescriptor<T> implements BeanType<T>, STreeType, SpiBeanType {
     return ebi.isReference() || referenceIdPropertyOnly(ebi);
   }
 
+  boolean hasIdPropertyOnly(EntityBeanIntercept ebi) {
+    return propertiesBaseScalar.length > 0 && ebi.hasIdOnly(idPropertyIndex);
+  }
+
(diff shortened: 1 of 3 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptor.java
```diff
@@ -3038,6 +3038,10 @@ public boolean isReference(EntityBeanIntercept ebi) {
     return ebi.isReference() || referenceIdPropertyOnly(ebi);
   }
 
+  boolean hasIdPropertyOnly(EntityBeanIntercept ebi) {
```
rbygrave, line 3041: This method should not be required and instead use the existing `referenceIdPropertyOnly()`.  That takes into account generated properties - idOnlyReference true when there is other non-generated scalar properties on the bean.

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanDescriptorJsonHelp.java
```diff
@@ -152,6 +152,9 @@ private T jsonReadProperties(SpiJsonReader readJson, EntityBean bean, String pat
     }
     if (contextBean == null) {
       readJson.beanVisitor(bean, unmappedProperties);
+      if (!isNullOrZero(id)) {
+        desc.setReferenceIfIdOnly(bean._ebean_getIntercept());
```
rbygrave, line 156: I think this should also happen before beanVistor() ... just after read.

Issue #1838, opened 2019-10-10
Fix for Joda LocalTime timezone handling for MySql


rbygrave:
MySql only issue with timezone handling for LocalTime (which doesn't have timezone).

PR #2443, opened 2021-11-25
Big refactor and bug fixing of date time handling

rPraml:
Hello Rob, this is a very big PR and @NSzemenyei and I tried to split it up in several commits.

It adressess the problem of date time handling in ebean and I found several bugs or detected inconsistencies
I think not all in this PR is correct, so feel free to give me feedback
(I recommend watching this [video ](https://youtu.be/-5wpm-gesOY) to relax before your head explodes)

I imagine the following scenario:
- distributed database around the world (one replica in germany, an other in USA)
- multiple application servers around the world
all of them are in different local timezones

*Assumptions*
1. A time only value (LocalTime / java.sql.Time) is never converted. So if `05:15:15` is stored in the database, you will get this value in USA and in germany
2. A time only value is serialized as ISO string (like `05:15:00`) or as GMT based millis value in JSON. So `00:00:00` is equivalent to 0 ms
3. A date only value (LocalDate / java.sql.Date) is never converted. So if `25.11.2021` is stored in the database, you will get this value in USA and in germany
4. A date only value is serialized as ISO string (like `2021-11-25`) or as GMT based millis value in JSON. So `1970-01-01` is equivalent to 0 ms
5. A LocalDateTime is never converted ( see 1 & 3)
6. A LocalDateTime is serialized as ISO String (see 2 & 4)
7. A Timestamp/Instant/ZonedDateTime is stored as relative offset to your current data timezone - so if you set your dataTimeZone to GMT - the offset is 0. It's up to the user (e.g. for Instant) or to the datatype (e.g. for Calendar) to convert the value to your local jvm timezone (e.g. Calendar.toString will return different values in USA and Germany, while Instant.toString will return the same)
8. A Timestamp/Instant/ZonedDateTime/Calendar is serialized as ISO String (Instant.toString() / Instant.parse) and as GMT based millis value in JSON. The value 0 ms represents the 1970-01-01 00:00 GMT timestamp. 

We have written a test an extensive [DateAndTimesTest](https://github.com/ebean-orm/ebean/commit/2f853c11df1c5a9b88cd9477fd39ef984f5045ff#diff-9e37d94c940992598d4a8374c37f5f06c6c27f21b7208780f4b6bce85f86e125) that tries to check the following things for all date/time datatypes:
- how does it read a value stored in the database
- can I find a model with `where().eq("someTimeProperty", timeValue)
- Is the scalarType.convertToMillis/convertFromMillis, toInstant/fromInstant and so on symmetric
- Is the JSON roundtrip symmetric

Especially in the JSON roundtrip, we found a lot of bugs. This seems to be untested for the different JSON variations like ISO/Millis/Nanos.

We also tested different database engines and found annomalities in the driver. It was a nightmare as you've already mentioned here: https://jira.mariadb.org/browse/CONJ-433

All in all, we have a bunch of fixes, which work on most database systems now.
I am also aware, that this type of change may introduce breaking changes. So we tried to split up into some non dangerous commits:

https://github.com/ebean-orm/ebean/commit/8945833d912502a2a54ecbd62b5c28e23e5f43c2 (non danger)
This commit fixes a bug when the test tries to find the bean with `where().eq("calendar", value)` because calendar was not a `java.util.Calendar` but a subtype (GregorianCalendar). If a ScalarType was not found, all scalar types have to be checked, if one of them is able to handle this datatype. I'm aware, that this check may be improved, so that we do not get collisions.

https://github.com/ebean-orm/ebean/commit/5e700fa2f9701f77a387ae88e26cc637241b012a, https://github.com/ebean-orm/ebean/commit/4ff3c7868d6311f61fb95d07fbd79b6605e4b822, https://github.com/ebean-orm/ebean/commit/4fcd3be6c5cbf76ad309ae2efc2cb336a2a44d3c https://github.com/ebean-orm/ebean/commit/1b5ce51ab165374334f890b76931e8ba2b1d0187 https://github.com/ebean-orm/ebean/commit/1d9ba5b4b40541f95ee9f2516799d4966325f25f (non danger)
Fixes ISO-Format of some scalar types

https://github.com/ebean-orm/ebean/commit/00a7a0d88d136f88aaf7058ddfd75fb056ed2f85 (danger)
This introduces a flag in DataBinder and RsetDataReader, that controls, if the read type is a localType or not.
Note, that we tried to disable conversion for date only and time only values (see assumption 1 & 3)

https://github.com/ebean-orm/ebean/commit/ce706ad36220f179f853edc2ca669fd0b83a4c6f (mid danger)
DataTypes from the java.time package should not use `IsoJsonDateTimeParser.formatIso` as it will truncate nanoseconds

https://github.com/ebean-orm/ebean/commit/45b90dc1345336dc9b211219b4556a82d87d2511 (mid danger)
This fixes some conversion problems in json datatypes and it addresses a **severe bug** : We cannot handle dates before 1970-01-01 in Json-Mode NANOS

https://github.com/ebean-orm/ebean/commit/540f15ac51e0300872e25d9908d408f5538d78a5 (danger)
This will change the millis conversion of localDate to a GMT based value, so 1970-01-01 is alway 0, no matter in which timezone we are. See also https://github.com/ebean-orm/ebean/commit/b4899fb9a8274db57d5f21834a93dde8539a64ff and https://github.com/ebean-orm/ebean/commit/a3b6090def2867560624a02ea807aa1245ca297a for joda

https://github.com/ebean-orm/ebean/commit/168709119308c09d95c3d44fef2d6fb200e627bf (breaking change, but bug)
Timestamp and Util.date had returned millis value in NANO Json mode

https://github.com/ebean-orm/ebean/commit/9ab33427e4ffe64ed76d2c0b4fe85d223171fb2d (bug)
Problem with negative dates (< 1970-01-01) fixed

https://github.com/ebean-orm/ebean/commit/98ae7d43a5148b726d3afee9fec6fea6f35e29c0 (danger)
Fixes for MariaDB

https://github.com/ebean-orm/ebean/commit/25ad2426ac7222997e5c37580acdd6b03700b28b (java 8 incompat)
Returned to jdk8 compatible oracle driver. Unfortunately, the drivermanager stops loading drivers, if there is one that is in wrong class format. This prevents drivers, that are in order after the oracle driver from being loaded.

https://github.com/ebean-orm/ebean/commit/ea415e1f75041882c1dd437202bf9466393c3325 (addon)
add starter for the new DB2 container

https://github.com/ebean-orm/ebean/commit/4369de4a2393edf093b808d58e6d92a5896e5041 (bug)
Fix a test, introduced trough the changes. 

We tried to work here test driven and wrote a test, that verifies the assumptions.

Maybe you can take a look at this big PR and cherry pick the one or other uncritical change.
Feel also free to call the assumptions into question.

cheers
Noemi & Roland

### DIFF
--- a/ebean-core-type/src/main/java/io/ebean/core/type/DataBinder.java
+++ b/ebean-core-type/src/main/java/io/ebean/core/type/DataBinder.java
@@ -117,7 +117,7 @@ public interface DataBinder {
   /**
    * Bind a timestamp value.
    */
-  void setTimestamp(Timestamp value) throws SQLException;
+  void setTimestamp(Timestamp value, boolean isLocal) throws SQLException;
 
   /**
    * Bind a time value.
--- a/ebean-core-type/src/main/java/io/ebean/core/type/DataReader.java
+++ b/ebean-core-type/src/main/java/io/ebean/core/type/DataReader.java
@@ -39,7 +39,7 @@ public interface DataReader {
 
   java.sql.Time getTime() throws SQLException;
 
-  java.sql.Timestamp getTimestamp() throws SQLException;
+  java.sql.Timestamp getTimestamp(boolean isLocal) throws SQLException;
 
   BigDecimal getBigDecimal() throws SQLException;
 
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/InternalConfiguration.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/InternalConfiguration.java
@@ -451,7 +451,9 @@ public final class InternalConfiguration {
       }
       return new NoDataTimeZone();
     }
-    if (getPlatform().base() == Platform.ORACLE) {
+    if (isMySql(getPlatform())) {
+      return new MySqlDataTimeZone(tz);
+    } else if (getPlatform().base() == Platform.ORACLE) {
       return new OracleDataTimeZone(tz);
     } else {
       return new SimpleDataTimeZone(tz);
@@ -459,7 +461,7 @@ public final class InternalConfiguration {
   }
 
   private boolean isMySql(Platform platform) {
-    return platform.base() == Platform.MYSQL;
+    return platform.base() == Platform.MYSQL || platform.base() == Platform.MARIADB ;
   }
 
   public DataTimeZone getDataTimeZone() {
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/DataTimeZone.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/DataTimeZone.java
@@ -8,16 +8,18 @@ import java.util.Calendar;
 public interface DataTimeZone {
 
   /**
-   * Return the Calendar to use for Timezone information when reading/writing timestamps.
+   * Return the Calendar to use for Timezone information when reading/writing
+   * timestamps/Instants
    */
   Calendar getTimeZone();
 
   /**
-   * Return the Calendar to use for Timezone information when reading/writing date.
-   * A 'date' only value has normally no timezone information, but some platforms (like MySQL)
-   * reqire this.
+   * Return the Calendar to use for Timezone information when reading/writing
+   * LocalDate / LocalTime / LocalDateTime. A local date only value has normally
+   * no timezone information - and no conversion will occur. but some platforms
+   * (like MySQL) require this.
    */
-  default Calendar getDateTimeZone() {
+  default Calendar getLocalTimeZone() {
     return null;
   }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/MySqlDataTimeZone.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/MySqlDataTimeZone.java
@@ -19,7 +19,7 @@ public class MySqlDataTimeZone implements DataTimeZone {
   }
 
   @Override
-  public Calendar getDateTimeZone() {
+  public Calendar getLocalTimeZone() {
     return zone;
   }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/MySqlDataTimeZone.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/timezone/MySqlDataTimeZone.java
@@ -1,6 +1,7 @@
 package io.ebeaninternal.server.core.timezone;
 
 import java.util.Calendar;
+import java.util.TimeZone;
 
 /**
  * Implementation of DataTimeZone when single Calendar instance is used with local timezone.
@@ -8,9 +9,15 @@ import java.util.Calendar;
 public class MySqlDataTimeZone implements DataTimeZone {
 
   protected final Calendar zone;
+  protected final Calendar localZone;
 
+  public MySqlDataTimeZone(String zoneId) {
+    this.zone = Calendar.getInstance(TimeZone.getTimeZone(zoneId));
+    this.localZone = Calendar.getInstance();
+  }
   public MySqlDataTimeZone() {
     this.zone = Calendar.getInstance();
+    this.localZone = zone;
   }
 
   @Override
@@ -20,6 +27,6 @@ public class MySqlDataTimeZone implements DataTimeZone {
 
   @Override
   public Calendar getLocalTimeZone() {
-    return zone;
+    return localZone;
   }
 }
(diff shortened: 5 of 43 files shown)


rPraml:
@rbygrave 
FYI: our originating problem was introduced with this commit https://github.com/ebean-orm/ebean/issues/1838
With ebean 12 - time only values gets converted to the current time offset. Unfortunately, such a conversion cannot be done properly without knowing the date for the time only value, because the offset may vary, if the zone has DST. 
So a GMT value of "04:15" will be converted here to "05:15" in CET (winter-time) and to "06:15" in CEST (summer-time)

independent from the json & conversion fixes, Most problems for our originating problem (converting Time only values) could be fixed with this commit: https://github.com/FOCONIS/ebean/commit/454f57eb1aa01ee5b315921c37ff979866517ff3

I also slightly modified the test case https://github.com/FOCONIS/ebean/commit/f58ea1644c6bc09aa2124c0bd751fd443d4eccaf and we will accept, that LocalDateTimes are converted to the current time zone (we only use instants in our data model, so this should not be a problem here)

Here are some conversion fixes, you should take a look:  https://github.com/FOCONIS/ebean/commit/b1efed18ddc19ed24997937e30b9db8eca04ab50 and https://github.com/FOCONIS/ebean/commit/f4a1b654ad2d351b00f7a634ef287ebe7745c602

All in all, it seems, that ebean works now when config is set to "GMT" and JRE time zone is >= GMT, but there may occur strange things in negative time zones. (A localDate for example can jump backwards, when saved/read from database or json)

The JsonConfig.DateTime.NANOS is also not handled properly for some data types (See checkmes and fixmes in this test case: https://github.com/FOCONIS/ebean/blob/ebean-new/ebean-test/src/test/java/io/ebeaninternal/server/type/DatesAndTimesTest.java#L275)

Cheers
Roland

rbygrave:
Do we have a failing test (for the behaviour that can in with #1838) ?

rPraml:
Yes, I noticed in the DateAndTimesTest that I get an 'off-by-one'  error when reading time-only values.

So the test is doing an `insert into mdate_time (id,joda_local_time) value (1, '05:15:15')` reads back the value as `06:15:15` when DataTimeZone is set to `GMT` and JVM time zone is set to `Europe/Berlin`
This would be explainable, because we are now GMT+1 - but what happens in summer, when we are GMT+2

I did a short test, and JodaLocalTime reads back 07:15:15 when I change the date of my machine to 2022-04-01 (=summer time.) (fortunately jdk LocalTime works - it reads back 06:15 in DST and non DST)
I think you will agree, that this would be a no-go if all time values will jump +/- one hour when DST is active. 

In my opinion a time-only value should be never converted.
Under no circumstance a time only value should be converted differently depending on weather the current machine date is DST or not.

Also a date-only value should be never converted: Currently, the date-only value is somewhere converted to a GMT timestamp.
So if I store a `birthday` in the database and read it back. I will get the correct date back, as long as my application runs in GMT+xx time, but I get an off-by-one error, when the machine timezone is changed to a GMT-xx zone.

The same, or even worse problems may occur, if you have two applications in different jvm time zones and they share data via json (distributed cache / elasticsearch / ...)

This PR shows, that dealing with dates and times become complex as soon as time zones will play a role.

I think the first thing that has to be done is to define the desried behaviour for each of the date/time datatypes.

cheers
Roland

rbygrave:
*> I think the first thing that has to be done is to define the desired
behaviour for each of the date/time datatypes.*

Maybe but I'd be keen to see a small application that uses both Java
LocalTime and Joda LocalTime and see the problem reproduced.  As we need to
mess with the JVM timezone etc I suspect this needs to be a small separate
application.  The commit that I thought was pointed to would impact both
types so at this stage I still would like to clearly see the problem in
terms of a failing test and understand the impact and scope.


*> JodaLocalTime reads back 07:15:15*

Yeah this sucks.  From reading the above this only impacts JodaLocalTime
which is important and interesting noting that the Joda time library is no
longer maintained and we should be removing all support for it at some
point.


*> I think the first thing that has to be done *

I personally still do not have a clear understanding of the bug. Does it
only impact Joda LocalTime, or other Joda types?  If it is Joda specific
what is the underlying issue such that it doesn't impact Java LocalTime?

This PR is called "Big refactor and bug fixing of date time handling" and
right now it feels like it should be more like "Joda LocalTime gives
incorrect values" or something that more clearly communicates the scope and
nature of the issue to other people. As I  see it we should NOT be mixing
"refactor" with "bug fix" - these are different tasks with different
priorities.  I would prefer for us to focus on "Bug" and get a really good
clear specific understanding of the bug first.

rPraml:
@rbygrave We will provide a test case for this and will split this PR in separate ones (btw. we do not use joda, so if the support will be dropped, it will not hurt us)

rbygrave:
@rPraml  Should this PR be discarded now?  Any further changes put into new PR's right?

rPraml:
We can close this PR. Do you have a time line when you remove the Joda support #2457

PR #2446, opened 2021-11-30
NPE Fix with .setCountDistinct(CountDistinctOrder.NO_ORDERING)

NSzemenyei:
Hi @rbygrave ,

we got a NPE in our application when we used .setCountDistinct(CountDistinctOrder.NO_ORDERING) in a query.

Here is a test and a solution for it. Can you please check it?

Kind regards
Noemi


rbygrave:
Nice, thanks.

PR #2447, opened 2021-12-02
Traverse class hierarchy, when searching for ScalarType

rPraml:
## Actual behaviour
```
 DB.find(SomeNewTypesBean.class).where().le("calendar", Calendar.getInstance()).findList();
```
throws "No ScalarType registered for GregorianCalendar"

## Expected behaviour

The above statement should work.

The problem is, that `Calendar.getInstance()` will return a `GregorianCalendar extends Calendar` instance.
As the `DefaultTypeManager` searches only for exact classes, it will not find a ScalarType for `GregorianCalendar`

This PR changes the TypeManager in a way, that it will check also the super classes and interfaces, if there is a scalarType registered, that can be used for this class. 

## Notes & thoughts
*Why are there so much code-changes?* 
Unfortunately, the concurrentHashMap does not support `null` values, that's why this ScalarTypeNotFound class is required to handle `cache misses` (an other way, could be to change the map `ConcurrentHashMap<Key, Optional<Value>>`)

*Why did we not extend `checkInterfaceTypes` and add a `Calendar` case like it is already done for `Path`?*
Of course, this would fix the above problem in a less invasive way
 - but we run into the same problem in our application, where we have registered a ScalarTypeConverter on an Interface
 - the same code path makes the `addEnumType` obsolete that handels overridden enum methods (https://github.com/ebean-orm/ebean/pull/2447/files#diff-ddcb5f956b942664002a3bdaf0c65220dc37fa173acadb0050fbfd4d7671f17bR213-L221) 
 - I think a generic approach is better, than the TypeManager has to know, that Calendar and Path (and enums) are handled special.


rbygrave:
It isn't clear what the issue is here.  It would be good to have a summary statement of what the issue is that this PR is supposed to test and fix.

danglotb:
Hi, we are researchers in software engineering, working on a novel GitHub bot for helping understand build failures.

We have run our prototype bot [flacocobot](https://github.com/SpoonLabs/flacoco) on this failing PR and below are the results.

What do you think?

Thanks a lot! --Davide (@dginelli), André (@andre15silva), Matias (@martinezmatias), Benjamin (@danglotb), Martin (@monperrus)

**********************************

The line (13) of the file io/ebean/test/JsonAssertContains has been identified with a suspiciousness value of 100.00%.

<details>
     <summary>Failing tests that cover this line</summary>

- `io.ebean.test.JsonAssertContainsTest#path_when_empty`
</details>
io/ebean/test/JsonAssertContains

**********************************

The line (15) of the file io/ebean/test/JsonAssertContains has been identified with a suspiciousness value of 100.00%.

<details>
     <summary>Failing tests that cover this line</summary>

- `io.ebean.test.JsonAssertContainsTest#path_when_empty`
</details>
io/ebean/test/JsonAssertContains

**********************************

The line (128) of the file io/ebean/test/JsonAssertContains has been identified with a suspiciousness value of 100.00%.

<details>
     <summary>Failing tests that cover this line</summary>

- `io.ebean.test.JsonAssertContainsTest#path_when_empty`
</details>
io/ebean/test/JsonAssertContains

**********************************

The line (129) of the file io/ebean/test/JsonAssertContains has been identified with a suspiciousness value of 100.00%.

<details>
     <summary>Failing tests that cover this line</summary>

- `io.ebean.test.JsonAssertContainsTest#path_when_empty`
</details>
io/ebean/test/JsonAssertContains

jonasPoehler:
The problem here was, that with the current code you couldn't use e.g. `GregorianCalendar` in a Query on a `Calendar`-property or a `WindowsPath`/`LinuxPath` on a `Path`-property (while there was an ebean-internal workaround for exactly that). Similar it was not possible to define own `ScalarTypeConverters` for interface types without also implementing the same converters for the implementations as well (even if not necessary as the interfaces provided all needed methods for conversion). I also hope, that the extension of tests makes the problem clearer.

rPraml:
@rbygrave  We have updated the PR and the description.

rbygrave:
Right, I understand it, looks good. Thanks.

PR #2449, opened 2021-12-03
Setting different server-wide defaultJsonMutation was not honored

jonasPoehler:
Hello Rob,

we have uncovered a problem with the default for the JsonMutationDetection. By default for a server this is set to `MutationDetection.HASH`, but when we wanted to set it to `MutationDetection.SOURCE`, Ebean didn't hand this information in far enough and the properties were working with hash after all. This PR provides a test and a possible fix for this issue. I tried to determine the **actual** mutation detection for a property when constructing the `DeployProperty`, so I don't have to handle the DatabaseConfig further down in `DeployProperty` or even `Property`. In the end, I'm still unsure if this would be the correct approach for fixing this, so maybe you can put in your two cents on this.

All the best
Jonas

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanProperty.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanProperty.java
@@ -269,7 +269,6 @@ public class DeployBeanProperty {
   }
 
   public void setMutationDetection(MutationDetection dirtyDetection) {
-    assert dirtyDetection != MutationDetection.DEFAULT; // This has to be determined beforehand
     this.mutationDetection = dirtyDetection;
   }
 
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanProperty.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/meta/DeployBeanProperty.java
@@ -269,6 +269,7 @@ public class DeployBeanProperty {
   }
 
   public void setMutationDetection(MutationDetection dirtyDetection) {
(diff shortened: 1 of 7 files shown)


rbygrave:
Have a review of https://github.com/ebean-orm/ebean/issues/2452 @jonasPoehler 

Note that there are some "tricks" around isolating / testing this sort of issue related to - Deployment, ScalarType assignment etc. If we create a test that extends `BaseTestCase` that can be difficult because `BaseTestCase` implicitly starts the default server, and THAT will run the deployment logic on ALL the entity types in the test suite.  This makes it really hard to effectively add in a break point into code like the `ScalarTypeJsonObjectMapper` etc ... because there are too many models, properties etc that go through that code.

In 2452 I have use the "tricks" that  A) That test does not extend `BaseTestCase` and B) Only to register the EBasicJsonList.class when creating that Database instance. These 2 things mean then we can use debug break points more effectively.

In terms of the fix in 2452, the summary is that the "tags" property was not a ScalarTypeJsonObjectMapper but instead being just a `List<String>` it gets different treatment / different ScalarType.  The reason for that is that because it is a simple scalar collection we can use the built in dirty detection provided by `ModifyAwareList, ModifyAwareSet, ModifyAwareMap`.

The resulting code after the fix in 2452 probably isn't "ideal" because we have these 3 cases, scalars, ObjectMapper and JsonNode and 2 places where we deal with `MutationDetection.DEFAULT` but I think we go with this for now.

rbygrave:
With #2452 merged, that means we should be able to just discard this PR. That means you kind of miss out on getting the kudos for your good work with creating the test and investigation. Sorry about that ... but yes, good work !!

rbygrave:
I'll close this PR now, we are not merging this is it was replaced by #2452

PR #2450, opened 2021-12-06
Using TCK Tested JDK builds of OpenJDK Making the switch to Zulu

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Please see more details at Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       uses: actions/setup-java@v2
       with:
         java-version: ${{ matrix.java_version }}
-        distribution: 'adopt'
+        distribution: 'zulu'
     - name: Maven cache
       uses: actions/cache@v2
       env:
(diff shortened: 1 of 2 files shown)


rbygrave:
Nice, thanks !!!

carldea:
> Nice, thanks !!!

Glad to have helped contribute!

rbygrave:
FYI: Also prompted me to then add a workflow for building against `18-ea` which is nice.  Good to build against the early access release etc :)

carldea:
> FYI: Also prompted me to then add a workflow for building against `18-ea` which is nice. Good to build against the early access release etc :)

Beat me to the punch.🙂

Cool!
Future proof it.

PR #2475, opened 2021-12-23
Fix locale problem with TestCsvReader and TestCsvReaderWithCallback.

bigjimlefou:
Fix to make tests pass on non-English jvm


rbygrave:
Nice, thanks !!

PR #2485, opened 2022-01-11
Fix utf8 problems with specialchars

NSzemenyei:
Hello @rbygrave ,

in our application we use some special UTF8 characters, e.g. ö,ü,ä, etc...
We had a problem with generating idx-checksums for migration scripts. If we generated the checksum in Eclipse, it returned an other checksum as maven in the cmd build.

I found this problem for encodings in Stackoverflow: https://stackoverflow.com/questions/12272698/why-does-maven-give-me-different-utf-8-characters-than-eclipse-test-run-in-ecli 

I fixed this problem in MChecksum.java. 

@rPraml checked how files/migrations will be read and written in other classes. He fixed the encodings to UTF8.

Can you check this PR and give us feedback?

Kind regards
Noemi


ebean-ddl-generator/src/test/java/io/ebeaninternal/dbmigration/MChecksumTest.java
```diff
@@ -10,9 +10,9 @@
 
   @Test
   public void calculate() {
-    File file = new File("src/test/resources/dbmigration/index/1.0__hello.sql");
+    File file = new File("src/test/resources/dbmigration/index-special-chars/1.0__hello.sql");
```
rbygrave, line 13: I don't think we want to lose the existing test right?

ebean-core/src/main/java/io/ebeaninternal/server/core/DScriptRunner.java
```diff
@@ -66,8 +66,9 @@ private String content(URL resource) {
       throw new IllegalArgumentException("resource is null?");
     }
 
-    try (InputStream inputStream = UrlHelper.openNoCache(resource)) {
-      return readContent(new InputStreamReader(inputStream));
+    try (InputStream inputStream = UrlHelper.openNoCache(resource);
+        Reader reader = IOUtils.newReader(inputStream)) {
```
rbygrave, line 70: Note to review indentation

rbygrave (commented):
Good stuff. in MChecksumTest are we able to keep that original test and have the special chars one as a new additional test?

ebean-core/src/main/java/io/ebeaninternal/server/core/DScriptRunner.java
NSzemenyei: @rbygrave Thanks for the comments! I have updated the PR

ebean-api/src/main/java/io/ebean/text/csv/CsvReader.java
```diff
@@ -18,7 +18,7 @@
  * try {
  *   File f = new File("src/test/resources/test1.csv");
  *
- *   FileReader reader = new FileReader(f);
+ *   FileReader reader = new FileReader(f, encoding);
```
rPraml, line 21: Note: `FileReader(f, encoding)` is new with java 11

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/DdlGenerator.java
```diff
@@ -348,7 +347,9 @@ protected String readFile(String fileName) throws IOException {
     if (!f.exists()) {
       return null;
     }
-    return readContent(new FileReader(f));
+    try (Reader reader = IOUtils.newReader(f)) {
```
rPraml, line 350: fixes also a possible resource leak

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/MChecksum.java
```diff
@@ -13,9 +15,8 @@
    * Returns the checksum of the file. Agnostic of encoding and new line character.
    */
   static int calculate(File file) {
-    try {
+    try (BufferedReader bufferedReader = IOUtils.newReader(file)) {
```
rPraml, line 18: **Important** @rbygrave This will introduce a breaking change, because we are now using UTF-8
If checksum was computed on a JVM where `Charset.defaultCharset()` is not UTF-8 (e.g. on a german windows) you may get a different checksum when the file contains special chars like "umlauts".

The risk is small, but it exists/should be mentioned in the changelog

Note: With JDK 18 https://openjdk.java.net/jeps/400, the charset will be set to UTF-8 by default, so you may get this problem anyway.

ebean-test/src/test/java/io/ebean/json/EJsonTests.java
```diff
@@ -46,15 +44,13 @@ public void test_map_simple() throws IOException {
...
-    writer.close();
-
-    log.info("write to file {}", temp.getAbsolutePath());
+    try (Writer writer = IOUtils.newWriter(temp)) {
+      Map<String, Object> map = new LinkedHashMap<>();
+      map.put("foo", "bar");
+      EJson.write(map, writer);
+      writer.write("The end.");
+    }
+    assertThat(temp).hasContent("{\"foo\":\"bar\"}The end.");
```
rPraml, line 53: I put an assert here, otherwise the test makes no sense. @rbygrave FYI

rPraml (approved):

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/MChecksum.java
rbygrave: I think this is fine in that although we generate the index with checksums it isn't being used/read yet.  The migration runner isn't actually using this checksum index file yet.

In that way we can think of this as non-breaking.

rbygrave (approved):

rbygrave:
Great work @NSzemenyei  !!

PR #2486, opened 2022-01-11
Fix some tests failing on Windows

bigjimlefou:
Small changes for cross platform home path and file separator


rbygrave:
Nice !!

PR #2501, opened 2022-01-25
Support h2database v2.x

rbygrave:


ebean-core/src/main/java/io/ebeaninternal/server/expression/platform/H2DbExpression.java
```diff
@@ -10,6 +10,7 @@
 
   @Override
   public void bitwise(SpiExpressionRequest request, String propName, BitwiseOp operator, long flags, String compare, long match) {
-    bitwiseFunction(request, propName, operator, compare);
+    String funcName = functionName(operator);
+    request.append(funcName).append("(").append(propName).append(", cast(? as long)) ").append(compare).append(" cast(? as long)");
```
rPraml, line 14: @jonasPoehler  FYI, we use a lot bitwise operations in our app

ebean-test/src/test/resources/ebean.properties
```diff
@@ -72,7 +72,7 @@ datasource.db.databaseDriver=org.h2.Driver
 
 datasource.h2.username=sa
 datasource.h2.password=
-datasource.h2.url=jdbc:h2:mem:testsMem;DB_CLOSE_ON_EXIT=FALSE
+datasource.h2.url=jdbc:h2:mem:testsMem;DB_CLOSE_ON_EXIT=FALSE;NON_KEYWORDS=KEY,VALUE
```
rPraml, line 75: interesting workaround :)
Would you add some comments, when this is needed (maybe in documentation for elementCollection)

rPraml (approved):
Most of the changes were similar in my try. The workaround for `KEY,VALUE` is interesting :)

PR #2504, opened 2022-01-26
Cannot delete bean with more than 2100 child beans on SqlServer

rPraml:
When deleting a element, that has many dependent elements, this won't work on SqlSever due 2100 parameter limit for "in" queries.

This PR adds a test and a possible fix


rbygrave:
Yeah cool. Longer term we could have a platform "maxInParameters".

Issue #900, opened 2016-11-25
Error - Can't fetch JsonDB field - org.postgresql.util.PSQLException: ERROR: could not identify an equality operator for type json


0xbaadf00d:
Ebean version 9.1.2

We can't use `.fetch("settings", "field")` where _field_ is annotated with @DbJson. Using only `.fetch("settings")` works, but in this case we don't want to retrieve all columns, we only need paypalSettings.

```java
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b15, mixed mode)
```

```java
ContractModel.find
            .select("id,uid")
            .fetch("settings", "paypalSettings")
            .where()
            .eq("payoutRequests.status", PayoutRequestStatus.SCHEDULED)
            .eq("payoutRequests.payoutMethod", PaymentMethod.PAYPAL)
            .isNull("payoutRequests.payoutId")
            .isNull("payoutRequests.payoutBatchId")
            .findEach(this::pushPayoutRequestOnPaypalBatch);
```

```java
@DbJson
@Column(name = "paypal_settings")
private Map<String, Object> paypalSettings;
```

```console
[error] a.a.OneForOneStrategy - Query threw SQLException:ERROR: could not identify an equality operator for type json
  Position : 48 
Bind values:[SCHEDULED,PAYPAL] 
Query was:
select distinct t0.id c0, t0.uid c1, t1.id c2, t1.paypal_settings c3 from contract t0 join contract_settings t1 on t1.contract_id = t0.id  join payout_request u1 on u1.contract_id = t0.id  where u1.status = ?  and u1.method = ?  and u1.payout_id is null  and u1.payout_batch_id is null  


javax.persistence.PersistenceException: Query threw SQLException:ERROR: could not identify an equality operator for type json
  Position : 48 
Bind values:[SCHEDULED,PAYPAL] 
Query was:
select distinct t0.id c0, t0.uid c1, t1.id c2, t1.paypal_settings c3 from contract t0 join contract_settings t1 on t1.contract_id = t0.id  join payout_request u1 on u1.contract_id = t0.id  where u1.status = ?  and u1.method = ?  and u1.payout_id is null  and u1.payout_batch_id is null  


    at com.avaje.ebeaninternal.server.query.CQuery.createPersistenceException(CQuery.java:680)
[...]

PR #2511, opened 2022-01-27
Unselect LOBs when making a distinct select

### LINKED ISSUE #2495: DB2: select distinct with lob/clob fields
## Actual behavior

If a query needs ```distinct``` in a select clause (e.g. join many tables) and a lob/clob field is also in the select, DB2 throws an SQL Exception ```SQLCODE=-134, SQLSTATE=42907,```, cause: https://www.sqlerror.de/db2_sql_error_-134_sqlstate_42907.html 

@rPraml and I implemented a possible solution in our fork: https://github.com/FOCONIS/ebean/pull/55 . 
Our idea is to exclude the lob/clob properties from the ```select distinct``` clauses. These props have to be loaded lazy.
We assume, the same fix would be also good for Oracle,

### Steps to reproduce
Execute the test without the fixes in https://github.com/FOCONIS/ebean/pull/55

@rbygrave can you please check this issue and give us feedback?

rPraml:
Some platforms like DB2, Oracle and also Postgres do not allow to perform a `select distinct` when LOB columns are used.

There was a related issue #900 for postgres, but DB2 does not support `select distinct on (id)`
Workarounds like `cast(lob_column as varchar(32000))` or `DBMS_LOB.substr(lob_column , 32000)` would be not reliable enough. So for these platforms we must not select lobs, when performing a distinct query.

Ebean automatically unselect these columns and rely on lazy load of the properties.

Note: There are some edge cases, when that will not work:
- Ebean unselects the lobs and lazy load is disabled
- the lob is part of a findSingleAttribute-query and distinct is used.
  In this case you can use `select("cast(lobProp as varchar(32000))")` or `fetch("el.path", "cast(lobProp as varchar(32000))")` in your application

This fixes #2495

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
@@ -1160,7 +1160,26 @@ public class BeanProperty implements ElPropertyValue, Property, STreeProperty {
   public boolean isLob() {
     return lob;
   }
-
+  
+  /**
+   * Returns true if this <code>isLob()</code> or the type will effectively map to a lob. 
+   */
+  public boolean isDbLob() {
+    if (lob) {
+      return true;
+    }
+    switch (dbType) {
+    case DbPlatformType.JSON:
+    case DbPlatformType.JSONB:
+      return dbLength == 0; // must be analog to DbPlatformTypeMapping.lookup
+    case DbPlatformType.JSONBlob:
+    case DbPlatformType.JSONClob:
+      return true;
+    default:
+      return false;
+    }
+  }
+  
   public static boolean isLobType(int type) {
     switch (type) {
       case Types.CLOB:
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
@@ -760,6 +760,10 @@ final class CQueryBuilder {
     return dbPlatform.isPlatform(Platform.POSTGRES);
   }
 
+  boolean isPlatformDistinctNoLobs() {
+    return dbPlatform.isPlatform(Platform.DB2); // CHECKME: Also oracle?
+  }
+
   /**
    * Return the 'for update' FROM hint (sql server).
    */
--- a/ebean-test/src/test/java/org/tests/json/TestJsonMapBasic.java
+++ b/ebean-test/src/test/java/org/tests/json/TestJsonMapBasic.java
@@ -3,13 +3,18 @@ package org.tests.json;
 import io.ebean.BaseTestCase;
 import io.ebean.DB;
 import io.ebean.Query;
+import io.ebean.annotation.ForPlatform;
+import io.ebean.annotation.Platform;
 import io.ebean.test.LoggedSql;
 import io.ebean.text.json.EJson;
+
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.tests.model.json.EBasicJsonMap;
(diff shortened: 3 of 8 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -1160,7 +1160,26 @@ public boolean isFetchEager() {
...
+    if (lob) {
+      return true;
+    }
+    switch (dbType) {
+    case DbPlatformType.JSON:
+    case DbPlatformType.JSONB:
+      return dbLength == 0; // must be analog to DbPlatformTypeMapping.lookup
+    case DbPlatformType.JSONBlob:
+    case DbPlatformType.JSONClob:
+      return true;
```
rPraml, line 1177: Should we delegate that decision to the DB-platform?

ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
```diff
@@ -760,6 +760,10 @@ boolean isPlatformDistinctOn() {
     return dbPlatform.isPlatform(Platform.POSTGRES);
   }
 
+  boolean isPlatformDistinctNoLobs() {
+    return dbPlatform.isPlatform(Platform.DB2); // CHECKME: Also oracle?
```
rPraml, line 764: I did not yet test that in oracle. That may be also delegated to DatabasePlatform
Maybe returning a enum: 
```java
enum DistinctMode { NORMAL, DISTINCT_ON, UNSELECT_LOBS}
```

ebean-test/src/test/java/org/tests/json/TestJsonMapBasic.java
```diff
@@ -18,15 +23,19 @@
...
-  public void whereManyPredicate() {
+  private final EBasicJsonMap bean = new EBasicJsonMap();;
 
-    if (!isPostgres()) {
-      // testing postgres specific select distinct on clause
-      return;
+  @AfterEach
+  void cleanup() {
+    if (bean != null && bean.getId() != null) {
+      DB.delete(EBasicJsonMap.class, bean.getId());
```
rPraml, line 31: refactored cleanup a bit

ebean-test/src/test/java/org/tests/json/TestJsonMapBasic.java
```diff
@@ -50,6 +59,36 @@ public void whereManyPredicate() {
     assertThat(query2.getGeneratedSql()).contains("select distinct on (t0.id) ");
   }
 
+  @Test
+  @ForPlatform(Platform.DB2)
+  public void whereManyPredicateDb2() {
```
rPraml, line 64: Unfortunately, the test is not yet starting on current master branch, because of the  `maxTableNameLength` and `maxConstraintNameLength` of 18 char limits in DB2-platform. (changing them to 128 and the test will run)
We will introduce a Legacy/DB2LUW/... platform as suggested  in #2484 soon

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
rbygrave: At the moment I think the above is good and that we should not delegate this to the DB-platform.  I **_believe_** JSON types are generally more "clob" than "varchar" in nature ... it is a little tempting to not even have the dbLength == 0 check.

rbygrave (approved):

ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
rbygrave: Hmmm, tests all passing with Oracle I believe (last local run, odd problems running oracle test via github workflow)

PR #2512, opened 2022-01-27
ADD: different DB2 platforms

rPraml:
This will split up in different DB2 platforms.
It is now required to specify which DB2 platform you have. See #2514


.github/workflows/multi-platform-test.yml
```diff
@@ -43,3 +43,5 @@ jobs:
       run: mvn clean test -Dprops.file=testconfig/ebean-sqlserver17.properties
     - name: mariadb
       run: mvn clean test -Dprops.file=testconfig/ebean-mariadb.properties
+    - name: db2
+      run: mvn clean test -Dprops.file=testconfig/ebean-db2.properties
```
rPraml, line 47: How can I run that workflow?

ebean-ddl-generator/src/test/resources/dbmigration/migrationtest/db2/1.1.sql
```diff
@@ -36,16 +36,16 @@ alter table migtest_e_basic alter column status set default 'A';
 alter table migtest_e_basic alter column status set not null;
 alter table migtest_e_basic add constraint ck_mgtst__bsc_stts check ( status in ('N','A','I','?'));
 alter table migtest_e_basic drop constraint ck_mgtst__b_z543fg;
-alter table migtest_e_basic alter column status2 varchar(127);
+alter table migtest_e_basic alter column status2 set data type varchar(127);
 alter table migtest_e_basic alter column status2 drop default;
-alter table migtest_e_basic alter column status2 set null;
+alter table migtest_e_basic alter column status2 drop not null;
```
rPraml, line 41: The old statements were wrong. I do not know if they have ever worked.
An `alter table migtest_e_basic alter column status2 set null;` will not work on DB2 11.5 LUW.
@bigjimlefou maybe you can test that on DB2ForI

ebean-core/src/main/java/io/ebeaninternal/server/core/DatabasePlatformFactory.java
```diff
@@ -159,7 +174,7 @@ private DatabasePlatform byDatabaseMeta(DatabaseMetaData metaData, Connection co
     } else if (dbProductName.contains("sqlite")) {
       return new SQLitePlatform();
     } else if (dbProductName.contains("db2")) {
-      return new DB2Platform();
+      throw new IllegalArgumentException("For DB2 please explicitly choose either db2legacy/db2luw/db2zos/db2fori platform. Refer to issue #2514 for details");
```
rPraml, line 177: @rbygrave Should we make some autodetection here?

.github/workflows/multi-platform-test.yml
rbygrave: Maybe it's a permissions thing ... let me see

ebean-core/src/main/java/io/ebeaninternal/server/core/DatabasePlatformFactory.java
rbygrave: I am happy for it to be explicit.  Maybe go with this for now and see later on if auto-detection is worthwhile?

PR #2513, opened 2022-01-27
Test against H2 1.4.199

rPraml:
I made some tests, if ebean is still H2 1.4 compatible. (Not everyone can upgrade immediately)

This PR shows, what has to be done, to be 1.4 compatible.

apart from the ConnectionString, the only critical point is `convertArrayType`
@rbygrave What do you think, should we add a compatibiliy-flag for H2 1.xxx

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
   <properties>
     <nexus.staging.autoReleaseAfterClose>false</nexus.staging.autoReleaseAfterClose>
     <jackson.version>2.12.1</jackson.version>
-    <h2database.version>1.4.199</h2database.version>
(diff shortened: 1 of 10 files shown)


rbygrave:
Yes, its a good idea.

pom.xml
```diff
@@ -99,6 +99,12 @@
     <profile>
       <id>release</id>
     </profile>
+    <profile>
+      <id>h2v1</id>
+      <properties>
+        <h2database.version>1.4.199</h2database.version>
+      </properties>
+    </profile>
```
rPraml, line 107: We can add a build plan with -Ph2v1

PR #2521, opened 2022-01-28
DB2: Fix getGenerateKeys (ugly workaround)

rPraml:
The IBM guys invented a new way how to retrieve keys in batch insert.
You have to use `ResultSet[] DB2PreparedStatement.getDBGeneratedKeys()` instead of `getGeneratedKeys` of the JDBC-API

See:
https://stackoverflow.com/questions/41725492/how-to-get-auto-generated-keys-of-batch-insert-statement

@rbygrave I know, this is a workaround. I would have feedback, what's the best way to solve this (delegate to database-platform? or retrieve a 'fetcher' from the platform)

OLD
`Tests run: 3104, Failures: 79, Errors: 97, Skipped: 116`

NEW
`Tests run: 3104, Failures: 71, Errors: 78, Skipped: 116`
Would fix 27 tests 👍


rbygrave:
Maybe just pull it into a separate class. Let me try that.

PR #2528, opened 2022-01-28
DB2 Fix tests for DB2

rPraml:
This PR modifies or disables some tests to be DB2 compatible


ebean-test/src/test/java/org/tests/model/m2m/TestM2MWithWhere.java
```diff
@@ -123,6 +125,11 @@ private void createTestData() {
     // System.out.println(sb); dump the table
   }
 
+  @AfterEach
+  void deleteTestData() {
+    DB.find(MnyEdge.class).delete();
+    DB.find(MnyNode.class).delete();
+  }
```
rPraml, line 132: Just clean up test data

PR #2558, opened 2022-02-08
Bug: save with cascade of @OneToOne + @Version model sets version to 1

AlexWagner:
When saving a model, that contains a @OneToOne Model with @Version, the child models' version will be set to 1.
Then, when the child model will be saved, a OptimisticLockException occures due to the wrong version.
(see test)

Furthermore the 'child'-model also has to have a xToMany-Relation to another model, so that the flag 'saveRecurseSkippable' in BeanPropertyAssoc is set to false.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -360,7 +360,7 @@ public final class EntityBeanIntercept implements Serializable {
    * Return true if the entity should be updated.
    */
   public boolean isUpdate() {
-    return forceUpdate || state == STATE_LOADED;
+    return forceUpdate || state == STATE_LOADED || state == STATE_REFERENCE;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
```diff
@@ -360,7 +360,7 @@ public void setForceUpdate(boolean forceUpdate) {
    * Return true if the entity should be updated.
    */
   public boolean isUpdate() {
-    return forceUpdate || state == STATE_LOADED;
+    return forceUpdate || state == STATE_LOADED || state == STATE_REFERENCE;
```
AlexWagner, line 363: In our oppinion this would be the fix.
Why would you want to insert a reference-bean?

rbygrave:
Apologies for taking a while to look at this.

ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
rbygrave: Yes, I agree with this fix. Thanks

Issue #2559, opened 2022-02-11
ENH: Support different ColumnDefinition for different database platforms


NSzemenyei:
Hi @rbygrave ,

with db2 we have the problem, we define columns as ```blob``` (without length):

Java Code:

```java
  public class ReportResult(
  ...
  @Lob
  private File fileContent;
  ...
  )
```

generated init migration:

```java
  create table report_result(
  ...
  file_content      blob,
  ...
  )
```
In this case we can only save a file that is not bigger as 1 MB (default size of db2 blob).
We would like to define the columnDefinition for a field more precise: different length for different platforms.

We would like to create a new annotation (e.g. DbColumn?) which uses the ```javax.persistence.@Column``` as a base and can be defined for every platform different:

```java
  public class ReportResult(
  ...
  @Lob
  @DbColumn(columnDefinition = "blob(64M)", platform = Platfrom.DB2)
  // for all other platforms default generated type
  private File fileContent;
  ...
  )
```

What do you think about it?

NSzemenyei:
I tried to implement the new annotation, but it would be a huge modification. The xmls of the column definitions had to be also modified (list of platform-columnDefinition pairs) and lot of code in ebean.

We designed a new proposed solution which has only small code changes in ebean but it "redefines" the usage of ``` @Column ```. (It's backward compatible)

Example:
```java
 @Column(columnDefinition = "db2;blob(64M);sqlserver,h2;varchar(227);varchar(127)")
  String description;
```
Using the pattern
```java
 @Column(columnDefinition = "PLATFORM1;DEFINITION1;PLATFORM2;DEFINITON2;DEFINITON-DEFAULT")
```

We use semicolon as separator because it's very unlikely that it would be used for column definition.

You can check the ebean code change here: https://github.com/FOCONIS/ebean/commit/3d83c6bd0e2bcca6fc04f73ce08983266b9df636 

What do you think about this?

rbygrave:
I like it - nice approach !!

NSzemenyei:
Thank you! I opened a PR for it: https://github.com/ebean-orm/ebean/pull/2563

rbygrave:
Closing with the merged PR.  Nice work !!

PR #2563, opened 2022-02-17
Different ColumnDefinition for different platforms

NSzemenyei:
Hello @rbygrave ,

I opened a PR for the issue https://github.com/ebean-orm/ebean/issues/2559

Can you please check it?

Best wishes
Noemi


ebean-test/src/test/java/misc/migration/v1_0/EBasic.java
```diff
@@ -41,6 +42,9 @@
 
   @Size(max=127)
   String description;
+  
+  @Column(columnDefinition = "db2;blob(64M);sqlserver,h2;varchar(227);varchar(127)")
+  String description2;
```
rPraml, line 47: Please add the same in v1_2. (or only in v1_1) The idea is to check both directions. The models in v1_0 and v1_2 should be equal

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/PlatformDdl.java
```diff
@@ -317,12 +319,32 @@ public String convert(String type) {
...
+    assert tmp.length % 2 == 1;
+    for (int i = 0; i < tmp.length - 2; i+=2) {
+      String[] platforms = tmp[i].split(",");
+      for (String plat : platforms) {
+        if (platform.isPlatform(Platform.valueOf(plat.toUpperCase(Locale.ENGLISH)))) {
+          return tmp[i+1];
+        }
+      }
+    }
+    return tmp[tmp.length-1]; // else
```
rPraml, line 345: reformat/indent whole block

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/PlatformDdl.java
```diff
@@ -317,12 +319,32 @@ public String convert(String type) {
...
     }
     return typeConverter.convert(type);
   }
 
+  // if columnType is different for different platforms, use pattern
+  // @Column(columnDefinition = PLATFORM1;DEFINITION1;PLATFORM2;DEFINITON2;DEFINITON-DEFAULT)
+  // e.g. @Column(columnDefinition = "db2;blob(64M);sqlserver,h2;varchar(227);varchar(127)")
+  private String extract(String type) {
+    String[] tmp = type.split(";");
+    assert tmp.length % 2 == 1;
```
rPraml, line 336:
```suggestion
    assert tmp.length % 2 == 1 : "You need an odd number of arguments. See Issue #2559 for details";
```
maybe throw an IAE (asserts may be disabled)

NSzemenyei:
@rbygrave  we implemented also, if the columnDefinition ends with a semicolon, the not given platform types will be calculated from ebean:

`@Column(columnDefinition = "db2;blob(64M);")` means: blob(64M) for db2, other platforms will be determined in `BeanProperty.renderDbType`

PR #2566, opened 2022-02-23
Refactored DDL Generation (use DdlWrite in alterColumn) and fix default and constraint handling for MySql/SqlServer

rPraml:
Hello Rob,

this PR does also some refactoring on DDL generation. It mainly targets the usage of one central "alterColumn" function instead of having one for type, default, notNull and a combined for baseAttributes.

the `alterColumnBaseAttributes` is renamed to `alterColumn` and the other 3 were deleted. The platform has to handle itself, if it uses one combined (MySql/SqlServer/Hana) or 3 individual statemets.

The effective code change in this PR is in this commit https://github.com/ebean-orm/ebean/pull/2566/commits/e6f95f4188eff87fb5396152a616521b9ab24f3d

My next steps are to sort the table alters by table-name, so that I can 
- call one reorg after table alter
- better handling of history (disable history before, and enable it laters)
- use multi-statents for mariadb/hana (instead of having two statements `alter table t1 alter col1 ...`, `alter table t1 alter col1 ...` we can do that in one statement for mariadb `alter table t1 alter col1 ..., alter col2 ...` which improves performance)

I've tested the DbMigrationTest for sqlserver & mariadb and regenerated the reference scripts

cheers Roland

PR #2567, opened 2022-02-24
Tidy up ddl generation (no effective code change)

rPraml:
These are the first two commits of #2566 and they contain no effective code change.

I decided to replace all `DdlWrite write` by `DdlWrite writer`, because the later one was already used more often in existing code.


rbygrave:
Nice, thanks !!!

Issue #2510, opened 2022-01-26
Monitor Yugabyte support


rbygrave:
Monitor https://github.com/yugabyte/yugabyte-db/issues/11186

For those unaware of Yugabyte, it has Postgres compatible types and API - uses Postgres JDBC driver etc (so like Cockroach in this respect).

Features supported: https://github.com/yugabyte/yugabyte-db/blob/master/architecture/YSQL-Features-Supported.md

--------
Issues:
- `create table like ` ... is not supported. Error executing stmt[create table hembi_bean_history(like hembi_bean)] error[ERROR: LIKE clause not supported yet
  Hint: See https://github.com/YugaByte/yugabyte-db/issues/1129. Click '+' on the description to raise its priority
  Position: 33]
- ERROR: column "plain_bean" is of type json but expression is of type character varying   Hint: You will need to rewrite or cast the expression.
- column "cidr" is of type cidr but expression is of type character varying   Hint: You will need to rewrite or cast the expression.
- ERROR: function pgp_sym_encrypt(character varying, character varying) does not exist   Hint: No function matches the given name and argument types. You might need to add explicit type casts.

rbygrave:
Add support in ebean-test-docker for yugabyte docker - https://github.com/ebean-orm/ebean-test-docker/pull/64

rbygrave:
Add yugabyte as platform defined in ebean-annotation Platform - https://github.com/ebean-orm/ebean-annotation/issues/50

rbygrave:
Initial platform support with DDL generation and Platform detection - https://github.com/ebean-orm/ebean/pull/2516

rbygrave:
Use Postgres JSON/JSONB ScalarTypes (but not HSTORE) - https://github.com/ebean-orm/ebean/pull/2529

rbygrave:
Modify ebean-test to support `ebean.test.platform=yugabyte` - https://github.com/ebean-orm/ebean/pull/2534

rbygrave:
Automated tests against Yugabyte passing (with dedicated github workflow) - https://github.com/ebean-orm/ebean/actions/workflows/yugabyte.yml

priyanshi-yb:
Thanks Ebean team for such quick support! 
[...]

PR #2572, opened 2022-02-25
[DDL-yugabyte] Adding workarounds for Yugabyte to make the DbMigrationTest passing

rPraml:
This adds workarounds for Yugabyte-DDL generation.
in Yugabyte, after a DML statement, (insert, update, delete), the transaction has to be commited.

I'm considering if the commit statements should be added automatically after insert/uptate/delete when generating the yugabyte-DDL file:

- adding `boolean needsDmlCommit()` to PlatformDdl (which returns true for YB)
- detecting, if we have an IUD statement and add commt after the statements (somewhere here https://github.com/ebean-orm/ebean/blob/master/ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/BaseTableDdl.java#L170)

On the other side, this could/should be also fixed in yugabyte: https://github.com/yugabyte/yugabyte-db/issues/3109

@rbygrave what do you think?


ebean-test/src/test/java/misc/migration/v1_1/EBasic.java
```diff
@@ -41,6 +41,10 @@
   Integer id;
 
   @NotNull
+  @DbMigration(platforms = Platform.YUGABYTE, preAlter = {
+      "update ${table} set ${column} = ${default} where ${column} is null",
+      "commit transaction /* 1 */" })
```
rPraml, line 46: yugabyte needs commits here
@rbygrave what do you think, if the ddl generator will handle the case. e.g. here: https://github.com/ebean-orm/ebean/blob/master/ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/BaseTableDdl.java#L130

ebean-test/src/test/resources/migrationtest/dbmigration/yugabyte/1.1.sql
```diff
@@ -55,13 +56,21 @@ alter table migtest_e_basic alter column status2 drop not null;
 alter table migtest_e_basic add constraint uq_migtest_e_basic_description unique  (description);
 
 insert into migtest_e_user (id) select distinct user_id from migtest_e_basic;
+commit transaction /* 3 */;
```
rPraml, line 59: @rbygrave what is the reason, that the duplicate lines are skipped in DdlRunner?

rbygrave: Not sure.  What is the code that does that?

rbygrave:
I don't think we should rush to put in workarounds for what are bugs in Yugabyte (or Cockroach).

rPraml:
I agree, as this PR contains no real fixes, it should not be merged hastily.

Nearly all workarounds are done with modified DbMigration annotations in src/test and the one in src/main could be also covered by a proper DbMigration annotation.

Should I keep that PR open (for documentation)?

The other question about duplicates targets this code https://github.com/ebean-orm/ebean-ddl-runner/blob/7137006f9e3ab97a323db3f43139bfd6349f002a/src/main/java/io/ebean/ddlrunner/DdlRunner.java#L67
Why is it necessary to have unique statements? What was the intention and can/should this behavior changed in DdlRunner? 
There is also a workaround-counter variable in HanaHistoryDdl and we also fell into that trap with hand written scripts.

Cheers
Roland

ebean-test/src/test/resources/migrationtest/dbmigration/yugabyte/1.1.sql
rbygrave: Hmmm. No git history there, it has been moved from ddl-generator I believe.  Now I'd say that isn't necessary and we can remove that `noDuplicates` list altogether.

rbygrave:
Lets keep this around for documentation for a while and see how we feel in a few weeks.

> DdlRunner.java#L67

I've removed that code and released `ebean-ddl-runner 1.3`

rbygrave:
Closing this for now, added comments into  https://github.com/ebean-orm/ebean/issues/2510

PR #2574, opened 2022-02-25
[DDL-Cockroach] Fixes for Cockroach platform

rPraml:
Disables history support (as this is not yet implemented) and adds a workaround for drop unique constraint


ebean-test/testconfig/ebean-cockroach.properties
```diff
@@ -1,4 +1,5 @@
 ebean.test.platform=cockroach
 ebean.test.dbName=unit
 datasource.default=cockroach
+#ebean.test.cockroach.version=v21.2.6
```
rPraml, line 4: We need at least v21.2 here, otherwise the test will not pass
Requires https://github.com/ebean-orm/ebean-test-docker/issues/67 first

PR #2575, opened 2022-02-25
[DDL-ClickHouse] fixes for ClickHouse platform

rPraml:
This will fix DDL generation for ClickHouse true/false variables and fix the DScriptRunner, so that DDL scripts could be executed in ClickHouse


ebean-core/src/main/java/io/ebeaninternal/server/core/DScriptRunner.java
```diff
@@ -106,7 +106,10 @@ private void runScript(String content, String scriptName, Map<String, String> pl
       try (Connection connection = obtainConnection()) {
         DdlRunner runner = new DdlRunner(useAutoCommit, scriptName, platformName);
         runner.runAll(content, connection);
-        connection.commit();
+        if (!connection.getAutoCommit()) {
+          // Clickhouse fails: Cannot commit when auto-commit is enabled
+          connection.commit();
```
rPraml, line 111: as far as I see, clickhouse does not support transactions. So is Ebean able to use clickhouse at all?

rbygrave: This ignores the use of `boolean useAutoCommit` ??  This doesn't look right.

rPraml: The problem is in ClickHouse, that you have to set autoCommit to true and must not call `connection.commit()`
So if we are already in autoCommit mode (which is true for clickhouse only) there is no need for doing an additional commit,  isn't?

@rbygrave How good is the support of ClickHouse in ebean? In my tests, I could only read from the DB as write access always opens a transaction and is calling `commit` at the end of the transaction

PR #2584, opened 2022-03-04
Small code change for Hazelcast 4.1.5

NSzemenyei:
Hello @rbygrave ,

this code change is needed for Hazelcast 4.1.5 (https://github.com/ebean-orm/ebean-hazelcast/pull/10) because the hazelcast cache map is since 4.X an UnmodifiableMap.
As we analysed, the iterator will only used for iterating over the collection and it won't be used in any code block after.

Can you please check it, if we can remove this line?

Kind regards
Noemi


rbygrave:
Yes, looks fine.

PR #2590, opened 2022-03-11
Provide custom BackgroundExecutorWrapper to pass thread locals

NSzemenyei:
Hi @rbygrave ,

we might have found a bug with l2 cache updates if the update runs in background.

Our use case:
we are working on the hazelcast integration in our application. We have the ebean-hazelcast dependency in our pom.xml. We only want to use hazelcast if there is a hazelcast xml config file in the classpath --> in this case we would set `ebean.serverCachePlugin=io.ebean.hazelcast.HzCachePlugin` .

For our unittests with h2 we set `ebean.serverCachePlugin=io.ebeaninternal.server.cache.DefaultServerCachePlugin` .
We have now problems with the tests that are handling data in different tenants.

@rPraml and I created a unittest for this: `TestBeanCacheAsync`.

The problem is: in `InternalConfiguration.initServerCachePlugin()`, If we use `ebean.serverCachePlugin=io.ebeaninternal.server.cache.DefaultServerCachePlugin` , the variable `localL2Caching` won't be set to `true`. Afterwards if we call `db.save(bean)` in `PostCommitProcessing` the cache won't be updated in foreground, it will call `backgroundExecutor.execute(postCommit.backgroundNotify()` async. These does not "copy" the tenant infos to the thread of `backgroundExecutor`, it will run in the "wrong tenant".

To solve this problem we introduced the `BackgroundExecutorWrapper`. It can be defined in the `DatabaseConfig`.

We also have a second problem, which can theoretically also occur by hazelcast-cache: if the update of the cache happens async in a `backgroundExecutor`, it isn't guaranteed that it would be done before the next `DB.find()` will be called. That's why we had to add a `Thread.sleep(100)` in our `TestBeanCacheAsync`.

Can you please check the first problem with `BackgroundExecutorWrapper` if the fix would be ok?

Kind regards
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
@@ -1,11 +1,13 @@
 package io.ebeaninternal.server.executor;
 
 import io.avaje.lang.NonNullApi;
+import io.ebean.config.BackgroundExecutorWrapper;
 import io.ebeaninternal.api.SpiBackgroundExecutor;
-import org.slf4j.MDC;
 
-import java.util.Map;
 import java.util.concurrent.*;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The default implementation of the BackgroundExecutor.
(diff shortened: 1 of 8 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
```diff
@@ -1,119 +1,147 @@
...
   public Future<?> submit(Runnable task) {
-    return pool.submit(wrapMDC(task));
+    return pool.submit(wrap(task));
   }
 
+  
   @Override
   public void execute(Runnable task) {
-    submit(task);
+    submit(logExceptions(task));
```
rbygrave, line 105: My gut says we should move this to be on `submit(Runnable)`

ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
```diff
@@ -1,119 +1,147 @@
...
+        logger.error("Error while executing the task {}", task, t);
+        throw t;
+      }
+    };
   }
 
   @Override
   public <T> Future<T> submit(Callable<T> task) {
-    return pool.submit(wrapMDC(task));
+    return pool.submit(wrap(task));
```
rbygrave, line 91: Maybe add `// Note: Here is no "logExceptions", because it is intended to check Future.get() by the invoker`

rbygrave (approved):

PR #2595, opened 2022-03-15
NPE by lazy loading a cached bean

NSzemenyei:
Hi @rbygrave 

we were hunting an NPE for several months. It didn't occured in our test enviroments only by some of our customers in production, we were not able to reproduce it with unit or UI tests. We analyzed several logs, the database of the customer but in the end we had to create a heap dump after the NPE happened.

In the heap dump we checked EntityBeanIntercept:
image

This is the state after the getPropertyX() call. This helped us to understand and reproduce the NPE in a unit test: `loadedFromCache = true`, `state=2` and `lazyLoadedProperty = 34` --> it was not resetted to -1.

In `TestWithCacheAndLazyLoad.testBatch()` two cached entities & parents will be created. After that with a `DB.findById` the entity will be loaded from the database and put into the cache.

In a separate transaction we are searching for a list of the parent entities and we load an entity `temp` with `DB.findById` . By calling `temp.getChild().getName()` the cache will be used at `line 198 - DLoadBeanContext.loadBean(ebi)` and the `hit` will be removed from the `list`. The next call in the test (`list.get(0).getChild().getAddress()`) refers to the bean which was removed in the last iteration from the cache (`list.removeAll(hits)`) but the other entity is still in `list`.  That's why `list.isEmpty() = false` at line 194 --> `else if (context.hitCache) = true` and `if (list.isEmpty() || hits.contains(ebi)) = true`  --> the method returns and the bean won't be loaded.

Our suggestion would be at line 194: `if (list.isEmpty())` --> `if (!list.contains(ebi))`

In a third commit we changed the List to a Set to get better performance.

Can you please check it?

Kind regards
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/LoadBeanBuffer.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/LoadBeanBuffer.java
@@ -4,7 +4,7 @@ import io.ebean.bean.EntityBeanIntercept;
 import io.ebean.bean.PersistenceContext;
 import io.ebeaninternal.server.deploy.BeanDescriptor;
 
-import java.util.List;
+import java.util.Set;
 
 /**
  * A buffer of beans for batch lazy loading and secondary query loading.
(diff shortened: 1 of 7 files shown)


rbygrave:
Yes, excellent work. This fix looks exactly correct.  Also love the 3 separate commits - very nice !!!

Issue #2556, opened 2022-02-07
DB2 reorg table support


rPraml:
When tables/columns are altered, you might need a `CALL SYSPROC.ADMIN_CMD ('REORG TABLE ${table} ')` after certain statements (e.g. dropping a column, changing data type, ...)

Ebean currently supports auto-generation of migration scripts, that works well in ~95% of the cases, but not for DB2.
In order to generate also working scripts for DB2, we added a "needsReorg" set, that stores the tables that require a reorg (protoype is here: https://github.com/FOCONIS/ebean/blob/master/ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/BaseTableDdl.java#L94)

In short
- you can do as many `alter table`  on the tables (some alter require a reorg, other not)
- you have to do a `reorg table` if you have done an alter, that requires a reorg before
  - adding indices
  - accessing data via index

Unfortunately, ebean currently writes the DDL scripts per column in the following order:

```sql
-- prealter of column1
-- alter table1 colum1
-- reorg table1
-- postalter of column1

-- prealter of column2
-- alter table1 colum2
-- reorg table1
-- postalter of column2

-- create all indices
```

This order is not optimal, because you might need two reorgs. 
So I think it would be a good idea to change the order to write all prealter, alters, reorgs and postalters
```sql
-- prealter of column1
-- prealter of column2
-- alter table1 colum1
-- alter table1 colum2
-- reorg table1
-- postalter of column1
-- postalter of column2

-- create all indices
```
@rbygrave This problem is a bit DB2 specific, what do you generally think about this change?

rbygrave:
Yeah, I'm ok with it.  Noting that ebean-ddl-generator should generally only be a test scope dependency so adding this sort of DB2 specific handling is ok.

rbygrave:
@rPraml I think we can close this now due to #2596

PR #2596, opened 2022-03-15
DDl-Refactor: Add DB2 Reorg support and history support

rPraml:
This PR moves Ebean to a state where it has working DB2 migration (with proper reorg-table statements. See #2556)

We have also a working history now: [DB2: HistorySupport added](https://github.com/ebean-orm/ebean/pull/2596/commits/c9204eed3f5fc7d3472434404c69559c0d4c0b7c)

Unfortunately, DB2 needs an exact copy of the history table. So it is not allowed to have `not null` column in live table and `null` column in history table as it is common in trigger based history

@rbygrave There is currently a hack `syncExact = platformDdl.getPlatform().isPlatform(Platform.DB2)` for DB2
I would change that **for all platforms.** This means: 
- If you want to alter a column from `null` to `not null`, you have to migrate also the null values in history table.

Docker DB2 test is passing now


rbygrave:
Well done - a very big effort here !!

PR #2597, opened 2022-03-16
Allow Line Wrap in JavaMelody

rPraml:
We use Ebean & [JavaMelody](https://github.com/javamelody/javamelody) in our application.

This is a detail pane, that allows to inspect the executed SQLs.
Unfortunately, the pane is unusable, if there is an "IN" query with too many "?,?,?" parameters. As there are no spaces in that string, the browser cannot wrap that properly and you get an extrem long horizontal scroll bar

grafik

We use a browser plugin (FoxReplace) to replace "?," by "?,<SPACE>" as workaround. So that the browser can wrap the long queries.

So the result looks like this:
grafik
(Without spaces, this line is very long, causing horizontal scroll bars)

I'm still trying to figure out what the better approach would be 
1. replace `sb.append(",?")` generally by `sb.append(", ?")`,
2. Insert a space every 100th placeholder

Approach 1 will provide better wrappable queries, but would make the query string up to 33% longer. It will also break existing tests.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/platform/MultiValueBind.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/platform/MultiValueBind.java
@@ -61,7 +61,12 @@ public class MultiValueBind {
     }
     sb.append(" in (?");
     for (int i = 1; i < size; i++) {
-      sb.append(",?");
+      if (i % 100 == 0) {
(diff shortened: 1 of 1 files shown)


rbygrave:
I'm not really keen on making a JavaMelody UX presentation problem into an Ebean one.

Changing the sql interferes with Ebean's own monitoring approach - ebean insight.

rPraml:
Thanks for feedback. I'll try to fix that in JavaMelody

PR #2601, opened 2022-03-17
Extended @Column support for DB2

rPraml:
We add in this commit extended support for `@Column` for DB2 platform.
In DB2 we have the situation, that a `@Lob` has a default size of 1M, So we added the feature #2563 to add a platform specific definition.

This caused one issue, that unnecessary migrationscripts were created for platforms, that are not listed in `@Column`.
So `@Column(columnDefinition = "db2;clob(16K)")` -> `@Column(columnDefinition = "db2;clob(32K)")` should not generate migrations for other platforms than DB2 (fixed here: [DbMigration: Do only real type changes](https://github.com/ebean-orm/ebean/commit/74720a840d2241742c4254223bf03a3f5ccdf57f) =

We had also a problem, that it was not possible to use `@DbDefault` together with `@Column` because [dbColumDefault returned null](https://github.com/ebean-orm/ebean/commit/eb1ed0bdd0623e462850a553676de8af462243ca) in this case.

The next thing is, that we want to do complex definitions like `@Column(columnDefinition = "db2;clob(16K) inline length 500 compact;")` 
So this is valid for adding a new column, but not valid for an "alter column" statement (it must be done in two statements and things like 'compact' and 'logged' statements cannot be altered on exisiting columns.

We implemented a small parser that allow such statements both for column creation and for column alternation.
[Support for extra colum options for DB2](https://github.com/ebean-orm/ebean/commit/8f746196ab058e63c29d480c29c5bce64f859e09)


rbygrave:
Yes, looks nice.  Good stuff.

PR #2602, opened 2022-03-17
Handle lowerTableName/lowerColumnName

rPraml:
Hello Rob,

I'm wondering if PlatformDdl.lowerTableName/lowerColumnName are still required, as I think that will not work as intended.

I replaced them by a simple
`String lowerTableName(String name) { return name; }` and the SAME migration scripts were generated. 

Then, I replaced them by a 
`String lowerTableName(String name) { return "MARKER_" + name; }` and checked the migration scripts where the MARKER was missing.

In this PR I added a lot missing "lowerTableName" statements (I'm sure, I've not found all places, yet.)
There are also some edge-cases, where the tableName is used in trigger names or in User stored procedures.
How should we handle that? Especially, if the name contains quotes?

Note: There is also the `NamingConvention`, which makes everything lowercase by default. Do not mix up with `DbConstraintNaming`

As far as I understand, the intention of `DbConstraintNaming` was to handle quoted identifiers correctly.
But this is not (yet) used anywhere in the migration tests. And I think, it would not work for history tables and triggers

@rbygrave Maybe you can give me some feedback, if I should check DDL generation for missing lowerTableName/lowerColumnName statements
and add complexity or if you also think that we can remove them, as it will not have worked as intended in the past (at least in ebean 12)

cheers
Roland


rbygrave:
> I'm wondering if PlatformDdl.lowerTableName/lowerColumnName are still required, as I think that will not work as intended.

Yes, I think `PlatformDdl.lowerTableName() / lowerColumnName()` should be removed.

PR #2603, opened 2022-03-18
Refactored DDL generation to correctly handle quoted identifiers

rPraml:
This is a refactoring how DDL is generated, when quotedIdentifiers are used.

I've made an [ETable](https://github.com/ebean-orm/ebean/pull/2603/files#diff-0755fb46d84c05980a10e56229c4700780b342733c2b7f29f0e0742bbe12694d) entity, which uses a lot of reserved keywords.
This is something, that did not work for all platforms (especially MySql and SqlServer, as they use different quote chars than `"`)

I checked if the generated migration scripts will work for H2/MySql/DB2/SqlServer and Hana.

It prepares ebean generate working migration scripts, if `allQuotedIdentifiers=true` is set - this probably will also be an other workaround for `NONKEYWORDS=KEY,VALUE` in H2

To make this PR smaller, I've extracted some commits into separate PRs.
@rbygrave you can decide, if you either merge this PR or the smaller ones and I will update the PR next week


ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/ddlgeneration/platform/PlatformDdl.java
```diff
@@ -20,12 +20,17 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Controls the DDL generation for a specific database platform.
  */
 public class PlatformDdl {
 
+  // matches on pattern "check ( COLUMNAME ... )". ColumnName is match group 2;
+  private static final Pattern CHECK_PATTERN = Pattern.compile("(.*?\\( *)([^ ]+)(.*)");
```
rPraml, line 32: unfortunately, we must add also quotes to the check-patterns, if we want to support `allQuotedIdentifiers`.
So we have to parse `check ( "mycolumn" in ('A','B'))` and replace the quote chars by the platform specific.
Note: the check constraint is in the model.xml and is the same for all platforms

ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/model/MTable.java
```diff
@@ -574,11 +581,11 @@ public MColumn addColumnScalar(String dbColumn, String columnDefn) {
   /**
    * Add a 'new column' to the AddColumn migration object.
    */
-  private void diffNewColumn(MColumn newColumn) {
+  private void diffNewColumn(MColumn newColumn, MTable newTable) {
     if (addColumn == null) {
       addColumn = new AddColumn();
       addColumn.setTableName(name);
-      if (withHistory) {
+      if (newTable.isWithHistory()) {
```
rPraml, line 588: We must enable history only, when the new table has history.
if you disable history, then the old table has history, but not the new one

ebean-ddl-generator/src/test/resources/assert/ModelBuild_explicitSequencesTest/apply.sql
```diff
@@ -1,11 +1,11 @@
-create table persons (
-  id                            bigint generated by default as identity (start with 1000 increment by 40) not null,
-  surname                       varchar(64) not null,
-  name                          varchar(64) not null,
-  constraint pk_persons primary key (id)
+create table PERSONS (
+  ID                            bigint generated by default as identity (start with 1000 increment by 40) not null,
+  SURNAME                       varchar(64) not null,
+  NAME                          varchar(64) not null,
```
rPraml, line 4: @rbygrave due the removal of "lowerTableName", we will get uppercase names here.
I think, this is OK, because it is specified as `@Table(name = "PERSONS")` / ` @Column(name = "ID")`
If you want lower case names, specify them in the annotation

ebean-test/src/test/java/misc/migration/v1_0/ETable.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.persistence.Table;
+
+import io.ebean.annotation.DbComment;
+import io.ebean.annotation.History;
+import io.ebean.annotation.Index;
+
+@Table(name = "`table`")
+@Entity
+@History
+public class ETable {
```
rPraml, line 20: This is the "Reserved-keyword" entity. It uses a keyword for each fied and tries to address as many as possible features during DDL generation

ebean-test/src/test/java/misc/migration/v1_1/ETable.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+
+import io.ebean.annotation.DbComment;
+import io.ebean.annotation.History;
+import io.ebean.annotation.Index;
+
+@Table(name = "`table`")
+@Entity
+@History // FIXME: remove later
```
rPraml, line 19: @rbygrave It is currently impossible for some platforms to enable/disable history while there are pending drops.
(it will generate the wrong history table)
I don't mind this is a real limitation. I added a sanity check in Ddl-gneration

ebean-test/src/test/java/misc/migration/v1_2/ETable.java
```diff
@@ -0,0 +1,55 @@
...
+
+  // Note: This column should be removed, (as it is also not present in V1.0)
+  // There is a limitation in history generation, that you cannot enable history while you have pending drops
+  // History-table will be generyted in V1.3 - SQL without that column, while base table still contains that column
+  // When creating the "with_history" view, the DB complains, because the tables do not match.
+  // In V1.4 the column will be dropped in base table AND history table.
+  // This could be probably fixed, by generating history table also with dropped columns
+  @Column(name = "`select`")
+  @Index(unique = true)
+  private String select;
```
rPraml, line 47: Note to myself: This CAN be removed now, as long as previous version is under @History

PR #2605, opened 2022-03-21
DB2-DDL: Create history table with "create table as"

rPraml:
This PR is part of #2603 
For DB2, it is important, that the history table has the same columns in the same order as the base table.

if you append new columns, they will always go to the end of the current db-column list, but may not be in the same order in the java model. If you enable DbHistory later, then the history table may have a different column order.

We use `create table HISTORY_TABLE as (select * from BASE_TABLE) with no data` to create an exact copy of the existing table


rbygrave:
Part of merged #2603 so we can close this one.

PR #2606, opened 2022-03-21
Use correctly quoted historyTableName in DDL-History

rPraml:
This PR is part of #2603 and fixes History-Table generation if a quoted identifier is used


rbygrave:
Part of merged #2603 so we can close this one.

PR #2611, opened 2022-03-24
Fix javadoc generation

rPraml:
Hello Rob,
I noticed, that there are a lot of javadoc warnings during a maven build.
I do some investigation and noted, that it is currently impossible to use an `@` char in an `@{code ...}` block.

This is clearly a Javadoc-Bug, but it seems to exist since a long time, so I think, there will be no fix in the future (See: https://stackoverflow.com/questions/23812761/trying-to-escape-the-symbol-in-a-code-block-within-a-javadoc-comment-with)

To fix invalid javadoc (see https://ebean.io/apidoc/12/io/ebean/RowMapper.html) I changed them to `<code>`
There is also a missing dependency in `avaje-lang`, which causes `unknown enum constant When,MAYBE` warning.

Unfortunately, there are still a lot of files with invalid javadoc (mostly generated by xjc) which are not part of this PR

### DIFF
--- a/ebean-api/pom.xml
+++ b/ebean-api/pom.xml
@@ -39,6 +39,12 @@
       <artifactId>avaje-lang</artifactId>
       <version>1.0</version>
     </dependency>
+    
+    <dependency>
+      <groupId>io.avaje</groupId>
+      <artifactId>avaje-jsr305</artifactId>
+      <version>1.2</version>
+    </dependency>
 
     <dependency>
       <groupId>io.ebean</groupId>
--- a/ebean-api/src/main/java/io/ebean/BeanFinder.java
+++ b/ebean-api/src/main/java/io/ebean/BeanFinder.java
@@ -10,11 +10,11 @@ import java.util.Optional;
  * <p>
  * Note that typically users would extend BeanRepository rather than BeanFinder.
  * </p>
- * <pre>{@code
+ * <pre><code>
  *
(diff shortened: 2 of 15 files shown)


ebean-api/pom.xml
```diff
@@ -39,6 +39,12 @@
       <artifactId>avaje-lang</artifactId>
       <version>1.0</version>
     </dependency>
+    
+    <dependency>
+      <groupId>io.avaje</groupId>
+      <artifactId>avaje-jsr305</artifactId>
+      <version>1.2</version>
+    </dependency>
```
rPraml, line 47: This dependency has provided scope in avaje-lang.
@rbygrave please check, if we really want to have this dependency here or if it should be fixed in avaje-lang instead

rbygrave:
The javadoc bug is fixed in 15.

rbygrave:
No, we don't want these changes as it makes the example code in the javadoc harder to read.

If you bump you JDK to 15 or 17 then you won't see the javadoc warnings so that is what I'm suggesting you do or just ignore the warnings.

rPraml:
Thanks, I ignore the errors
(We recently made the jump to 11 now. It will probably be a while before we can go to 15 in our company. ;) )

rbygrave:
Ha ha. I note that there are some pretty impressive improvements to GC in 18 which has just gone GA :)

PR #2616, opened 2022-03-28
PersistVisitor: a method to validate data that should be saved

rPraml:
We use this in our application to validate beans before they will be saved.

`DB.visitSave(bean, visitor)` will call the PersistVisitor for each bean that is dirty and will be saved.


rbygrave:
I don't think this is the right long term path in terms of validation.

rPraml:
I'll close this issues

PR #2617, opened 2022-03-28
ENH: JsonContext accepts a bean as target

rPraml:
We have the following use case, that we have a json string and we want to apply this to a given bean.

e.g.
customer = DB.find(Customer.class, 1);
json=`{"name":"Roland"}`

Now we want to apply this JSON to the customer and save it again.

This extends the JsonContext with `toBean(target, json)` method, where `target` could be an instance instead of a class.

cheers
Roland


rbygrave:
Nice !!

Issue #2374, opened 2021-09-15
Proposal: Move from Java 8 to Java 11 - March 2022


rbygrave:
# 1. When to move Ebean on from Java 8?
# 2. What should the next version be?  Java 11?

**Q**: Can Ebean fully support java module system (module-info) and stay on Java 8
_Rob_: Yes, we have a PR for that and I'll merge that in for version 12.12.0

**Q**: What is the motivation for moving on from compiling to Java 8?
_Rob_: There is no super strong immediate need per say. If everyone has moved to Java 11 then I feel we could move and take some small but nice little improvements to JDK API, bring the module-info into src/main/java (which makes it enforced at compile time/better when refactoring internals).

My question is, who is still locked into Java 8 and what is the reason for not getting to at least Java 11?

jnehlmeier:
> My question is, who is still locked into Java 8 and what is the reason for not getting to at least Java 11?

Probably only very slow moving enterprises. But these usually hesitate upgrading libraries as well if there is no immediate reason to upgrade.

vnnv:
I am working on enterprise projects using ebean and the reason I am still using java8 is a legacy frontends written in Vaadin 7. We are currently rewriting the frontends in angular and when we are ready we are going to switch to java 11 or even java 17. Hopefully this will be done till the end of March 2022.

vladimirfx:
+1 For Java 11

We also have a few legacy (actually deps frozen) projects. For it, Ebean can not be updated as well as Java. So no reason to stay on Java 8 for legacy projects.

rbygrave:
Just checking in, we are 1 month away from a move to Java 11. All good to go with Java 11 @vnnv @nedge and everyone else?

vnnv:
I hope we'll manage to switch in a month. In case we did not make it, will keep older ebean version until we migrate all our codebase. So I think we are good to go :)

rbygrave:
Merged into master branch. Closing this now.

PR #2622, opened 2022-03-30
#2374 - Bump to Java 11 and module-path

rbygrave:
In terms of module-path, changing to enforced at compile time src/java/module-info.java from the multi-release jars and moditech plugin


rbygrave (commented):
all good.

PR #2654, opened 2022-04-14
Fix: Setting file with same content should not make bean dirty

rPraml:
If you set a file with the same file content, the bean gets dirty, but should not, because no update would be required

### DIFF
--- a/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
+++ b/ebean-api/src/main/java/io/ebean/bean/EntityBeanIntercept.java
@@ -6,6 +6,11 @@ import io.ebean.ValuePair;
 
 import javax.persistence.EntityNotFoundException;
 import javax.persistence.PersistenceException;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.Serializable;
 import java.math.BigDecimal;
 import java.net.URL;
@@ -910,7 +915,36 @@ public final class EntityBeanIntercept implements Serializable {
       // use the string format to determine if dirty
       return !obj1.toString().equals(obj2.toString());
     }
+    if (obj1 instanceof File && obj2 instanceof File) {
+      File file1 = (File) obj1;
+      File file2 = (File) obj2;
+      if (file1.exists() && file2.exists() && file1.length() == file2.length()) {
+        return notEqualContent(file1, file2);
+      }
+    }
     return !obj1.equals(obj2);
+  }
+
+  private static boolean notEqualContent(File file1, File file2) {
+    try (InputStream is1 = new FileInputStream(file1); InputStream is2 = new FileInputStream(file2)) {
+      byte[] buf1 = new byte[16384];
+      byte[] buf2 = new byte[16384];
+      int len1;
+      int len2;
+      while ((len1 = is1.read(buf1)) != -1 && (len2 = is2.read(buf2)) != -1) {
+        if (len1 != len2) {
+          return true;
+        }
+        if (!Arrays.equals(buf1, buf2)) {
+          // it does not matter, if we compare more than len1/len2 as the remainig
+          // bytes in the buffers are either 0 or equals from the prev. loop.
+          return true;
+        }
+      }
+      return false;
+    } catch (IOException e) {
+      return true; // handle them as "not equals"
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


rbygrave:
... just pondering the case where the file is very very large (it can be arbitrarily large) and question if we have introduced a serious cost here of comparing every byte until it finds a difference.  So this would not be great per say when the file is exactly the same size and when the change occurs very close to the end of the file. Hmmm.

rPraml:
I understand your concerns, on the other hand, how much IO would you save if the property doesn't get dirty and the file isn't even transferred to the DB (yes, this depends on the use case, it happens quite often in our code)
Compare-speed on an SSD:  ~ 100 MB/sec (estimated)
Write-speed to DB: ~ 10 MB/sec (estimated)

Maybe this can be handled optionally with mutation detection (I am thinking of MutationDetection.HASH or similar)

Anyway, I have a question for you about file handling in Ebean.
I don't quite understand who you think should clean up all the temp files. (You may remember, we are using the [TempFileProvider](https://github.com/ebean-orm/ebean/pull/1593) in our fork, which does the job)

Files are something special, as the disk content is not automatically garbage collected as other properties (this is what the TempFileProvider tries to address) To me it looks like File and the ScalarTypeFile currently doesn't fit very good into the Ebean ecosystem. IMHO we need something, where the temp files are registered, so that they are safely deleted. 
E.g. the TempFileProvider (could be also implemented as UserSessionTempFileProvider, which deletes files if the user session terminates) or after transaction or something like a 'fileService'. Please see the following problem scenarios, we have in our application:

**Problem 1**
When I fetch an entity and call `getContent()`, Ebean creates a tempfile but doesn't delete it.
Of Course, the application code can delete the file after use. But what happens, if I call `getContent()` again on that entity. I get a File reference, to a deleted file, or the application has to delete the file AND unload the property, so that the file gets loaded again on next `getContent()`
```java
MyEntity entity = DB.find(MyEntity.class, 1);
File file = entity.getContent();
... do something with file
file.delete(); // TODO: also unload property?

MyEntity entity2 = DB.find(MyEntity.class, 1); // may come from persistence cache
File file2 = entity2.getContent(); // will crash, as file is deleted
```
**Problem 2**
When I set a file, ebean compares the new file with the old value. It loads the old value - into a temp file. Who should delete that temp file? The application can't, because it does not know the name
```java
MyEntity entity = DB.find(MyEntity.class, 1);
entity.setContent(file); // will call getContent() and call "areEqual" - there is no chance to get the temp file here
// possible workaround: call entity.getContent() first
```
**Problem 3**
When I set a file and want to save it, I've to wait with deletion until ebean does not need the file any more. (e.g. after save)
```java
MyEntity entity = new MyEntity();
entity.setContent(file);
DB.save(entity);
file.delete(); // Now the file is not needed any more - but wait! What if we are using batching?
```
You see, you can run into problems, as soon as batching is used.


@rbygrave I had a few thoughts on a 'blob service', but that doesn't seem mature to me either:
```java
/**
 * Will retrieve the content as stream. This will effectively pass through ResultSet.getBinaryStream 
 * and will 're-fetch' the stream from the database on each call
 */
DB.blob(entity, "content").getStream();
// In many cases, we will need a stream (e.g. a download button will stream the data directly to ServletOutputStream), so instead of doing 'DB->file->ServletOutputStream' it would be better to fetch it directly from DB. (Note: Read restrictions in https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#getBinaryStream-int-) 

/**
 * Will retrieve an individual file, that HAS to be deleted. Multiple calls will give you multiple files. (Effectively, it will call ResultSet.getBinaryStream()
 */
DB.blob(entity, "content").getFile()

/**
  * Will register a stream. This stream is then consumed when entity is saved and effectively passed to PreparedStatement.setBinaryStream
  */
DB.blob(entity, "content").setStream(inputStream, length)

```
The "stream" thing looks nice to me, but a stream can only be read once, so we will need a file buffer (or off heap buffer) as soon as we are using features like mutation detection / Jackson / setStream & getStream on the same property or similar.

rbygrave:
> problem 1

For myself, "do something with file" often means do a file **_move_**.  That is, there is some local (to the app server) cache of file content.  When we are fetching the file from the db, we are effectively populating that local cache. So "do something with file" is often ... "move the file content to some directory that represents a local cache of file content".

> problem 2

Well, a stateless update would be really the better approach when we want to update the [relatively large] content generally (with the thought that we are not looking to compare content to see if it's changed) so that is:
```java
MyEntity entity = new MyEntity();
entity.setId(1);
entity.setContent(file);
entity.setVersion(...); // set the version if we want OCC check
DB.update(entity);
```

> problem 3

Use transaction.flush() as needed when using jdbc batch?

PR #2664, opened 2022-04-20
Fix: CompareTo when discriminatorString is null

NSzemenyei:
Hello @rbygrave ,

we found a problem in our application with ebean 13.
In case of starting the database, all the entity classes will be scanned and the class-tree will be calculated (`deployMap in DeployInherit`). In some cases any classes wont't be added as a `children` to it's parents.

@rPraml and I debugged it and found the cause: in `DeployInheritInfo` line 87 children will be added to the `children TreeMap`, and the method `compareTo(DeployInheritInfo o)` will be also called. When `discriminatorStringValue` is null for the two classes, the two elements will be found as the "same" and the new element won't be added to the `TreeMap`.

Can you please check it?

Kind regards,
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/DeployInheritInfo.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/DeployInheritInfo.java
@@ -260,14 +260,14 @@ public final class DeployInheritInfo implements Comparable<DeployInheritInfo> {
 
   @Override
   public int compareTo(DeployInheritInfo o) {
-    if (o == this || o.discriminatorStringValue == null && discriminatorStringValue == null) {
+    if (o == this) {
       return 0;
     } else if (o.discriminatorStringValue == null) {
       return 1;
     } else if (discriminatorStringValue == null) {
       return -1;
     } else {
-      return discriminatorStringValue.compareTo(o.discriminatorStringValue);
+      return type.getName().compareTo(o.getType().getName());
     }
   }
 


rPraml:
FYI, discriminator may be null when you have children/parents/grand parents, e. g 
C1 extends AbstractB1 extends AbstractA and
C2 extends AbstractB2 extends AbstractA
B1 and B2 do not have a discriminator value (but are no root nodes) , so they are considered as "equals" for the treeMap

ebean-core/src/main/java/io/ebeaninternal/server/deploy/parse/DeployInheritInfo.java
```diff
@@ -260,14 +260,14 @@ public String toString() {
 
   @Override
   public int compareTo(DeployInheritInfo o) {
-    if (o == this || o.discriminatorStringValue == null && discriminatorStringValue == null) {
+    if (o == this) {
       return 0;
     } else if (o.discriminatorStringValue == null) {
       return 1;
     } else if (discriminatorStringValue == null) {
```
rPraml, line 267: @NSzemenyei discriminatorValue should be removed completely here

rbygrave:
> check it?

Pondering the thought ... "we really should try to create the failing test _first_" ... which isn't always easy but the mindset I try to get into is - _fail first, then fix_.  The existence of the test also proves that we have created something that is _testable_ plus provides a starting point for someone in the future to add another test case when we later on identify that we missed a scenario.

_"Regression"_ should also be used as a trigger for the thought - _"oh, I better add a test ... "_

For myself, I am far far from perfect in this mindset and approach ... but I keep trying to get better :)

rbygrave:
Replacing this PR with https://github.com/ebean-orm/ebean/pull/2666

rbygrave:
Closing, replaced by #2666

PR #2669, opened 2022-04-26
FIX: StackOverflowException when using recursive relation within Map

NSzemenyei:
Hello @rbygrave ,

using ebean 13 and ebean-agent 13.X we got a StackOverflowException by calling toString for Map properties.
@jonasPoehler and I created a simple unittest to simulate the exception. We also implemented a solution for the problem.
Can you check it?

Kind regards
Noemi


rbygrave:
Very nice!!

Confirmed with:
- Pull branch
- run test -> pass, 
- comment out the fix + run test again -> fail with stack overflow as expected

Good stuff, well done. Thanks.

PR #2693, opened 2022-05-16
DDL - Add support for creating schema via database migrations (and create all ddl)

rbygrave:
Initially for H2 and Postgres, with commented out DDL for other platforms.


rbygrave (commented):
LGTM

PR #2706, opened 2022-05-24
ENH: Add support for findSingleAttributeSet()

nPraml:
Hello @rbygrave ,

in order to use the faster computation and operations of a java 'Set' instead of 'List' in our application, @rPraml and I implemented the findSingleAttributeSet() method.
Can you please check it?

Kind regards
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/SpiQuery.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/SpiQuery.java
@@ -93,6 +93,11 @@ public interface SpiQuery<T> extends Query<T>, SpiQueryFetch, TxnProfileEventCod
     ATTRIBUTE(FIND_ATTRIBUTE, "findAttribute", false, false),
 
     /**
+     * Find single attribute set.
+     */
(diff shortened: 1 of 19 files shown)


ebean-core/src/main/java/io/ebeaninternal/api/SpiQuery.java
```diff
@@ -92,6 +92,11 @@ enum Type {
      */
     ATTRIBUTE(FIND_ATTRIBUTE, "findAttribute", false, false),
```
rbygrave, line 93: This all looks fine.

I'm putting a comment here because it looks like we overload this exisint ATTRIBUTE to mean both findSingleAttribute and findSingleAttributeList ... and that is ok only because we don't have query caching for findSingleAttribute [only on findSingleAttributeList].

So this comment is here because I'm noting that. PR looks good.

rbygrave:
So yes, all looks great.  I'll do a double look later and then merge this in.

Side note: Down the track we could look at `ATTRIBUTE` being used for both findSingleAttribute and findSingleAttributeList.

PR #2708, opened 2022-05-31
Part 1: Fix read/write/query LocalTime values

nPraml:
Hello @rbygrave ,

we decided to split our big refactoring PR https://github.com/ebean-orm/ebean/pull/2443 in smaller ones.
This PR would be the first which tests the write & read & query of the LocalTime values.

In the Javadoc I noticed how it can be tested in different java time zones with -Duser.timezone.

Can you take a look at this PR?

Kind regards
Noemi

----
## Summary of this change
This changes how `Time` type is read (ResultSet) and bound (PreparedStatement) when there is a time zone set via `DatabaseConfig`.

Previously, `Time` was using the time zone method for `Timestamp types` rather than `Date types` and this changes it to use the method for `Date types` (which is actually null for everything apart from MySql).

Effectively no change for MySql.
Effectively for non-MySql, the `Time` types where incorrectly using the timezone if specified and this change means they no longer use that configured time zone.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/type/DataBind.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/type/DataBind.java
@@ -158,7 +158,7 @@ public class DataBind implements DataBinder {
 
   @Override
   public final void setDate(java.sql.Date v) throws SQLException {
-    Calendar timeZone = dataTimeZone.getDateTimeZone();
+    Calendar timeZone = dataTimeZone.getTimeComponentTimeZone();
     if (timeZone != null) {
       pstmt.setDate(++pos, v, timeZone);
     } else {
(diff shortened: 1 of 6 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/type/DataBind.java
```diff
@@ -178,7 +178,7 @@ public final void setTimestamp(Timestamp v) throws SQLException {
 
   @Override
   public final void setTime(Time v) throws SQLException {
-    Calendar timeZone = dataTimeZone.getTimeZone();
+    Calendar timeZone = dataTimeZone.getTimeComponentTimeZone();
```
rPraml, line 181: Here...

ebean-core/src/main/java/io/ebeaninternal/server/type/RsetDataReader.java
```diff
@@ -156,7 +156,7 @@ public final String getString() throws SQLException {
 
   @Override
   public final Time getTime() throws SQLException {
-    Calendar cal = dataTimeZone.getTimeZone();
+    Calendar cal = dataTimeZone.getTimeComponentTimeZone();
```
rPraml, line 159: ... and here is the fix. I suggested to @nPraml to rename that method.

PR #2722, opened 2022-06-17
Migrate to Jakarta JAXB (jakarta.xml.bind <- javax.xml.bind)

rbygrave:
We use JAXB & XML for DDL generation and running DDL in test, and also at runtime for ebean-externalmapping-xml (which I suspect is only rarely used these days, it's now an optional dependency that is not included by default ).

The main uses of JAXB should all be in the test classpath only. Still, this migrates from using `javax.xml.bind` to `jakarta.xml.bind` so probably need to assess how disruptive that would be. Will people using javax.xml.bind for other things get messed up?  In theory they should not.

### Moving to jakarta.xml.bind
```xml
    <dependency>
      <groupId>jakarta.xml.bind</groupId>
      <artifactId>jakarta.xml.bind-api</artifactId>
      <version>4.0.0</version>
    </dependency>
    <dependency>
      <groupId>com.sun.xml.bind</groupId>
      <artifactId>jaxb-impl</artifactId>
      <version>4.0.0</version>
      <scope>runtime</scope>
    </dependency>
```

### From ...

```xml
    <dependency>
      <groupId>org.glassfish.jaxb</groupId>
      <artifactId>jaxb-runtime</artifactId>
      <version>2.3.6</version>
    </dependency>
```

### DIFF
--- a/ebean-autotune/pom.xml
+++ b/ebean-autotune/pom.xml
@@ -32,10 +32,15 @@
 
     <!-- needed for java 11+ -->
     <dependency>
-      <groupId>org.glassfish.jaxb</groupId>
-      <artifactId>jaxb-runtime</artifactId>
-      <version>2.3.6</version>
-      <scope>provided</scope>
+      <groupId>jakarta.xml.bind</groupId>
+      <artifactId>jakarta.xml.bind-api</artifactId>
+      <version>4.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <version>4.0.0</version>
+      <scope>runtime</scope>
     </dependency>
 
     <dependency>
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/Autotune.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/Autotune.java
@@ -1,9 +1,9 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlAccessType;
-import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlType;
+import jakarta.xml.bind.annotation.XmlAccessType;
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
 import java.util.ArrayList;
 import java.util.List;
 
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ObjectFactory.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ObjectFactory.java
@@ -1,6 +1,6 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlRegistry;
+import jakarta.xml.bind.annotation.XmlRegistry;
 
 
 /**
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/Origin.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/Origin.java
@@ -1,10 +1,10 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlAccessType;
-import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlAttribute;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlType;
+import jakarta.xml.bind.annotation.XmlAccessType;
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlAttribute;
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
 
 
 /**
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileDiff.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileDiff.java
@@ -1,9 +1,9 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlAccessType;
-import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlType;
+import jakarta.xml.bind.annotation.XmlAccessType;
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
 import java.util.ArrayList;
 import java.util.List;
 
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileEmpty.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileEmpty.java
@@ -1,9 +1,9 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlAccessType;
-import javax.xml.bind.annotation.XmlAccessorType;
-import javax.xml.bind.annotation.XmlRootElement;
-import javax.xml.bind.annotation.XmlType;
+import jakarta.xml.bind.annotation.XmlAccessType;
+import jakarta.xml.bind.annotation.XmlAccessorType;
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
 import java.util.ArrayList;
 import java.util.List;
 
--- a/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileNew.java
+++ b/ebean-autotune/src/main/java/io/ebeaninternal/server/autotune/model/ProfileNew.java
@@ -1,9 +1,9 @@
 package io.ebeaninternal.server.autotune.model;
 
-import javax.xml.bind.annotation.XmlAccessType;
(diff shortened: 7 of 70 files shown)


rbygrave:
Interesting, 1K downloads a month for `ebean-externalmapping-xml` (a lot higher than I was expecting? plus a lot in 13.x version range) vs 18K for `ebean-externalmapping-api`. [and 80k/month for ebean] 

Ok people, if you are one of those using `ebean-externalmapping-xml` I'd love to hear if you think a migration to `jakarta.xml.bind` is going to mess with your mojo.

rPraml:
Hello @rbygrave Our (dev) application creates migration scripts on startup.

Today, we noticed that we've an issue with this change, as other frameworks (like Spring 2.7.x, [Sardine](https://github.com/lookfirst/sardine) [PdfBox](https://github.com/apache/pdfbox)) still use javax.xml.bind and it seems that they are not yet prepared for jakarta.xml.bind.

We also have some jaxb-models (which we can migrate) 

I have to investigate, what is the exact impact to go to jakarta.xml.bind (do you know, if I can have them parallel?)

But for now I tend to revert this change in our branch

rbygrave:
> if I can have them parallel?

I believe you can.

rbygrave:
Let me know what errors you saw and what you find. If this impacts lots of users in ways I wasn't expecting then we might be forced to revert this.

rPraml:
We have built 2 versions now:

- 13.7.1-FOC1 is the version with jakarta
- 13.7.1-FOC2 is the version with javax (reverted #2722)

in our depencency tree, we've glassfish 2.3.6, as depencency, which also pulls jakarta.xml.bind:jakarta.xml.bind-api:2.3.3

```
[INFO] |  +- org.glassfish.jaxb:jaxb-runtime:jar:2.3.6:compile
[INFO] |  |  +- jakarta.xml.bind:jakarta.xml.bind-api:jar:2.3.3:compile
[INFO] |  |  +- org.glassfish.jaxb:txw2:jar:2.3.6:compile
[INFO] |  |  \- com.sun.istack:istack-commons-runtime:jar:3.0.12:compile
```

While our ebean version with javax pulls only these dependencies (because glassfish & jakarta.xml.bind-api is already pulled)
```
[INFO] |  |  \- io.ebean:ebean-test:jar:13.7.1-FOC2:test
[INFO] |  |     +- io.ebean:ebean-ddl-generator:jar:13.7.1-FOC2:test
[INFO] |  |     \- io.ebean:ebean-test-containers:jar:6.2:test
```

the ebean version which depends on jakarta pulls these additional dependencies:
```
[INFO] |  |  \- io.ebean:ebean-test:jar:13.7.1-FOC1:test
[INFO] |  |     +- io.ebean:ebean-ddl-generator:jar:13.7.1-FOC1:test
[INFO] |  |     +- io.ebean:ebean-test-containers:jar:6.2:test
[INFO] |  |     \- com.sun.xml.bind:jaxb-impl:jar:4.0.0:test                          <-- NEW
[INFO] |  |        \- com.sun.xml.bind:jaxb-core:jar:4.0.0:test                       <-- NEW
[INFO] |  |           \- org.eclipse.angus:angus-activation:jar:1.0.0:test            <-- NEW
[INFO] |  |              \- jakarta.activation:jakarta.activation-api:jar:1.2.2:test  <-- NEW
```

When running the test, we get a `NoClassDefFoundError: jakarta/xml/bind/JAXBException` (See stacktrace below)

If I add 
```
    <dependency>
      <groupId>jakarta.xml.bind</groupId>
      <artifactId>jakarta.xml.bind-api</artifactId>
      <version>4.0.0</version>
    </dependency>
```
to have it in the class path, I cannot compile the code any more, as it overrides the 2.3.x depencency from glassfish.
(and we have code, that imports "javax.xml.bind")
So what I would need is to have both versions in class path (but this does not work, as they have the same name space)

Jonas and I discovered, that adding the depencencies
```
		<dependency>
			<groupId>jakarta.xml.bind</groupId>
			<artifactId>jakarta.xml.bind-api</artifactId>
			<version>4.0.0</version>
		</dependency>
		<dependency>
			<groupId>javax.xml.bind</groupId>
			<artifactId>jaxb-api</artifactId>
			<version>2.3.1</version>
		</dependency>
		<dependency>
			<groupId>jakarta.activation</groupId>
			<artifactId>jakarta.activation-api</artifactId>
			<version>2.1.0</version>
		</dependency>
```
This would solve the compile problem and it seems, that also all tests will pass.

I investigated also to check other dependencies. While a lot of classes (like (Base64Util)[https://github.com/apache/logging-log4j2/blob/release-2.x/log4j-api/src/main/java/org/apache/logging/log4j/util/Base64Util.java#L33]) are prepared for Java 11
we have still older dependencies like PdfBox2 https://github.com/apache/pdfbox/blob/2.0/xmpbox/src/main/java/org/apache/xmpbox/DateConverter.java#L382 that rely on the older API
And of course, Sardine (and maybe other dependencies) use the older API https://github.com/lookfirst/sardine/blob/master/src/main/java/com/github/sardine/DavAce.java#L25


Possible migration plans
------------------------
- Keep both (javax & jakarta) libraries in class path and migrate our own code
- Update dependent projects (Sardine/PdfBox/...) to the latest version and hope that they do not use javax.xml bind
- Identify other problems
- There is also an option to use a java-agent that does the translation: https://stackoverflow.com/questions/67592237/create-a-jar-file-that-supports-java-and-jakarta

Nevertheless going this step NOW is a bit risky for us, as our application is already productive. So we've to do a lot of tests.
This means not, that Ebean should not go the transition now, as we currently have our own version and it is not too difficult to revert that particular change.
I would suggest, you should wait for feedback from other people if they have similar problems.

Cheers
Jonas & Roland


Relevant stack trace
```
Caused by: java.lang.NoClassDefFoundError: jakarta/xml/bind/JAXBException
        at io.ebeaninternal.dbmigration.model.MigrationResource.read(MigrationResource.java:42)
        at io.ebeaninternal.dbmigration.model.MigrationModel.readMigrations(MigrationModel.java:57)
        at io.ebeaninternal.dbmigration.model.MigrationModel.read(MigrationModel.java:36)
        at io.ebeaninternal.dbmigration.DefaultDbMigration$Request.<init>(DefaultDbMigration.java:580)
        at io.ebeaninternal.dbmigration.DefaultDbMigration.createRequest(DefaultDbMigration.java:557)
        at io.ebeaninternal.dbmigration.DefaultDbMigration.generateMigrationFor(DefaultDbMigration.java:370)
        at io.ebeaninternal.dbmigration.DefaultDbMigration.generateMigration(DefaultDbMigration.java:329)
        at io.ebeaninternal.dbmigration.DbMigrationPlugin.online(DbMigrationPlugin.java:31)
        at io.ebeaninternal.server.core.DefaultServer.executePlugins(DefaultServer.java:256)
        at io.ebeaninternal.server.core.DefaultContainer.initServer(DefaultContainer.java:149)
        at io.ebeaninternal.server.core.DefaultContainer.createServer(DefaultContainer.java:109)
        at io.ebeaninternal.server.core.DefaultContainer.createServer(DefaultContainer.java:29)
        at io.ebean.DatabaseFactory.createInternal(DatabaseFactory.java:136)
        at io.ebean.DatabaseFactory.create(DatabaseFactory.java:85)
        at de.foconis.core.api.app.DbFactory.postConstruct(DbFactory.java:120)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleElement.invoke(InitDestroyAnnotationBeanPostProcessor.java:389)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleMetadata.invokeInitMethods(InitDestroyAnnotationBeanPostProcessor.java:333)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor.postProcessBeforeInitialization(InitDestroyAnnotationBeanPostProcessor.java:157)
        ... 84 common frames omitted
Caused by: java.lang.ClassNotFoundException: jakarta.xml.bind.JAXBException
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
```

rbygrave:
Another option:

Copy/fork `jakarta.xml.bind:jakarta.xml.bind-api:4.0.0` and release it with different maven co-ordinates like say `io.avaje:avaje-jaxb-api:4.0.0`

In this way it could live along side the `2.3.3` version of the dependency (that has `javax.xml`).


Dang, this looks to me like it has the potential to be a serious mess for people (given they used the same maven coordinates).

rbygrave:
My current expectation is that we need to revert this pending the discussion at: https://github.com/eclipse-ee4j/jaxb-api/issues/263

rPraml:
@rbygrave FYI other modules like https://logback.qos.ch/news.html maintain two versions:

> 2022-08-28 Release of logback versions 1.3.0 and 1.4.0
> 
> Logback 1.3.x supports the Java EE edition whereas logback 1.4.x supports Jakarta EE, otherwise the two versions are feature identical.
> 
> Given that downstream users are likely to depend on either Java EE or (in the javax namespace) or on Jakarta EE (in the jakarta namespace) in their projects, it was deemed important for logback to support both alternatives.
> 
> As such, logback 1.3.x supports the Java EE edition whereas logback 1.4.x supports Jakarta EE, otherwise the two versions are feature identical. Here is a summary of the differences

Maybe the same will be an option for the ddl-generator, autotune and externalmapping module.
(I wonder if there is a simple automatism that can do this conversion on each build)

rbygrave:
Yeah, its a possibility but to me I'd suggest it is not the preferred option at this stage. 

For example, I maintain 2 branches for avaje-inject to support `javax.inject` and `jakarata.inject` but that isn't an issue because avaje-inject is (A) super fast to release and (B) at a really nice boring stable phase now so it's released very infrequently.

With Ebean we approximately release every 2 weeks and I think maintaining two versions at the moment would be a pain. My current thought is to (A) Wait for folks to demand support for `jakarta.xml.bind` and (B) Investigate moves to alternate xml binding libraries.

There is also the chance that apps don't actually migrate to `jakarta.xml.bind` for years.

Issue #2727, opened 2022-06-29
Stateless update on one-to-many relationship with children soft delete results in a hard delete instead of only inserting new child


Ichtil:
## Expected behavior
Domain has 3 entities - Goods, Workflow (softdelete), WorkflowOperation (softdelete).

Relationships are 
- Goods with one-to-one relationship to Workflow (orphanRemoval and cascade.all)
- Workflow has one to many relationship to WorkflowOperation (orphanRemoval, and cascade.all)

When adding Operation to Workflow and saving Goods instance, we expect to see insert of the new operations (with possibly several updates). No delete should be issued on existing operation (and even if it was issued, it should have been soft delete). 

## Actual behavior

There is a hard delete (`delete from`) on existing Operation

The update will throw exception if there is one-to-many cascade.all relation on workflow_operation_entity that is still be referenced from other table - the delete is not cascaded (this error is not showcased in the test as I believe it shouldn't be an issue in the first place).

### Steps to reproduce
Please see attached PR with unit test.

```java
@Entity
public class GoodsEntity extends BaseDomain {
    private String name;

    @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
    private WorkflowEntity workflowEntity;
```

```java
@Entity
public class WorkflowEntity  extends BaseDomain {
    private String revision;

    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
    @JoinColumn(name = "workflow_id")
    private List<WorkflowOperationEntity> operations = new ArrayList<>();

    @SoftDelete
    private boolean deleted;
```

```java
@Entity
public class WorkflowOperationEntity extends BaseDomain {

  private String name;

  @ManyToOne
  @JoinColumn(name = "workflow_id")
  private WorkflowEntity workflowEntity;

  @SoftDelete
  private boolean deleted;
```

```java
    var goods = new GoodsEntity();
    var workflow = new WorkflowEntity();
    var operation1 = new WorkflowOperationEntity();
    goods.setWorkflowEntity(workflow);
    workflow.setOperations(List.of(operation1));

    DB.save(goods);

    LoggedSql.start();

    // statelessly add another operation to the workflow and save goods
    var goodsStateless = new GoodsEntity();
    goodsStateless.setId(1L);
    var workflowStateless = new WorkflowEntity();
    workflowStateless.setId(1L);
    var operation1Stateless = new WorkflowOperationEntity();
    operation1Stateless.setId(1L);
[...]

PR #2728, opened 2022-06-29
Testcase demonstrating stateless update/insert on a child resulting in a delete of existing

Ichtil:
Attached issue #2727


rbygrave:
So something that is possibly misleading in this test is that the beans only have the id values populated (vs having other properties like name or version etc populated). This is misleading because a bean with only the id populated is the same as a reference bean.

I'll adjust the test a little so that the existing operation isn't confused as a reference bean.

There looks to be a bug in here, around the delete bug let me tweak this test a little first.

Issue #2746, opened 2022-07-13
Problems with findSingleAttributeList - queries and oneToMany relations


AlexWagner:
Hi there.
I added some tests for findSingleAttributeList in Class TestQuerySingleAttribute.
Focus are queries on oneToMany relations (Query on customer, oneToMany-Field: orders.status)

some queries don't work as they produce invalid sql
some queries deliver wrong result

See PR in our fork: https://github.com/FOCONIS/ebean/pull/68

rbygrave:
I'd ask to focus on 1 thing / 1 failing test and generally include an assert on the expected SQL and for a case where the SQL is not what is expected put a comment in for what the invalid SQL produced is.  

Can you pick one failing case and produce a PR for that?

rbygrave:
I think we have resolved those now. I'll close this.

PR #2754, opened 2022-07-20
Bug findsingleattributelist o2m notNull

AlexWagner:
See Issue #2746


rbygrave:
Please review the change to the test query to make use of **_filterMany()_**

PR #2757, opened 2022-07-20
Add notIn(Query) to query beans

tobias-:
Missing from query beans, but is available in the `ExpressionList` class.


rbygrave:
Cool - thanks !!

PR #2763, opened 2022-07-28
Fix: toJson must use setIntercept if target bean is passed

rPraml:
This is a bug introduced with #2617

When an existing entity is passed, toJson must use the intercept methods, otherwise the bean may not recognize updates and does not get dirty/saved


rbygrave:
Nice tests. Its a pretty cool feature really!!

PR #2764, opened 2022-07-28
Fix: DDL will get wrong hint for special migration

rPraml:
This is uncritical, as only a comment in DDL may be produced
(probably a regression of #2721)


rbygrave:
Cool.  All the little things add up.

Issue #2766, opened 2022-08-01
Reference bean created from JSON can invoke lazy loading by default (didn't prior to 12.13.1)


rbygrave:
Prior to 12.13.1 https://github.com/ebean-orm/ebean/pull/2440 ... a reference bean created by deserializing JSON didn't invoke lazy loading.

Since #2440, the behavior has changed such that it does invoke lazy loading.
See https://github.com/ebean-orm-examples/example-minimal/pull/9

Q: Should we put the behavior back such that it does not invoke lazy loading without the explicit use of JsonReadOption ?
A: I think so

rbygrave:
@fyrnaga - you can check that this failing test matches your case

Hi @jonasPoehler @nPraml, do you want to look at this in terms of impact of putting the behaviour back to the original which is that reference beans created from json do not invoke lazy loading.

rbygrave:
@rPraml FYI, this reverts the change in behavior that came in with #2440

xabolcs:
Hi Rob!

Will this (revert / behavior change) be noted somehow in the 12.x versions?

rbygrave:
> be noted somehow in the 12.x versions?

What do you mean?  Do you mean patch 12.x?

We didn't pick up that we had a behavior change that occurred back in 12.13.1 (via #2440) until now. Prior to 12.13.1 there would not have been lazy loading occur unless it was explicitly requested via JsonReadOptions.  After 12.13.1 the behavior changed such that reference beans would lazy load. We have only just noted that unfortunately ... and now the question is if we should revert the behavior back to the original behavior or not.

I'm suggesting we revert back because:
- That was the original intent, that no beans created from json would lazy load by default
- We have JsonReadOptions to give us explicitly the ability to enable lazy loading on those beans if desired

xabolcs:
Hi!

Yes, my first thought was to backporting to 12.x. But I'm not an Ebean user, just an Ebean reader.

Noting this regression in 12.13.1's release notes would be more than enough for me.

fyrnaga:
> @fyrnaga - you can check that this failing test matches your case
> 
> Hi @jonasPoehler @nPraml, do you want to look at this in terms of impact of putting the behaviour back to the original which is that reference beans created from json do not invoke lazy loading.

Hi @rbygrave, it seems indeed to match our situation.

PR #2767, opened 2022-08-01
#2766 Explicitly specify disableLazyLoad on beans created from JSON by default 

rbygrave:
- Beans created from JSON default to disableLazyLoad true
- Explicitly use new JsonReadOptions().setEnableLazyLoading(true) when lazy loading is needed


rbygrave:
@nPraml  @jonasPoehler @rPraml  ... this is a behavior change (revert to prior behavior). Are you ok with this being merged?

rPraml:
Hi Rob, yes it's OK, I think it would not break anything here, because we have disabled lazy load in most cases anyway.
(If we deserialize references, we mostly do this in a 'backup & restore' use case)

rPraml (approved):

PR #2770, opened 2022-08-02
Add SqlQuery usingTransaction() for consistency and deprecate ExtendedServer API

rbygrave:
- The ExtendedServer API is replaced by use of query usingTransaction()
- Add SqlQuery usingTransaction() for consistency with Query and DtoQuery

### DIFF
--- a/ebean-test/src/test/java/org/tests/query/sqlquery/SqlQueryTests.java
+++ b/ebean-test/src/test/java/org/tests/query/sqlquery/SqlQueryTests.java
@@ -196,6 +196,23 @@ public class SqlQueryTests extends BaseTestCase {
     assertThat(minCreated).isBefore(OffsetDateTime.now());
   }
 
+  @Test
+  public void typeQuery_usingTransaction() {
+    ResetBasicData.reset();
+
+    String sql = "select max(unit_price) from o_order_detail where order_qty > ?";
+    try (Transaction transaction = DB.createTransaction()) {
+
+      BigDecimal maxPrice = DB.sqlQuery(sql)
+        .setParameter(1, 2)
+        .mapToScalar(BigDecimal.class)
+        .usingTransaction(transaction)
+        .findOne();
+
+      assertThat(maxPrice).isNotNull();
+    }
+  }
+
   static class CustDto {
 
     long id;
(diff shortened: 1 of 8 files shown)


ebean-test/src/test/java/org/tests/query/sqlquery/SqlQueryTests.java
```diff
@@ -224,6 +241,44 @@ public CustDto map(ResultSet rset, int rowNum) throws SQLException {
...
 
+  @Test
+  void queryUsingTransaction() {
+    ResetBasicData.reset();
+
+    try (Transaction txn = DB.createTransaction()) {
+      String sql = "select id, name, status from o_customer where name is not null";
+      AtomicInteger counter = new AtomicInteger();
+      DB.sqlQuery(sql)
+        .usingTransaction(txn)
```
rPraml, line 252: The test would/should pass also without this line.
I'm wondering if the test should check if the passed transaction is REALLY used.
Maybe check, if `txn.logPrefix()` appears in logged sqls?

ebean-test/src/test/java/org/tests/query/sqlquery/SqlQueryTests.java
```diff
@@ -224,6 +241,44 @@ public CustDto map(ResultSet rset, int rowNum) throws SQLException {
...
+
+    try (Transaction txn = DB.createTransaction()) {
+      String sql = "select id, name, status from o_customer where name is not null";
+      AtomicInteger counter = new AtomicInteger();
+      DB.sqlQuery(sql)
+        .mapTo(CUST_MAPPER)
+        .usingTransaction(txn)
+        .findEach(custDto -> {
+          counter.incrementAndGet();
+          assertThat(custDto.name).isNotNull();
```
rPraml, line 275: I'm not sure, if a findEach will update Transaction.current(). If so, we can `assertThat(Transaction.current()).isEqualTo(txn)` here

ebean-test/src/test/java/org/tests/query/sqlquery/SqlQueryTests.java
rbygrave: I'll adjust this in master.  Having merged the jakarta.xml change it's now hard to work on this branch for some reason.

PR #2772, opened 2022-08-03
Add ebean-annotation to BOM

rPraml:
We should add ebean-annotation also to the maven-bom


rPraml:
@rbygrave could you take a look at the failing build.
I get
```
Missing POM for io.ebean:kotlin-querybean-generator:jar:13.6.6-SNAPSHOT
```
which I cannot explain. It complains about missing **13.6.6-SNAPSHOT** version, but I've greped all poms and there is no 13.6.6-SNAPSHOT (or I do not see it)

rbygrave:
> We should add ebean-annotation also to the maven-bom

Hmmm, not sure. The ebean-annotation version is always driven from ebean-api and it isn't optional etc.  Adding it to bom doesn't really add anything.

rPraml:
We use the ebean-bom in our root-pom as managed dependency. We have maven modules, that only rely on ebean-annotation, but we currently add ebean-api, that has correct dependency to ebean-annotation.

Of course, we can add ebean-annotation instead ebean-api in the particular maven module (maybe as provided scope) but this means, that we have to maintain an additional version.

So I thought, it would be a good idea, if the bom provides the correct annotation dependency.

rbygrave:
> We have maven modules, that only rely on ebean-annotation

Yeah ok, that makes sense - merging.

PR #2773, opened 2022-08-03
Bug: fetch and where with formula join will produce incorrect joins

AlexWagner:
When using a fetch and where on a formula joined field (maybe others are also affected) in a query,
the joins are build incorrectly.
The table alias is wrong, if you have a where clause with 3 path components and a fetch with 2.

See test case with childPerson -> parentPerson -> effectiveBean

The problem is somewhere located here: 
https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java#L680

"where" generates an extra join which is effectivly added to the root bean und not to the already existing child (from the select/fetch)

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
@@ -677,11 +677,6 @@ public final class SqlTreeBuilder {
         } else {
           // look in register ...
           String parentPropertyName = includeProp.substring(0, dotPos);
-          if (selectIncludes.contains(parentPropertyName)) {
-            // parent already handled by select
-            return childJoin;
-          }
-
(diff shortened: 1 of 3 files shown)


AlexWagner:
@rbygrave any ideas on this?

rbygrave:
Sorry, I've been focused on the enhancement stuff. Looking to get a release out and then I hope to get some time to look at this.

rbygrave:
I feel that we really need a logical version of `@Formula` based on paths that determines the joins using the normal orm mechanism.  That is, I see `@Formula` as really `@SqlFormula` and the fixed joins are problematic (once a formula is on a cardinality ToMany path it should be LEFT JOIN etc).

rbygrave:
This is solved by using the new `@Formula2` with logical paths and automatic join support rather than `@Formula` with the explicit sql joins.

rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3798 I'll close this.

Issue #2782, opened 2022-08-10
Illegal fetch size when using most recent MariaDB driver


Wooza:
Using [`setFetchSize(Integer.MIN_VALUE)`](https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/query/CQuery.java#L317) (introduced by #56) is not allowed according to the documentation of [`setFetchSize`](https://docs.oracle.com/en/java/javase/18/docs/api/java.sql/java/sql/Statement.html#setFetchSize(int)):

It seems that older MariaDB drivers (<3.x.x) accepted `Integer.MIN_VALUE` for compatibility reasons, however the new major driver version was apparently rewritten from scratch to closely follow the JDBC specs and thus raises an exception when called with `Integer.MIN_VALUE`.

According to the [documentation](https://mariadb.com/kb/en/about-mariadb-connector-j/#streaming-result-sets) the proper solution is to use `setFetchSize(1)`.
My current workaround is to stick with the older driver but I would really like to update.

rbygrave:
Do you want to submit a PR that bumps the jdbc driver the tests are using for MariaDB ?  (that will then reproduce the problem)

PR #2783, opened 2022-08-10
Bump mariadb-java-client to 3.0.7

Wooza:
Bumped mariadb-java-client version to 3.0.7 to see which tests may break (see #2782)

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/CQuery.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/CQuery.java
@@ -314,7 +314,7 @@ public final class CQuery<T> implements DbReadContext, CancelableQuery, SpiProfi
       if (forwardOnlyHint) {
         // Use forward only hints for large resultSet processing (Issue 56, MySql specific)
         pstmt = conn.prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
-        pstmt.setFetchSize(Integer.MIN_VALUE);
+        pstmt.setFetchSize(1);
       } else {
         pstmt = conn.prepareStatement(sql);
       }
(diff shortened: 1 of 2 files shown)


Wooza:
Given that all tests passed I would assume that the relevant code path was not taken by any of the tests? I'm not sure about your CI setup but there seems to be no coverage data to verify this?

rbygrave:
> relevant code path was not taken by any of the tests?

It's more that we have different github actions for the different databases and we don't run those database specific actions on each PR.  So the PR's are run against H2 database.

Normally I can choose the MariaDB action and run that against a branch ... but it isn't seeing this PR branch. Hmmm.

rbygrave:
Yeah, I can't run the action against the forked branch.  I'll look to merge this in and then run using the MariaDB action and that should fail.

Background:
From memory ebean is using that weird setting to get mysql jdbc driver to "stream" results rather than "pull all results into local buffer" and ebean does this for findEach, findStream, findIterate queries. There is also another "bit of a hack/limitation" we hit with mysql around firing secondary queries while streaming results so perhaps the mariadb driver also fixed that issue.

ebean-core/src/main/java/io/ebeaninternal/server/query/CQuery.java
```diff
@@ -314,7 +314,7 @@ ResultSet prepareResultSet(boolean forwardOnlyHint) throws SQLException {
       if (forwardOnlyHint) {
         // Use forward only hints for large resultSet processing (Issue 56, MySql specific)
         pstmt = conn.prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
-        pstmt.setFetchSize(Integer.MIN_VALUE);
+        pstmt.setFetchSize(1);
```
rbygrave, line 317: We can't do this part of course because that breaks MySql.

PR #2819, opened 2022-09-01
ENH: Add Transaction.rollbackAndContinue()

rbygrave:
Typically useful for handling DuplicateKeyException where we expect
DuplicateKeyException to be thrown and catch it with the intention of
continuing processing using the same transaction.

Note that some databases like Oracle do not require this explicit
rollback() and would work without the rollbackAndContinue(). Postgres
in particular requires the rollback() call on the underlying connection
such that we can continue using that transaction/java.sql.Connection.

Note that in the existing test we can see that rollbackAndContinue()
is pretty close to being syntactic sugar. I think adding rollbackAndContinue()
is justified and complements the existing commitAndContinue().

### DIFF
--- a/ebean-api/src/main/java/io/ebean/DuplicateKeyException.java
+++ b/ebean-api/src/main/java/io/ebean/DuplicateKeyException.java
@@ -2,6 +2,28 @@ package io.ebean;
 
 /**
  * Thrown when a duplicate is attempted on a unique constraint.
+ * <p>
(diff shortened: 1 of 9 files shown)


vilmosnagy:
@rbygrave the `isActive` method was made final by this PR, and broke the spring-jdbc transactions.See: https://github.com/ebean-orm/ebean-spring-txn/issues/28

Do you have any better idea than to make the `isActive` method non-final again?

PR #2837, opened 2022-09-19
Migrate to Jakarta JAXB (jakarta.xml.bind <- javax.xml.bind)

rbygrave:
Reverts ebean-orm/ebean#2787

This is actually a revert of a revert of ... Migrate to Jakarta JAXB

That is, we migrated to Jakarta JAXB, found that apps had issues as this conflicted with 3rd party libraries that had a dependency on Javax JAXB and had to revert back to Javax JAXB.

When we merge this in we could face the same issue that apps with 3rd party dependency on javax jaxb have issues resolving that. At this stage we are waiting a little bit to see when people demand this change (to say run on the very latest version of Spring). The alternative is to migrate away from jaxb.

### DIFF
--- a/ebean-autotune/pom.xml
+++ b/ebean-autotune/pom.xml
@@ -32,10 +32,15 @@
 
     <!-- needed for java 11+ -->
     <dependency>
-      <groupId>org.glassfish.jaxb</groupId>
-      <artifactId>jaxb-runtime</artifactId>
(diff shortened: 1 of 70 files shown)


rbygrave:
This PR is now not required as we are using scripts to maintain a Jakarta fork of Ebean.

Issue #2854, opened 2022-10-12
Possible memory leak with BackgroundExecutor using BackgroundExecutorWrapper ThreadLocals (for scheduled tasks)


rPraml:
Hello Rob, we introduced with https://github.com/ebean-orm/ebean/pull/2590 a concept, that each operation that is executed by the BackgroundExecutor may inherit ThreadLocals by the originating thread

## Expected behavior

when a runnable/callable will be submitted to the executor, ThreadLocals are available in the async thread.
After execution, the memory should be free-ed.

## Actual behavior

The ThreadLocals are also copied to periodic threads and will stay there forever.

### Investigation

I discovered this issue, because I was wondering why a certain user session was not garbage collected.

What happened:

We store the user session in a ThreadLocal, so that it is always available in the current thread.
This ThreadLocal acts also as CurrentTenant/CurrentUserProver. We also have a BackgroundExecutorWrapper that copies this information.

When a User now accesses a cache the first time, a ServerCache is created with computeIfAbsent [here](https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/cache/DefaultCacheHolder.java#L77)

The cache registers a periodicTrim-task in the BackgroundExecutor and the BackgroundExecutorWrapper will copy the current ThreadLocal (where it stays forever)

long story short: @rbygrave What would you suggest, how to fix it
- only wrap "one-time-tasks" that are executed or submitted (IMHO it is not needed - in our use cases - to copy ThreadLocals to periodic tasks)
- add a boolean flag to BackgroundExecutorWrapper.wrap that indicates, if this is a periodic task. The code in the wrapper can then decide, if it is necessary to copy threadLocals
- add a BackgroundExecutorWrapper.wrapPeriodically

I can create a PR when I get feedback

rbygrave:
> only wrap "one-time-tasks"

I agree. Yes, I think this is the change / fix we should go with.

> not needed - in our use cases - to copy ThreadLocals to periodic tasks

Yes and I believe that is **_generally_** true.  Scheduled tasks by their nature should not "inherit" ThreadLocals - I think that makes good sense to me. Lets make it so.

> I can create a PR when I get feedback

Awesome, thanks!!

PR #2859, opened 2022-10-17
Possible memory leak with BackgroundExecutor using BackgroundExecutorWrapper ThreadLocals (for scheduled tasks)

nPraml:
Hello @rbygrave ,
hello @rPraml ,

I created a PR with the fix of https://github.com/ebean-orm/ebean/issues/2854. 
Could you please check it?

@rbygrave the maven build does not work local on my machine, `CsvReaderTest` throws a `java.time.format.DateTimeParseException: Text '01-Jan-2009' could not be parsed at index 3`.
The test works only if I use `-Duser.language="en"`.

Cheers
Noemi

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
@@ -106,23 +106,23 @@ public final class DefaultBackgroundExecutor implements SpiBackgroundExecutor {
 
   @Override
   public ScheduledFuture<?> scheduleWithFixedDelay(Runnable task, long initialDelay, long delay, TimeUnit unit) {
-    return schedulePool.scheduleWithFixedDelay(wrap(logExceptions(task)), initialDelay, delay, unit);
+    return schedulePool.scheduleWithFixedDelay(logExceptions(task), initialDelay, delay, unit);
   }
 
   @Override
   public ScheduledFuture<?> scheduleAtFixedRate(Runnable task, long initialDelay, long delay, TimeUnit unit) {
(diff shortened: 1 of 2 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/executor/DefaultBackgroundExecutor.java
```diff
@@ -106,23 +106,23 @@ public void execute(Runnable task) {
...
 
   @Override
   public ScheduledFuture<?> scheduleAtFixedRate(Runnable task, long initialDelay, long delay, TimeUnit unit) {
-    return schedulePool.scheduleAtFixedRate(wrap(logExceptions(task)), initialDelay, delay, unit);
+    return schedulePool.scheduleAtFixedRate(logExceptions(task), initialDelay, delay, unit);
   }
 
   @Override
   public ScheduledFuture<?> schedule(Runnable task, long delay, TimeUnit unit) {
-    return schedulePool.schedule(wrap(logExceptions(task)), delay, unit);
```
rPraml, line 119: I'm not sure, if we should keep it here. The task will be scheduled one time. But this may keep objects also long in memory.
Maybe the caller is responsible to pass a proper runnable/callable

rbygrave: I'm not sure either - it almost looks like a 50/50 call to me. 

If we do wrap() here it would be a problem if the delay is _"a long time"_, and if that is a problem for a particular use case then what is the workaround? I don't see a nice workaround so that is suggesting to me that we don't wrap() here.

rbygrave (approved):
Approving, we just need agreement on whether to wrap() on the 2 `schedule()` methods.

nPraml:
Hello @rbygrave ,
we adapted the javadoc in `BackgroundExecutorWrapper`.
We would not use wrap in the two schedule methods.

PR #2863, opened 2022-10-25
FIX: LogFormat for SystemLogger

rPraml:
Found two wrong formatted log messages (Search regexp `\.log\(.*\{\}`)


rbygrave:
Nice :)

PR #2864, opened 2022-10-25
NEW: Override service objects with DatabaseConfig.putServiceObject

rPraml:
I've extended the 'putServiceObject' mechanism in the serverconfig.

So you can override (some) objects that are normally injected via SPI in DatabaseConfig for a certain server instance.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
+++ b/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
@@ -615,12 +615,34 @@ public class DatabaseConfig {
   }
 
   /**
-   * Put a service object into configuration such that it can be passed to a plugin.
+   * Put a service object into configuration such that it can be used by ebean or a plugin.
    * <p>
    * For example, put IgniteConfiguration in to be passed to the Ignite plugin.
    */
   public void putServiceObject(String key, Object configObject) {
     serviceObject.put(key, configObject);
+  }
+
+  /**
+   * Put a service object into configuration such that it can be used by ebean or a plugin.
+   * <p>
+   * For example, put IgniteConfiguration in to be passed to the Ignite plugin.
+   * You can also override some SPI objects that should be used for that Database. Currently, the following
+   * objects are possible.
+   * <ul>
+   *   <li>DataSourceAlertFactory (e.g. add different alert factories for different ebean instances)</li>
+   *   <li>DocStoreFactory</li>
+   *   <li>XmapService</li>
+   *   <li>SpiLoggerFactory (e.g. add custom logger for a certain ebean instance)</li>
+   *   <li>AutoTuneServiceProvider</li>
+   *   <li>SpiProfileHandler</li>
+   *   <li>SlowQueryListener (e.g. add custom query listener for a certain ebean instance)</li>
+   *   <li>ServerCacheNotifyPlugin</li>
+   *   <li>SpiDdlGenneratorProvider</li>
+   * </ul>
+   */
+  public <T> void putServiceObject(Class<T> iface, T configObject) {
+    serviceObject.put(serviceObjectKey(iface), configObject);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


rPraml:
what do you think about this change?

Background:
When our application runs in debug mode, we add a `SpiLoggerFactory` that intercepts the SQLs and effectively sends the executed SQLs with some javascript-magic to the browser console,
This allows us to debug executed SQLs directly in the browser window.

It is not so easy to use the service-loader pattern, as the factory is loaded in all cases, but we want to control this with a "debug" parameter (and only for one ebean server)

rbygrave:
Conceptually fine but using interfaces in  `io.ebeaninternal.api` isn't going to work in module-path (as that package is not open/exported).

Any custom services provided in this manor really should be either moved to the `ebean-api` module into the `io.ebean.service` package or we should adjust `ebean-core` to export a suitable package (for public use) like say `io.ebean.core.api`.

So yes we could merge this change but pretty quickly we'd want to move the SpiLoggerFactory and SpiLogger I think.

Or perhaps what people really desire is to capture the SQL (they don't care about io.ebean.SUM or io.ebean.TXN ?) and we haven't made it particularly easy for folks to do that (but then folks that do this might like all the control we give here).

jnehlmeier:
Maybe a different approach that would also work for you:

I use a set of `Delegating[DataSource|Connection|Statement|PreparedStatement]` classes that do nothing except delegating every method to the wrapped Datasource/Connection/etc. Basically a small dedicated library.

To capture SQL I would then implement a `SqlCapturingDataSource extends DelegatingDataSource`, override `getConnection` to return a `SqlCapturingConnection` and implement capturing code in that class.

With that pattern you can do many extension on datasource level. I use it to add metrics (e.g. how many select/insert/update/deletes), to track connection closing, to implement a datasource that can reconfigure itself while being in active use, ..., ...

rPraml:
Our main goal is really to intercept the SQL/SUM logs in a special debug mode.

> I use a set of Delegating[DataSource|Connection|Statement|PreparedStatement] classes that do nothing except delegating every method to the wrapped Datasource/Connection/etc. Basically a small dedicated library.

I know this concept (We use JavaMelody, which does exactly that, builds also statistics and so on), but I like the extra bits in `io.ebean.SQL` that adds the bind log and batch binds in a compressed form. I'm sure the same functionality can be implemented in such a library, but why implement it twice

> Or perhaps what people really desire is to capture the SQL

We use LoggedSql in tests.

> (they don't care about io.ebean.SUM or io.ebean.TXN ?) 

unfortunately, we do (not in production, but in tests). We check for `+lazy` in SUM and throw an exception, if a certain code block triggers a lazy-load. If someone changes code and forget to select the new property, the test will fail with a stack trace causing the lazy load

This is active in most of our unit tests and helps us to keep the application fast.
Of course, you cannot avoid all lazy loads, so it is possible to define exclusions (e.g `@Cache` entities / lob fields) or disable the check temporary for a certain code block.

> So yes we could merge this change but pretty quickly we'd want to move the SpiLoggerFactory and SpiLogger I think.

Maybe we should sleep a night, if there are better solutions fore these use cases

1. We need a mechanism to start logging at the begin of a web request (Similar to LoggedSql, but with a ThreadLocal) and when the web request ends, we add some javascript (`<script>console.log(...)</script>`) that prints the SQL/SUM/TXN in the browser console. (Note: This is only for development!)
2. We need a mechanism to detect/avoid lazy loads, as they may slow down the application. (Grepping the log was not a good option, because we did not have the stack, where it comes from) Ebean does a good job in handling lazy loads, but in most cases it is better to avoid them

rbygrave:
I am happy with this PR except for the javadoc comment - I don't think we should advertise any service in that list that isn't public in module-path (for example, anything in `io.ebeaninternal` like `SpiLoggerFactory` isn't public in module-path).

I am happy enough for folks like yourself to use classpath and know that you can config SpiLoggerFactory this way but we should not advertise it until we make it public in module-path.

At some point not too far away we need to move anything that we want to configure like `SpiLoggerFactory` into a package that is public in module-path.  (Perhaps a `io.ebean.core` or `io.ebeanservice.api` or `io.ebean.extension`).  In this sense the `io.ebeaninternal.server.cache` package should probably  be renamed to something more public friendly that does not start with `io.ebeaninternal` like `io.ebean.core.cache`assuming we want it public for advanced users.

PR #2866, opened 2022-10-25
FIX: @Formula / @Where Annotations should also match on base platform

rPraml:
A property annotated with
```java
@Formula(select = "...", platforms = SQLSERVER)
```
should match for SQLSERVER17 and SQLSERVER19


rPraml:
... I should add a test ;)

PR #2870, opened 2022-10-25
FIX: Run CsvReaderTest also in german locale

rPraml:


xabolcs:
Hi @rPraml ,
Would you mind include a GitHub Action to uncover these language related issues?

Btw this was reported by Noemi in PR #2859:
> **rbygrave** the maven build does not work local on my machine, `CsvReaderTest` throws a `java.time.format.DateTimeParseException: Text '01-Jan-2009' could not be parsed at index 3`.
> The test works only if I use `-Duser.language="en"`.

rPraml:
@xabolcs Yes, it'll make sense, but there are so many coordinates in the environment, that have influence:

- Locale (English, german, [Turkey test](http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html))
- Timezone (GMT, positive / negative)
- DB-Platform (H2)
- OS (Filepaths in Windows/Linux)
- Java (e.g. different Instant precision between Java8/11)
-... not yet talking from the Locale/TZ of the DB-server that may be different to the JVM Locale/TZ

In theory, you should test that with all combinations, which will lead in 100s of CI plans

@rbygrave what do you think, Should we add one or two GH-actions with some "strange" settings and run them against H2 on java 17? e.g.
- Locale=tr_TR, TZ=UTC+14 (Line Islands)
- Locale=de_DE, TZ=UTC-8 (negative)
Or do you have other ideas?

xabolcs:
> In theory, you should test that with all combinations, which will lead in 100s of CI plans

Yeah those combinations explode in an instant, but it would be nice to include a few, as you propose it already.

rbygrave:
Well, every CI build consumes power in some data center and I think we need to be conscience of this fact. Also, to some extent we run the tests locally with the JVM in (NZ and DE) timezones & locales frequently.

Right now I'd suggest Ebean doesn't use Locale much at all so I don't see it as a priority and personally there are too many other more important things to do.

Issue #2774, opened 2022-08-05
Ebean should support also IN for local encrypted properties.


rPraml:
While the current [documentation ](https://ebean.io/docs/features/encryption) is very clear, that **EQ** (and only EQ) is supported for encrypted properties, I noticed, that **IN** is currently not supported (which is semantically only a concatenation of multiple EQ queries)
So the documentation is right here ;)

I ponder if all exact match queries ("eq","ne","in","notIn") should work for encrypted properties? I see no technical reason, why they should not.

What do you think?

## Expected behavior
```
found = Ebean.find(EBasicEncryptClient.class)
      .where()
      .in("description", "goodbye")
      .in("status", EBasicEncryptClient.Status.TWO)
      .in("dob", today) // <-- throws a Data Conversion error
      .findOne();
```
should work

rbygrave:
The implementation detail for this to do database side (using database functions) probably isn't conducive to _IN_. 

Have you looked at how this could be done with _IN_ (especially on non-Postgres dbs that don't have array binding)?  Everything is doable but my impression is that this would probably be messy implementation wise.
 
Do you need this functionality?

rbygrave:
Oh, are you saying ONLY for client side encryption support this case?

rPraml:
Yes, I meant client side encrypted properties.

I can check, if it will also work with server side encryption.

rPraml:
closed. See #2872

PR #2872, opened 2022-10-27
#2774 Locally encrypted properties can be used in "in" queries

rPraml:
Ebean already supports encrypted properties in "eq" queries. (See #2774)

Now we can also use them in "in" queries.

Note: locally encrypted properties can be used only in exact match queries https://ebean.io/docs/features/encryption#encrypted


rPraml:
This PR mainly adds the missing local encryption feature for "in" query.

For the sake of completenes, I reviewed the existing expressions and made this overview:
```
Expression           | dbEnc | local |
---------------------+-------+-------+
InQuery              | No¹   |   -   |
Like                 | Yes   |   -   |
Simple               | Yes   | Yes   |
CaseInsensitiveEqual | Yes   |   -   |
Null                 | Yes²  | Yes²  |
ArrayContains        | No¹   | No¹   |
NativeILike          | Yes   |   -   |
InRange              | No¹   |   -   |
In                   | Yes   | Yes   |
JsonPath             |   -   |   -   |
InPairs              | No¹   | No¹   |
Bitwise              | No¹   |   -   |
AbstractText         |   -   |   -   |
Between              | No¹   |   -   |
ArrayIsEmpty         |   -²  |   -   |
IsEmpty              |   -²  |   -   |

-) Technically not possible/makes no sense
¹) Could be implemented
²) Null/Empty values are not encrypted
```

@rbygrave There are some expressions left, that do not yet support dbEncryption (although ebean claims it's fully transparent)
It should be easy to add "dbEncryption" support for `InRange` /  `Bitwise` / `Between` (it is straightforward to `Like` and if you want, I can do that in a separate PR)

`InPairs` woud be more difficult, as it should be possible also to add local encryption support and I'm unsure if it will make sense
I'm unsure, what to do with `ArrayContains` - It's technically possible, but as far as I see, this is very postgres specific.
I would not add dbEncryption support for `InQuery`- if you use a subquery on an encrypted property, it is very likely, that the result of the subquery is also encrypted (at least it should be)

Cheers
Roland

PR #2881, opened 2022-11-02
Add StdFunctions, StdExpressions, TQColumn marker interface

rbygrave:


ebean-querybean/src/main/java/io/ebean/typequery/PBaseComparable.java
```diff
@@ -197,8 +198,8 @@ public final R inRange(T lower, T upper) {
    * The most common use of this could be called "effective dating" where 2 date or
    * timestamp columns represent the date range in which
    */
-  public final R inRangeWith(TQProperty<R> highProperty, T value) {
```
rPraml, line 200: I wonder, if this is correct.
Before, you could pass only properties from the model R (but with any type)
Now you can pass any property, as long as the type matches

ebean-querybean/src/main/java/io/ebean/typequery/PBaseString.java
```diff
@@ -5,7 +5,7 @@
  *
  * @param <R> the root query bean type
  */
-public abstract class PBaseString<R,T> extends PBaseCompareable<R, String> {
+public abstract class PBaseString<R, T> extends PBaseComparable<R, T> {
```
rPraml, line 8: What was the intention of this change?
I see, that you compare Cidr and other with PBaseString.

ebean-api/src/main/java/io/ebean/StdOperators.java
```diff
@@ -0,0 +1,198 @@
...
+    StringBuilder expression = new StringBuilder(50);
+    expression.append("concat(").append(property.toString());
+    for (Object value : values) {
+      expression.append(",").append(sqlConcatString(value));
+    }
+    expression.append(")");
+    return Property.of(expression.toString());
+  }
+
+  private static String sqlConcatString(Object value) {
```
rPraml, line 54: Use sqlValue instead?

ebean-api/src/test/java/io/ebean/StdFunctionsTest.java
```diff
@@ -0,0 +1,69 @@
...
+    assertThat(upper(foo).toString()).isEqualTo("upper(foo)");
+  }
+
+  @Test
+  void testConcat() {
+    assertThat(concat(foo, "+").toString()).isEqualTo("concat(foo,'+')");
+    assertThat(concat(foo, 42).toString()).isEqualTo("concat(foo,'42')");
+    assertThat(concat(foo, LocalDate.of(2022, 9, 1)).toString()).isEqualTo("concat(foo,'2022-09-01')");
+    assertThat(concat(foo, bar).toString()).isEqualTo("concat(foo,bar)");
+    assertThat(concat(foo, ":", 42, bar).toString()).isEqualTo("concat(foo,':','42',bar)");
```
rPraml, line 57: Please ensure that no SQL can be injected:
`assertThat(concat(foo, ":
it's my name").toString()).isEqualTo("concat(foo,'it''s my name')");`

ebean-querybean/src/main/java/io/ebean/typequery/PBaseComparable.java
rbygrave: I believe this is better here to allow any property as long as the type matches. This specifically allows that case that Jens had where the range check was across 2 different types via a ManyToOne path (but any path should be fine).

ebean-querybean/src/main/java/io/ebean/typequery/PBaseString.java
rbygrave: The intention is for types like Cdir to nicely support operators (eq, ne, like etc) with both their natural T type (Cdir) and String.

Given a cdir type we want to:

```java
.cdir.eq(myCdirValue)
```
and also just use strings (especially for like, startsWith, contains etc)  ...
```java
.cdir.eq("some-string-value")
```

... but let me check this.

ebean-api/src/main/java/io/ebean/StdOperators.java
rbygrave: I don't think so no unless we leave it up to the database to cast numbers to strings? I'll adjust this for sql injection.

ebean-api/src/test/java/io/ebean/StdFunctionsTest.java
rbygrave: This should be `isEqualTo("concat(foo,': it''s my name')");`  right?

ebean-querybean/src/main/java/io/ebean/typequery/PBaseString.java
rbygrave: Yes, I'm happy with this.  Effectively we overload all the types that extend PBaseString so that they can take as parameters String or their type.

For example with Inet we have both `eq(Inet value)`  and `eq(String value)`

This deprecates the odd existing methods `equalToType(T value)` and `notEqualToType(T value)`

PR #2902, opened 2022-11-27
Fixed issue with using offset without limit

serg1236:
This PR contains the fix for the issue https://github.com/ebean-orm/ebean/issues/2903: `LimitOffsetSqlLimiter` that applies `LIMIT 0` to the query even if `firstRow` is set only and `maxRows` is untouched.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/config/dbplatform/LimitOffsetSqlLimiter.java
+++ b/ebean-api/src/main/java/io/ebean/config/dbplatform/LimitOffsetSqlLimiter.java
@@ -31,12 +31,12 @@ public class LimitOffsetSqlLimiter implements SqlLimiter {
     int firstRow = request.getFirstRow();
     int maxRows = request.getMaxRows();
 
-    if (maxRows > 0 || firstRow > 0) {
+    if (maxRows >= 0) {
       sb.append(" ").append(LIMIT).append(" ").append(maxRows);
-      if (firstRow > 0) {
-        sb.append(" ").append(OFFSET).append(" ");
-        sb.append(firstRow);
-      }
+    }
+    if (firstRow > 0) {
+      sb.append(" ").append(OFFSET).append(" ");
+      sb.append(firstRow);
     }
 
     String sql = request.getDbPlatform().completeSql(sb.toString(), request.getOrmQuery());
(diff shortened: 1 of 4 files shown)


ebean-api/src/main/java/io/ebean/config/dbplatform/LimitOffsetSqlLimiter.java
```diff
@@ -31,12 +31,12 @@ public SqlLimitResponse limit(SqlLimitRequest request) {
     int firstRow = request.getFirstRow();
     int maxRows = request.getMaxRows();
 
-    if (maxRows > 0 || firstRow > 0) {
+    if (maxRows >= 0) {
```
rbygrave, line 34: Why do you think that maxRows 0 is valid?  Why is this not `maxRows > 0` ?

serg1236: LIMIT 0 is a proper SQL expression. Not sure about the use cases but still valid.

rPraml: Hmm. What would we expect when maxRows = 0?
- empty result (so query needs not to be executed) 
- the complete result (without limit)
Both cases are valid for me.

rbygrave: LIMIT 0 means it will **_always_** return an empty set. Ebean has never supported that and I'd argue it has no utility value for ORM per say.

Changing this allowing LIMIT 0 in this PR means that this PR introduces 2 changes rather than just the one that is stated - allowing LIMIT 0 is a behavior change, I'm not keen to see that included in this change.

serg1236: After a short web research I found out LIMIT 0 is sometimes used to check the query validity against the DB server. 
It seems pretty confusing that explicit setMaxRows(0) doesn't produce LIMIT 0. The other way is to restrict the argument to [1;MAX_INTEGER]

rbygrave: As I see it this PR should only have the 1 change (allowing OFFSET without LIMIT).

The change to support LIMIT 0 needs to be in a separate PR. I see support for LIMIT 0 as a debatable behavior change as opposed to a bug fix so it would need to go into a major | minor version bump release and not a bug fix release etc.

serg1236: Ok. I'll split the PRs.

serg1236: @rbygrave Please check https://github.com/ebean-orm/ebean/pull/2904 and https://github.com/ebean-orm/ebean/pull/2905 PRs

PR #2912, opened 2022-12-01
Refactoring SpiLogger to better support conditional logging

rbygrave:
Initial step, remove trace() logging from SpiLogger
- Not bothered to log when changing isolation level
- Not bothering to log at end of query only transaction

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
@@ -932,16 +932,16 @@ public final class PersistRequestBean<T> extends PersistRequest implements BeanP
     String name = beanDescriptor.name();
     switch (type) {
       case INSERT:
-        transaction.logSummary("Inserted [" + name + "] [" + (idValue == null ? "" : idValue) + draft);
+        transaction.logSummary("Inserted [" , name , "] [" , (idValue == null ? "" : idValue.toString()) , draft);
         break;
       case UPDATE:
-        transaction.logSummary("Updated [" + name + "] [" + idValue + draft);
+        transaction.logSummary("Updated [" , name , "] [" , idValue.toString() , draft);
         break;
       case DELETE:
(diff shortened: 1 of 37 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
```diff
@@ -932,16 +932,16 @@ private void logSummaryMessage() {
     String name = beanDescriptor.name();
     switch (type) {
       case INSERT:
-        transaction.logSummary("Inserted [" + name + "] [" + (idValue == null ? "" : idValue) + draft);
+        transaction.logSummary("Inserted [" , name , "] [" , (idValue == null ? "" : idValue.toString()) , draft);
```
rPraml, line 935: Maybe change logger signature to `logSummary(Object... msg)`

ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestCallableSql.java
```diff
@@ -86,7 +82,7 @@ public void postExecute() {
       persistExecute.collectSqlCall(label, startNanos);
     }
     if (transaction.isLogSummary()) {
-      transaction.logSummary("CallableSql label[" + callableSql.getLabel() + "]" + " rows[" + rowCount + "]" + " bind[" + bindLog + "]");
+      transaction.logSummary("CallableSql label[", callableSql.getLabel(), "] rows[", String.valueOf(rowCount), "] bind[", bindLog, "]");
```
rPraml, line 85: +1 for `logSummary(Object... msg)`

rPraml (commented):
looks good so far, looking forward to trying it out

ebean-core/src/main/java/io/ebeaninternal/server/core/PersistRequestBean.java
rbygrave: Yes agreed. It's starting to tell me it _could even_ be `logSummary(String msgFormat, Object... args)` ...

Issue #2918, opened 2022-12-15
findStream() / findEach() memory leak


serg1236:
## Expected behavior
Query::findStream() and Query::findEach() are supposed to not cache any items in the persistence context.

## Actual behavior

Items are being cached in the persistence context causing OutOfMemoryError after some time. 
I've done simple tests:
```
//...
myQuery.findStream().forEach(System.out::println);
```

```
//...
myQuery.findEach(System.out::println);
```

Here is a memory snapshot from the profiler after about 20 sec of execution:
<img width="1255" alt="image" src="https://user-images.githubusercontent.com/10487569/207874853-c1920dd5-d205-40b3-939f-e16e3e5577f2.png">

P.S. I use Postgresql DB.

rPraml:
Hello seg1236
which ebean version are you using? Can yo check, if the issue #2411 (and commit #2413) are in your version. I would expect, that a findeach should use weak references, which should be cleaned up, if a bean is no longer referenced.
Do you store the beans in a list or something similar? 

Roland

serg1236:
@rPraml I use 13.11.0. I don't collect any data. I also tried to run GC explicitly and it didn't help: 
```
try (var iterator = vectorsQuery.findIterate()) {
            while (iterator.hasNext()) {
                if (counter.incrementAndGet() % 2000 == 0) {
                    System.out.println("COLLECTOR");
                    System.gc();
                }
                System.out.println(counter.get() + " " + iterator.next());
            }
        }
```

rPraml:
@serg1236 I tried to reproduce that, so I write a simple test (See below)

The test uses a simple entity and creates 10M of them i a DB file. (H2 in memory will not work)
It also uses its own serverConfig. Otherwise, the classpathScan will find the [TDChangelogListener](https://github.com/ebean-orm/ebean/blob/master/ebean-test/src/test/java/org/tests/changelog/TestChangeLog.java#L201) and this ChangelogListener will put all models in an internal list. (Which makes it difficult/impossible to create 10M entries)
Note to @rbygrave: this should be fixed in the test framework. The default server should not find classes like TDChangelogListener and use them.

But back th the test.
After the "test-set" is created (resulting DB file is ~2,7GB) I started the second test, which does a findEach and then a findList.
The output was:
```
Doing findEach
Read 10000000 entries
Doing findStream
Read 10000000 entries
Doing findIterate
Read 10000000 entries
Doing FindList

[...]

PR #2922, opened 2023-01-04
FIX memory leak in streaming queries when embedded-ids are used #2918

rPraml:


rPraml:
This shoud fix issue #2918

ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
```diff
@@ -279,6 +280,10 @@ private void put(Object id, Object bean) {
       Object existing;
       if (useReferences) {
         weakCount++;
+        if (id instanceof EntityBean) {
+          // we are an embedded ID
+          id = cloneEmbeddedId((EntityBean) id);
+        }
         existing = map.put(id, new BeanRef(this, id, bean, queue));
```
serg1236, line 287: It still can run out of memory if ids are quite large

rPraml: Unfortunately I can't quite follow you here. Do you mean "one ID is big" or do you mean "many IDs are in the cache"?

- "one ID is big": An ID should not be big in DB terms. You should keep your primary keys small. There are often limitations: MariaDB can only use 16 columns and has a max key length of 767 bytes. So I think an ID itself should always fit in < 1kb of ram. For me it makes also no sense to have large text fields as primary key.
- So I'm sure you ment "many IDs": There should not be too much entries in the cache. If the memory goes low, the gc will put the `BeanRefs` in the refQueue (if there are no strong-references of course) and `DefaultPersistenceContext.expungeStaleEntries` will remove the whole map entry.

If I've missed something here, please report back :)

Nevertheless, this PR is not the best solution, because there are many caches in ebean (and in our app) that have the same problem if embedded IDs were used.

cheers
Roland

serg1236: @rPraml I mean, `map` is a regular HashMap. Id is put there as a "strong" reference. As I understand context is not cleared until the streaming is over. So even if we have "weak" values in map, the key set itself may cause OutOfMemory. It will just take longer.

rbygrave: > the key set itself may cause OutOfMemory

No. As Jonas points out the key is removed.

rbygrave:
This does not look like it will be the preferred approach. Instead, I'm looking at just NOT setting the embeddedOwner for the **_Id_** embedded beans.  

That is, if it is an Id embedded bean we don't need the dirty state to propagate back to the embeddedOwner (so we can not set it and left it as null for those Id embedded beans).

ebean-core/src/main/java/io/ebeaninternal/server/transaction/DefaultPersistenceContext.java
serg1236: @rbygrave Why do you need to cache entities while streaming in the first place?

rPraml: You need it, to guarantee that you get the same object.

Assume you iterate over invoices and they reference a customer.
So you can have two invoices pointing to the same customer.
What we want is, that we don't have multiple instances of the same customer in memory. They should all point to the same object.

You are partially right, that we do not need this in findeach... as long as you do not hold references to previous objects.
Assume that you try to compare previous object with the current one, while iterating.

serg1236: @rPraml Interesting. I never expect it to be the same object, to be honest. Each entity should have proper `equals()` and `hashCode()` implementations. That's it.

rbygrave: If the projection includes a ManyToOne, OneToOne, OneToMany or ManyToMany we need the persistence context to de-duplicate the result in order to produce consistent object graphs.

This is the fundamental difference between an ORM query and a 'row mapping' query like DtoQuery.

If your query does not include any relationships then yes the persistence context isn't doing any de-duplication.

rbygrave:

PR #2923, opened 2023-01-04
Other approach: Change owner to weak reference

rPraml:


rPraml:
Other approach for #2918

rbygrave:

PR #2924, opened 2023-01-04
Other approach: Use ReadOnlyIntercept for embedded ids when read from DB

rPraml:
Other approach for #2918 - can we change the creation of EmbeddedIds and use the read-only intercept?

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/AssocOneHelpEmbedded.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/AssocOneHelpEmbedded.java
@@ -32,7 +32,7 @@ final class AssocOneHelpEmbedded extends AssocOneHelp {
 
   @Override
   Object read(DataReader reader) throws SQLException {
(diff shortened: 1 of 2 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/deploy/AssocOneHelpEmbedded.java
```diff
@@ -32,7 +32,7 @@ Object readSet(DataReader reader, EntityBean bean) throws SQLException {
 
   @Override
   Object read(DataReader reader) throws SQLException {
-    EntityBean embeddedBean = property.targetDescriptor.createEntityBean();
+    EntityBean embeddedBean = property.targetDescriptor.createEntityBean2(true);
```
rbygrave, line 35: This look like it includes non-id embedded beans

rPraml: Hmm, yes. This is should be done only for ID beans

rbygrave:

PR #2927, opened 2023-01-04
NEW: BeanCache for findSet

rPraml:
This adds beancache support to "findSet" queries.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
@@ -515,12 +515,12 @@ public final class OrmQueryRequest<T> extends BeanRequest implements SpiOrmQuery
       if (query.getType() == Type.MAP) {
         mergeCacheHitsToMap(result);
       } else {
-        mergeCacheHitsToList(result);
+        mergeCacheHitsToCollection(result);
       }
     }
   }
 
-  private void mergeCacheHitsToList(BeanCollection<T> result) {
+  private void mergeCacheHitsToCollection(BeanCollection<T> result) {
     for (T hit : cacheBeans) {
       result.internalAdd(hit);
     }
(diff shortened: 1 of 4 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
```diff
@@ -579,6 +579,15 @@ private ElPropertyValue mapProperty() {
     return property;
   }
 
+  @Override
+  public Set<T> beanCacheHitsAsSet() {
+    OrderBy<T> orderBy = query.getOrderBy();
+    if (orderBy != null && !orderBy.isEmpty()) {
+      beanDescriptor.sort(cacheBeans, orderBy.toStringFormat());
```
rPraml, line 586: should we support ordering on sets? Maybe return a normal hashSet, if no orderBy is set

ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
```diff
@@ -515,12 +515,12 @@ public void mergeCacheHits(BeanCollection<T> result) {
       if (query.getType() == Type.MAP) {
         mergeCacheHitsToMap(result);
       } else {
-        mergeCacheHitsToList(result);
+        mergeCacheHitsToCollection(result);
```
rPraml, line 518: better name

ebean-test/src/test/java/org/tests/cache/TestBeanCache.java
```diff
@@ -157,6 +158,80 @@ public void idsInExpression() {
     }
   }
 
+  @Test
+  public void testFindSet() {
```
rPraml, line 162: This is the same test as 'idsInExpression' maybe refactor, so that we can pass a `Colletion finder(Query)` lambda expression?

ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
rbygrave: Happy to just use LinkedHashSet here.  No need to add HashSet use here imo.

PR #2929, opened 2023-01-12
#2773 Fix wrong query in combination with fetch & where

rPraml:
Hello Rob,

we may have a fix for #2773 

we debugged and noticed, that combinations of joins between 'fetch' & 'where' will not always work.

By trial&error we found out, that deleting of
```java
          if (selectIncludes.contains(parentPropertyName)) {
            // parent already handled by select
            return childJoin;
          }
```
will solve the problem.... and will not break any other test.

I admit that I don't quite understand the fix. Maybe you can take another look to see if this is correct.

cheers
Roland

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
@@ -677,11 +677,6 @@ public final class SqlTreeBuilder {
         } else {
           // look in register ...
           String parentPropertyName = includeProp.substring(0, dotPos);
-          if (selectIncludes.contains(parentPropertyName)) {
-            // parent already handled by select
-            return childJoin;
-          }
-
           SqlTreeNodeExtraJoin parentJoin = joinRegister.get(parentPropertyName);
           if (parentJoin == null) {
             // we need to create this the parent implicitly...
--- a/ebean-test/src/test/java/org/tests/model/family/ChildPerson.java
+++ b/ebean-test/src/test/java/org/tests/model/family/ChildPerson.java
@@ -29,7 +29,7 @@ public class ChildPerson extends InheritablePerson {
   @Formula(select = "coalesce(${ta}.address, j1.address, j2.address)", join = PARENTS_JOIN)
   private String effectiveAddress;
 
-  @Formula(select = "coalesce(${ta}.some_bean_id, j1.some_bean_id, j2.some_bean_id)")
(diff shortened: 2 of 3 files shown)


rbygrave:
Yup cool, I'll try to have a look tomorrow night.

rbygrave:
Note: In `DefaultDbSqlContext.addJoin()` there is the logic below that ensures that we only add a particular join (based on joinKey below).  The effect of removing the "// parent already handled by select" return is that we get "duplicate joins" defined in the SqlTreeNodeExtraJoin ... but that doesn't matter for the existing tests etc because the "duplicate join" is de-duplicated by this logic in `DefaultDbSqlContext.addJoin()`.

```java
    String joinKey = table + "-" + a1 + "-" + a2;
    if (tableJoins.contains(joinKey)) {
      return;
    }
```

Hence removing that logic doesn't seem to break any existing tests (we are getting extra joins defined in the SqlTree but those are being de-duplicated in `DefaultDbSqlContext.addJoin()`.

So with that, it's now a question of looking at the failing test(s).

rbygrave:
So I'm looking at the test case and what is being attempted here with `@Formula` and the joins part and ... 

... to take a step back and look at the commented out: 
`@Coalesce({ "familyName", "parent.familyName", "parent.parent.familyName" })`

Well Ebean does not support that today but it almost does, in that it can _almost_ do:

`.select("name, coalesce(someBean, parent.someBean) as effectiveBean")`

For example, today we _can_ do:

`.select("name, coalesce(someBean, 42) as effectiveBean")`

The reason why I am saying this is because those formula joins look fragile wrt the extent that they are being used with this test. This "fix" to me is potentially fragile because there looks like an ordering / precedence issue wrt the formula joins and the "extra joins" and the fix proposed is more "lucky" than "good".

Stepping back, to me the approach taken here wrt using the formula joins like this doesn't look ideal due to the dynamic nature of the joins that are desired and `coalesce(someBean, parent.someBean)` looks more robust longer term to me.

rbygrave:
My thoughts are:
- `@Formula` should really have been called `@SqlFormula` darn it
- We _should_ be able to use logical property names and do `@LogicalFormula("coalese(familyName, parent.familyName, parent.parent.familyName")` ... and Ebean automatically extracts that the "extra join paths" needed being "parent" and "parent.parent" in this case ... and that the logical names get translated with the appropriate table alias' being used (and it's nice and dynamic and imo more robust handling different fetch/join scenarios - dynamically using left join etc).

... and then ideally we could have called  `@LogicalFormula`  _`@Formula`_

rPraml:
Closing this, as it is a duplicate of #2773

PR #2930, opened 2023-01-13
BUG: @OrderBy on chained properties do not work with distinct (testcase)

rPraml:
We discovered a bug, when a property has a chained orderBy (e.g. `@OrderBy("ref.name")`), it will not work in combination with "distinct".

Test fails with:

```
Error:    TestOrderByWithDistinct.testOrderByOnPropWithDistinct:66 » Persistence Query threw SQLException:
Order by expression "T2.NAME" must be in the result list in this case; 
SQL statement:
select distinct t0.id, t0.parent_id, t0.ref_id, t1.id, t1.parent_id, t1.ref_id 
from e_basic_tree t0 
left join e_basic_tree t1 on t1.parent_id = t0.id join e_basic_tree u1 on u1.parent_id = t0.id 
join e_basic u2 on u2.id = u1.ref_id left join e_basic t2 on t2.id = t1.ref_id where u2.status = ? 
order by t0.id, t2.name 
```

I'll provide a fix with the next commit

### DIFF
--- a/ebean-api/src/main/java/io/ebean/OrderBy.java
+++ b/ebean-api/src/main/java/io/ebean/OrderBy.java
@@ -232,18 +232,6 @@ public class OrderBy<T> implements Serializable {
   }
 
   /**
-   * Return true if this order by can be used in select clause.
-   */
-  public boolean supportsSelect() {
-    for (Property property : list) {
-      if (!property.supportsSelect()) {
-        return false;
-      }
-    }
-    return true;
-  }
-
-  /**
    * A property and its ascending descending order.
    */
   public static class Property implements Serializable {
@@ -403,12 +391,6 @@ public class OrderBy<T> implements Serializable {
       this.ascending = ascending;
     }
 
-    /**
-     * Support use in select clause if no collation or nulls ordering.
-     */
-    boolean supportsSelect() {
-      return nulls == null;
-    }
   }
 
   private void parse(String orderByClause) {
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
@@ -585,16 +585,11 @@ final class CQueryBuilder {
         }
         if (distinct && dbOrderBy != null) {
           // add the orderBy columns to the select clause (due to distinct)
-          final OrderBy<?> orderBy = query.getOrderBy();
-          if (orderBy != null && orderBy.supportsSelect()) {
-            String trimmed = DbOrderByTrim.trim(dbOrderBy);
-            if (query.isSingleAttribute() && select.selectSql().startsWith(trimmed)) {
-              // NOP, already in SQL
-              // TODO: what to do if we select("id").orderBy("prop,id")?
-              // Can we live with a query like "select t0.id, t0.prop, t0.id from"
-              // or should we elliminate the second "t0.id" from select
-            } else {
-              sb.append(", ").append(trimmed);
+          String[] tokens = DbOrderByTrim.trim(dbOrderBy).split(",");
+          for (String token : tokens) {
+            token = token.trim();
+            if (!DbOrderByTrim.contains(select.selectSql(), token)) {
+              sb.append(", ").append(token);
             }
           }
         }
@@ -643,7 +638,7 @@ final class CQueryBuilder {
     private void appendHistoryAsOfPredicate() {
       if (query.isAsOfBaseTable()) {
         query.incrementAsOfTableCount();
-        if (!historySupport.isStandardsBased()){
+        if (!historySupport.isStandardsBased()) {
           appendAndOrWhere();
           sb.append(historySupport.asOfPredicate(request.baseTableAlias()));
         }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/DbOrderByTrim.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/DbOrderByTrim.java
@@ -4,7 +4,7 @@ import java.util.regex.Pattern;
 
 final class DbOrderByTrim {
 
-  private static final Pattern orderByTrim = Pattern.compile("(?i) asc\\b| desc\\b|\\b nulls first\\b|\\b nulls last\\b");
+  private static final Pattern orderByTrim = Pattern.compile("(?i) asc\\b| desc\\b| nulls first\\b| nulls last\\b");
 
   /**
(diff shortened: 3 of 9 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/query/CQueryBuilder.java
```diff
@@ -585,16 +585,11 @@ private void appendSelect() {
...
-          if (orderBy != null && orderBy.supportsSelect()) {
-            String trimmed = DbOrderByTrim.trim(dbOrderBy);
-            if (query.isSingleAttribute() && select.selectSql().startsWith(trimmed)) {
-              // NOP, already in SQL
-              // TODO: what to do if we select("id").orderBy("prop,id")?
-              // Can we live with a query like "select t0.id, t0.prop, t0.id from"
-              // or should we elliminate the second "t0.id" from select
-            } else {
-              sb.append(", ").append(trimmed);
+          String[] tokens = DbOrderByTrim.trim(dbOrderBy).split(",");
```
rPraml, line 588: here, we do not longer check `query.getOrderBy()`, but we check only `dbOrderBy` which is effectively added also by the old code. It seems, that `supportsSelect` is no longer used.

> Can we live with a query like "select t0.id, t0.prop, t0.id from or should we elliminate the second "t0.id" from select

the new code adds only the tokens, that are not yet part of the `selectSql`

ebean-core/src/main/java/io/ebeaninternal/server/query/DbOrderByTrim.java
```diff
@@ -4,7 +4,7 @@
 
 final class DbOrderByTrim {
 
-  private static final Pattern orderByTrim = Pattern.compile("(?i) asc\\b| desc\\b|\\b nulls first\\b|\\b nulls last\\b");
+  private static final Pattern orderByTrim = Pattern.compile("(?i) asc\\b| desc\\b| nulls first\\b| nulls last\\b");
```
rPraml, line 7: this will not remove `.order("min(customer) asc nulls last")` properly.

rPraml: CHECKME: 
See TestOrderByParse: What about `nulls high` & `nulls low`... is this valid SQL?
What about collations?

ebean-test/src/test/java/org/tests/query/orderby/TestOrderByWithDistinct.java
```diff
@@ -89,12 +119,12 @@ public void testOrderByWithDistinct() {
...
-		 * https://groups.google.com/forum/?fromgroups=#!topic/ebean/uuvi1btdCDQ%5B1-25-false%5D
-		 *
-		 * This test exposes what may be a general problem with columns required by the order by phrase being omitted from the select.
-		 * I'm not sure this exposes all causes of the problem.
-		 */
+     * Original conversation:
+     * https://groups.google.com/forum/?fromgroups=#!topic/ebean/uuvi1btdCDQ%5B1-25-false%5D
+     *
+     * This test exposes what may be a general problem with columns required by the order by phrase being omitted from the select.
+     * I'm not sure this exposes all causes of the problem.
```
rPraml, line 126: Maybe we've found new ones 😆

ebean-test/src/test/java/org/tests/query/orderby/TestOrderByWithDistinctTake2.java
```diff
@@ -60,9 +60,9 @@ public void testWithAscAndDesc() {
     String generatedSql = sqlOf(query);
 
     if (platformDistinctOn()) {
-      assertThat(generatedSql).contains("select distinct on (t0.name, t0.id) t0.id, t0.name, t0.id");
+      assertThat(generatedSql).contains("select distinct on (t0.name, t0.id) t0.id, t0.name");
     } else {
-      assertThat(generatedSql).contains("select distinct t0.id, t0.name, t0.id");
```
rPraml, line 65: This is explainable, because t0.id was twice in the select

ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
```diff
@@ -172,7 +172,7 @@ static String mergeOnDistinct(String idCols, String dbOrderBy) {
     String[] split = idCols.split(",");
     for (String col : split) {
       col = col.trim();
-      if (!dbOrderBy.contains(col)) {
+      if (!DbOrderByTrim.contains(dbOrderBy, col)) {
```
rPraml, line 175: I think, this was not so good before. What if dbOrderBy = `t0.name_long` and col = `t0.name`

ebean-api/src/main/java/io/ebean/OrderBy.java
```diff
@@ -231,18 +231,6 @@ public OrderBy<T> clear() {
     return this;
   }
 
-  /**
-   * Return true if this order by can be used in select clause.
-   */
-  public boolean supportsSelect() {
```
rPraml, line 237: dead code

ebean-core/src/main/java/io/ebeaninternal/server/query/DbOrderByTrim.java
```diff
@@ -14,4 +14,21 @@ static String trim(String dbOrderBy) {
...
+   * <p>
+   * SQL is normally comma separated: "t0.id, t1.id, t2.name"
+   */
+  static boolean contains(String sql, String column) {
+    if (sql.endsWith(column)) { // simplest way. sql ends with the column
+      return true;
+    } else if (sql.contains(column)) {
+      // We need to check, if it is really the correct column,
+      // sql="t0.name_short, t0.name_long" will match on column="t0.name"
+      return sql.contains(column + ",");
```
rbygrave, line 28: Just pondering if we want to get the index position of column and then check the following char == ',' rather than perform the string contains twice (+ a string concat).

Otherwise looks great !!

rbygrave (approved):

ebean-core/src/main/java/io/ebeaninternal/server/query/DbOrderByTrim.java
rbygrave: My comment about using index position is misplaced - this looks as good as we will get.

I'm happy here.

ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeBuilder.java
rbygrave: Agreed.

PR #2937, opened 2023-01-19
[In progress] Optimized dependent table handling (RPr)

rPraml:
*This is work in progress and may be splitted in separate PRs*

In this PR I'll try to improve the cache-handling in ebean.
- optimize "dependantTable" handling
- when both QueryCache and BeanCache is enabled, always check the queryCache first
- ...


ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
```diff
@@ -68,6 +70,7 @@ public boolean isDeleteByStatement() {
     } else {
       // delete by ids due to cascading delete needs
       queryPlanKey = query.setDeleteByIdsPlan();
+      queryPlan = null;
```
rPraml, line 73: when the queryPlanKey changes, we must throw away also the cached queryPlan

ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
```diff
@@ -172,10 +175,24 @@ private void adapterPreQuery() {
    */
   @Override
   public void prepareQuery() {
-    secondaryQueries = query.convertJoins();
-    beanDescriptor.prepareQuery(query);
-    adapterPreQuery();
-    queryPlanKey = query.prepare(this);
+    if (!prepared) {
```
rPraml, line 178: prepareQuery is only executed once now

ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
```diff
@@ -469,7 +486,10 @@ public BeanPropertyAssocMany<?> manyPropertyForOrderBy() {
    * query plan for this query exists.
    */
   public CQueryPlan queryPlan() {
-    return beanDescriptor.queryPlan(queryPlanKey);
+    if (queryPlan == null) {
+      queryPlan = beanDescriptor.queryPlan(queryPlanKey);
```
rPraml, line 490: here we cache the current query plan, that made the "dependentTables" refactoring possible

ebean-test/src/test/java/org/tests/basic/TestDeleteByIdCollection.java
```diff
@@ -41,8 +43,10 @@ public void test() {
     DB.deleteAll(Customer.class, ids);
     awaitL2Cache();
 
+    LoggedSql.start();
     c0Back = DB.find(Customer.class, c0.getId());
     c1Back = DB.find(Customer.class, "" + c1.getId());
+    assertThat(LoggedSql.stop()).isEmpty();
```
rPraml, line 49: This should not hit the DB (but it currently does on master branch)

rPraml (commented):
Hello @rbygrave 

I'm finished with this change. I tried to separate this PR in easily reviewable commits. 
Can you review this as one PR or should I create one PR after the other.

What do you think about the change of first hitting the (faster) QueryCache and then the (slower) BeanCache?

PR #2943, opened 2023-01-23
FEATURE: Disable overwriting generated properties at transaction level

rPraml:
We want to backup & restore some data by exporting it to JSON. For this use case we need to preserve

- whenCreated
- whoCreated
- whenModified
- whoModified

This PR allows to disable the update of these properties for a particular transaction

### DIFF
--- a/ebean-api/src/main/java/io/ebean/Transaction.java
+++ b/ebean-api/src/main/java/io/ebean/Transaction.java
@@ -261,6 +261,12 @@ public interface Transaction extends AutoCloseable {
   void setUpdateAllLoadedProperties(boolean updateAllLoadedProperties);
 
   /**
+   * If set to false (default is true) generated propertes are only set, if it is the version property or have a null value.
+   * This may be useful in backup & restore scenarios, if you want set WhenCreated/WhenModified.
+   */
+  void setOverwriteGeneratedProperties(boolean overwriteGeneratedProperties);
+
+  /**
    * Set if the L2 cache should be skipped for "find by id" and "find by natural key" queries.
    * <p>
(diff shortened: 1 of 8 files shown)


rbygrave:
I like the feature. 

When I first read it I expect the opposite meaning for  `overwriteGeneratedProperties` = `true` ... which to me means I the developer get to override the generated properties. I want to set it to _true_ to enable me the developer to override the values of the generated properties.  Currently in the PR it has the opposite meaning.

I'll ponder this ...

rPraml:
That depends on which side you look at it from:
- do the developer want to overwrite the generated propertes (and ebean shouldn't set them back)
- or should NOT ebean overwrite previously changed propertes.

I agree, after your comment, the name is not the best. I'm sure we'll find an other one
- noUpdateGenerated (butIfTheyAreNullSetThem) - also a point of view problem 
- disableGeneratedUpdate
- ...

jnehlmeier:
Is it allowed to toggle that switch on/off during the lifespan of the transaction or would that miserably cause problems? If it can be toggled on/off at any time then a `setter` name seems fine, e.g. `setAutoUpdateGeneratedFields(boolean)`. If we can not toggle the setting at any time then just `disableAutoUpdateGeneratedFields()` to disable the default setting.

Side note: It would be so cool if that whole generated fields thing would be extendible/configurable. For example I like `Who/WhenDeleted` as well, because soft-deletion is not a modification of data.

rPraml:
Generated properties are updated during save, so it is possible to modify it at nearly any time
(you might not change it in a persist callback or in an other thread).

> `setAutoUpdateGeneratedFields(boolean)`

I like that name

rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3799

PR #2945, opened 2023-01-27
NEW: Support for lazy add on BeanList

rPraml:
When just adding new entries to an existing BeanList, there is no need to load the underlying collection.

this should provide a performance benefit, where you just want to add new entries to a BeanList


rPraml:
This seems to be buggy, please do not merge, yet.

rPraml:
We tested this a whil in our production code. It works with https://github.com/ebean-orm/ebean/pull/2945/commits/a7e07ae2c9edcde73601df550956bd35ddffb932

Issue #2946, opened 2023-01-27
Add support for specifying select clause of RawSQL query


buchtajz:
# Rob Edit:

This is NOT a bug, Ebean never supported specifying a select clause on a RawSql query.

This is a change adding support for specifying the select clause on RawSql. The side-effect of this change is that `findCount` will look to produce different sql than before by only including the Id column/property in the select clause - it will only do this IF the Id column/property is mapped in the RawSql query and otherwise should produce the same query as before.

-----------

## Expected behavior

query.select().asDto() working as in older ebean version (we was used 11.x)

## Actual behavior
query.select().asDto()   ignores select parameter and returned entity is query basic class instead of Dto class

final sql query is 

`select id, name, anniversary, city from (select c.id ...`

should be 
`select count(1) as count, sum(id) as sum from (select c.id ...`

### Steps to reproduce
We dynamically create Report java classes based on user specified SQL query.
So I simplify it to this example RawQuery.

```java
    RawSql rawSql = RawSqlBuilder.parse("select id, name, anniversary, city from (select c.id, c.name, c.anniversary, a.city from o_customer c left join o_address a ON a.id = c.billing_address_id order by c.id) w").create();

    var query = DB.find(SampleReport.class);
    query.setRawSql(rawSql);
    query.setMaxRows(10);

    List<SampleReport> list = query.findList(); // this returns 4 rows, ok works fine
    assertNotNull(list);

    String sumSql = "COUNT(1)::Long AS count, SUM(id) AS sum";

    //query.select(sumSql);   // tried both versions
    query.select(FetchGroup.of(SampleReport.class, sumSql));
    query.asDto(QuerySumResponse.class);

    var countSum = query.findOneOrEmpty().orElseThrow();
    // generates exception, see bellow

    // commented for compile error due class conflict
    //assertThat((countSum instanceof QuerySumResponse));  

```

```console
  javax.persistence.NonUniqueResultException: Unique expecting 0 or 1 results but got 4
```

buchtajz:
Unit test pull request
[https://github.com/ebean-orm/ebean/pull/2947](https://github.com/ebean-orm/ebean/pull/2947)

rbygrave:
Can you tell me exactly which version of Ebean 11.x that this worked in?

buchtajz:
Hi, 11.17.5 worked

buchtajz:
[...]

PR #2947, opened 2023-01-27
Unit test and fix for issue #2946

buchtajz:


rbygrave:
This PR is effectively replaced by https://github.com/ebean-orm/ebean/pull/2996

PR #2956, opened 2023-02-02
FEATURE: Tenant partitioned caches could improve cache-hit ratio

rPraml:
Hello @rbygrave,

this is more or less a proof-of-concept, where I need your feedback

we have a multi-tenant application with ~ 50 tenants. I noticed, that we get a lot of 'cache clears' from other customers.
We cache a lot of things in our applications (user-directory, settings, ...) and if one customer updates the settings e.g. it will invalidate the caches for all other. So I tried to experimet with tenant partitioned caches, which means, each tenant has its own set of caches.

On the other hand, caches had to be created dynamically and would - at least in current implementation - stay there forever, also if the tenant does not longer exist.

I don't have exact numbers about this improvement, but do you think, this could be useful?


rbygrave:
Updated into and replaced by https://github.com/ebean-orm/ebean/pull/3809

Issue #2961, opened 2023-02-10
Followup for #2952 - @OneToMany + orphanRemoval + @SoftDelete + non-BeanCollection collection results in hard deletes


rbygrave:
As noted in comments in https://github.com/ebean-orm/ebean/issues/2952

In the internals of SaveManyBeans we have:

1. **_BUG:_** the deleteByParentId is **hard delete** and does not care for soft delete
2. **_YUK:_** internally we have 3 ways of performing the orphan removal when we really want 2

This change fixes the BUG and fixes the YUK. It does this by removing the special case at: https://github.com/ebean-orm/ebean/blob/ebean-parent-13.11.3/ebean-core/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java#L347-L350 ... and replacing it with the more common orphan removal code used when we do not have BeanCollection modifications.

The result of this change is that in SaveManyBeans internals we get back to have 2 ways to remove orphans.

1. A BeanCollection with modifications: Orphans explicitly deleted using the known elements removed from the collection
2. All other cases: Orphans as everything _NOT in the collection_ that is going to be updated

Issue #2952, opened 2023-01-30
@OneToMany orphanRemoval not occurring (was: insert instead update on changing  @OneToMany field)


buchtajz:
## Expected behavior
Should work as in 11.x version

DB.save/update  should use SQL update not insert

## Actual behavior

When changed @OneToMany List via setter, ebean generate insert sql query.

When using getter and List clear/ addAll, it is using update.

**EDIT by Rob:** 
No it is _NOT using update_ here with clear/addAll. It IS using an extra delete for the orphans.

### Steps to reproduce

```java
    // this forces insert and throws exception due primary key conflict
    persistedGoods.setAttachments(marshaledGoods.getAttachments());

    // this works good
    persistedGoods.getAttachments().clear();
    persistedGoods.getAttachments().addAll(marshaledGoods.getAttachments());

    // Note: This is working because we mutate the collection and so the orphans that are 
    // removed by the clear() is known and results in the extra delete for those orphans
```

```console
  io.ebean.DuplicateKeyException: Error when batch flush on sql: insert into attachment (id, goods_entity_id, name, version, when_created, when_modified, created_by, updated_by) values (?,?,?,?,?,?,?,?)
```
#

rbygrave:
Can you check the release notes of 12.1.1 - https://github.com/ebean-orm/ebean/releases/tag/ebean-12.1.1

I believe this is the change in behavior this issue is referring to. You can confirm that by comparing 12.1.1 and previous version which was 11.45.1

rbygrave:
We can also review this list: https://github.com/ebean-orm/ebean/issues?q=is%3Aissue+is%3Aclosed+label%3Areason-for-version-bump

Closed issues tagged with "reason-for-version-bump"

buchtajz:
Hi Rob, thanks for response.

I read these issues, it looks simillary but i think it's not the same problem. There is some correlation maybee.

There is difference on generated sql  setAttachments() vs. getAttachments().clear() + .add()

I added one more attachment to marshaled object to see difference

>    var marshaledGoods = DB.json().toBean(GoodsEntity.class, DB.json().toJson(goods));
>    marshaledGoods.getAttachments().add(attachment3);

the first 
>    setAttachments(marshaledGoods.getAttachments()))

on save generate
```
txn[] insert into attachment (id, goods_entity_id, name, version, when_created, when_modified, created_by, updated_by) values (?,?,?,?,?,?,?,?)
[...]

Issue #2127, opened 2020-12-08
Adding a new list of children to @OneToMany with orphanRemoval true does not delete existing orphans


almothafar:
I'm not sure if you consider this as an issue or not, but for me, it is a little annoying, I have a parent bean which has a list of children with `@OneToMany` relation, so when I update the list of children with totally new a new list, old children stay there, so let say I have "Company" and list of "Employees"

```java
@Entity
@Table(name = "Company", schema = "dbo")
public class Company extends Model {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    @Column(name = "Id")
    private Integer id;

    @Column(name = "Name")
    private Integer name;

    @OneToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST, CascadeType.REFRESH}, mappedBy = "request", orphanRemoval = true)
    private List<User> employees;
}
```

```java
@Entity
@Table(name = "User", schema = "dbo")
public class User extends Model {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    @Column(name = "Id")
    private Integer id;

    @Column(name = "Name")
    private Integer name;

    @ManyToOne(cascade = {CascadeType.MERGE, CascadeType.PERSIST, CascadeType.REFRESH})
    @JoinColumn(name = "CompanyId")
    @DbForeignKey(onDelete = ConstraintMode.CASCADE)
    private Company company;
}
```

so when first when I do:

```java
User employee1 = new User().setName("Employee 1");
User employee2 = new User().setName("Employee 2");
List<User> employees = new ArrayList<>();
employees.add(employee1);
employees.add(employee2);

Company company = new Company().setName("My Company").setEmployees(employees);
company.insert();
```

I get everything in the table correct, users inserted with company id and everything, but not if I get that company back, and add "a new list", old employees stay in that table:

```java
User employee1 = new User().setName("Employee 1");
User employee3 = new User().setName("Employee 3");
List<User> employees = new ArrayList<>();
employees.add(employee1);
employees.add(employee3);

Company company = DB.find(Company .class).eq("name", "My Company").findOne();
company.setEmployees(employees);
company.update();

```
I get in the database
```
1        Employee 1
2        Employee 2
3        Employee 1
4        Employee 3
```

## Expected behavior
To have 2 employees:
```
3        Employee 1
4        Employee 3
```

## Actual behavior
Users just stack up:
```
1        Employee 1
2        Employee 2
3        Employee 1
[...]

PR #2962, opened 2023-02-10
#2961 - Followup for #2952 - @OneToMany + orphanRemoval + @SoftDelete + non-BeanCollection collection results in hard deletes

rbygrave:
As noted in comments in #2952

In the internals of SaveManyBeans we have:

- BUG: the deleteByParentId is hard delete and does not care for soft delete
- YUK: internally we have 3 ways of performing the orphan removal when we really want 2

This change fixes the BUG and fixes the YUK. It does this by removing the special case at: https://github.com/ebean-orm/ebean/blob/ebean-parent-13.11.3/ebean-core/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java#L347-L350 ... and replacing it with the more common orphan removal code used when we do not have BeanCollection modifications.

The result of this change is that in SaveManyBeans internals we get back to have 2 ways to remove orphans.

- A BeanCollection with modifications: Orphans explicitly deleted using the known elements removed from the collection
- All other cases: Orphans as everything NOT in the collection that is going to be updated

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java
@@ -35,8 +35,8 @@ final class SaveManyBeans extends SaveManyBase {
   private final boolean untouchedBeanCollection;
   private final Collection<?> collection;
   private final boolean hasOrderColumn;
+  private final boolean forcedUpdate;
   private int sortOrder;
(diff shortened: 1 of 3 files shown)


rbygrave:
Also refer to previous issue https://github.com/ebean-orm/ebean/issues/2127 (which introduces that special case handling that we have now merged into the more normal handling of orphan removal for non-BeanCollection

PR #2980, opened 2023-02-22
FIX: do not format log messages, when no parameter was specified

rPraml:
How to reproduce:
if have a `@Where("someField = 'someValue'")` annoation, the CaptureLogger may remove the single quotes from the string, because it passes the `msg` argument to `MessageFormat`, although there are no parameters specified.

This PR changes the behaviour of the CaptureLogger to line up with the default logger (= if args are null or empty, don't format them)

### DIFF
--- a/ebean-test/src/test/resources/logback-test.xml
+++ b/ebean-test/src/test/resources/logback-test.xml
@@ -78,6 +78,7 @@
   <logger name="io.ebean.docker" level="DEBUG"/>
   <logger name="io.ebean.test" level="TRACE"/>
   <logger name="io.ebean.MarkedAsDeleted" level="DEBUG"/>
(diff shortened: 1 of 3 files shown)


ebean-test/src/test/resources/logback-test.xml
```diff
@@ -78,6 +78,7 @@
   <logger name="io.ebean.docker" level="DEBUG"/>
   <logger name="io.ebean.test" level="TRACE"/>
   <logger name="io.ebean.MarkedAsDeleted" level="DEBUG"/>
+  <logger name="io.ebean.test.loggerTest" level="DEBUG"/>
```
rPraml, line 81: we can remove this debug here. It was just verify that the console output matches with the assertion

rbygrave:
Ah yes, I see that now ... :)

Issue #3003, opened 2023-03-21
Where Raw clause with subquery generate wrong column names in sql query 


buchtajz:
## Expected behavior
Not change column names (aliases) in subquery.

## Actual behavior
Adds t0. prefix to column names when there is column with same name in parent query.

### Steps to reproduce

```java
  DB.find(Customer.class)
      .select("name")
      .where()
      .raw("id in (select c.id as id from o_customer c)")
      .findList();
```
generates invalid query

`select t0.id, t0.name from o_customer t0 where t0.id in (select c.id as t0.id from o_customer c)` 

and throws exception

```console
  Syntax error in SQL statement "select t0.id, t0.name from o_customer t0 where t0.id in (select c.id as t0[*].id from o_customer c)"; expected ",, FROM, WHERE, GROUP BY, HAVING, WINDOW, QUALIFY, INTERSECT, UNION, EXCEPT, MINUS, ORDER BY, OFFSET, FETCH, LIMIT, FOR, ,, )";
    SQL statement:
    select t0.id, t0.name from o_customer t0 where t0.id in (select c.id as t0.id from o_customer c) [42001-214] Bind values:[null] Query was:select t0.id, t0.name from o_customer t0 where t0.id in (select c.id as t0.id from o_customer c)
```
It's not limited only to id column and this simple example. Subquery can be complex CTE query, with many column aliases.

buchtajz:
- Subqueries are dynamic, written by customers (individually correct)
- they can be complex CTE queries - bug appears at CTE too 
   `WITH some_cte AS (SELECT p.xyz AS firstname FROM b JOIN p ... ), ...`
   is generated as 
   `WITH some_cte AS (SELECT p.xyz AS t0.firstname FROM b JOIN p ... ), ...`
  when there is conflict in column names in t0 table (t0.firstname exists), so query is invalid and throws exception

rbygrave:
> Subqueries are dynamic, written by customers (individually correct)

What do you mean here?  Do you mean a customer sends a String with SQL in it via some Rest API and the application just includes it in the query?

buchtajz:
Not so crazy security maddnes :)

It's not public available. Customer, or better some emloyees, can create sql query in company private application instance. 

Security is not problem there. This is part of bussines process modelling.

rbygrave:
Are you saying that you are unable to modify the SQL that makes up the subquery?  

That there is some code that does string concatenation to build the string expression that goes into the raw() expression?

[...]

PR #3004, opened 2023-03-21
failing test for Where raw clause subquery

buchtajz:
unit test for issue #3003


rbygrave:
We can close this as it is replaced by https://github.com/ebean-orm/ebean/pull/3009/commits/79fd689e9a84455da9b4760def06ed4c28085fd8

PR #3017, opened 2023-03-30
Initial DB2Luw9Platform using row_number() for row limiting sql

rbygrave:


platforms/db2/src/main/java/io/ebean/platform/db2/DB2Luw9Platform.java
```diff
@@ -0,0 +1,15 @@
+package io.ebean.platform.db2;
+
+import io.ebean.annotation.Platform;
+
+/**
+ * DB2 platform for Linux/Unix/Windows version 9.x.
+ */
+public class DB2Luw9Platform extends BaseDB2Platform {
```
rPraml, line 8: Note: There is still the `DB2LegacyPlatform` as fallback which has very restrictive limits to 18 characters. (For V7 and below).
I think it makes sense to include the limiters also there.

platforms/db2/src/main/java/io/ebean/platform/db2/DB2RowNumberBasicLimiter.java
```diff
@@ -0,0 +1,31 @@
...
+ * Row limiter for Oracle 9,10,11 using rownum.
+ */
+final class DB2RowNumberBasicLimiter implements BasicSqlLimiter {
+
+  @Override
+  public String limit(String dbSql, int firstRow, int maxRows) {
+    if (firstRow < 1 && maxRows < 1) {
+      return dbSql;
+    }
+    StringBuilder sb = new StringBuilder(60 + dbSql.length());
```
rPraml, line 15:
```suggestion
    StringBuilder sb = new StringBuilder(120 + dbSql.length());
```
60 is not enough here. The appended constant strings has at least 89 chars

Issue #3026, opened 2023-04-13
Postgis with cache is not working


Ryszard-Trojnacki:
## Expected behavior
I have class with Postgis fields:
```java
public class Robot {
    @Column(name = "id", unique = true, nullable = false)
    private long id;
    ...
    @Column(name = "position") @Basic(fetch = FetchType.EAGER)
    private Point position;
    ...
}
```
when I load this bean by Id (second time):
```java
private Database db;

db.find(Robot.class, id);
```
`position` field is null but it is present in database.

## Actual behavior
Column/field in database with valid not null value is returned as `null` if queried for second time (from cache).

### Steps to reproduce
Start Ebean with enabled cache (default) and get something from database with postgis Point value that is not null.
For the first time it will return valid (not null) value.
For the second time it will return null because of Ebean cache.

I tried to find what is wrong by debugging, but it seems that call:
```
DefaultServer.findId:1028
DefaultServer.find:930
DefaultServer.find:920
```
returns bean from cache without filled `position` value `DefaultServer`:
```java
  private <T> T findId(Query<T> query, @Nullable Transaction transaction) {
    SpiQuery<T> spiQuery = (SpiQuery<T>) query;
    spiQuery.setType(Type.BEAN);
    if (SpiQuery.Mode.NORMAL == spiQuery.getMode() && !spiQuery.isForceHitDatabase()) {
      // See if we can skip doing the fetch completely by getting the bean from the
      // persistence context or the bean cache
      T bean = findIdCheckPersistenceContextAndCache(transaction, spiQuery, spiQuery.getId());
      if (bean != null) {
        return bean;
      }
    }
```
Ebean returns bean from `findIdCheckPersistenceContextAndCache`.

I checked `DefaultServerCache` and it show position as null:
![image](https://user-images.githubusercontent.com/7201935/231700740-3ba18323-153a-441a-a4ec-62690302b692.png)

I also checked put call to cache `DefaultServerCache.java`:
```java
  @Override
  public void put(Object key, Object value) {
    map.put(key, new SoftReference<>(new CacheEntry(key, value)));
    putCount.increment();
    if (mutationCounter.incrementAndGet() > trimOnPut) {
      runEviction();
    }
  }
```
and data that is put here is **always** without `position` field (`position==null`):
![image](https://user-images.githubusercontent.com/7201935/231704389-5ffd28f0-521d-4b8d-9d40-4182ddec7d52.png)

[...]

PR #3036, opened 2023-04-21
Implemented missing cache functions for Postgis module.

Ryszard-Trojnacki:
This is a fix for #3026. It resolves the problem. JUnit tests confirms this.

There is also a problem with `latte` package mapping. This PR doesn't address `latte` package mappings.

### DIFF
--- a/ebean-postgis/src/main/java/io/ebean/postgis/ScalarTypePgisLineString.java
+++ b/ebean-postgis/src/main/java/io/ebean/postgis/ScalarTypePgisLineString.java
@@ -3,10 +3,20 @@ package io.ebean.postgis;
 import io.ebean.config.dbplatform.ExtraDbTypes;
 import org.postgis.LineString;
 
+import java.sql.SQLException;
+
 public class ScalarTypePgisLineString extends ScalarTypePgisBase<LineString> {
 
   public ScalarTypePgisLineString() {
     super(ExtraDbTypes.LINESTRING, LineString.class);
   }
 
+  @Override
+  public LineString parse(String value) {
+    try {
+      return new LineString(value);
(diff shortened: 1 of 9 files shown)


ebean-postgis/src/main/java/io/ebean/postgis/ScalarTypePgisLineString.java
```diff
@@ -3,10 +3,20 @@
...
   public ScalarTypePgisLineString() {
     super(ExtraDbTypes.LINESTRING, LineString.class);
   }
 
+  @Override
+  public LineString parse(String value) {
+    try {
+      return new LineString(value);
+    } catch (SQLException e) {
+      throw new IllegalStateException(e);
```
rPraml, line 19: I was wondering if it would make sense to include the erroneous value here as message argument, but as far as I can see LineString is already throwing a SqlException with the value included in the message

rPraml (commented):
It makes sense to me.

Side note:
I noticed that jsonRead/jsonWrite (and readData/writeData) is also not yet implemented.
If you mean, it makes sense to implement this, you may find a reference implementation [here](https://github.com/mayconbordin/postgis-geojson/tree/master/src/main/java/org/postgis/geojson) - but this should not be done in this PR

ebean-postgis/src/main/java/io/ebean/postgis/ScalarTypePgisLineString.java
Ryszard-Trojnacki: As I think, this error should never occur. String `value` passed to constructor is generated with this same code (`LineString#toString()`). This could happen, when WTK string would be from external source, but this is not the case.

PR #3041, opened 2023-04-26
IMPROVED: checkUniqueness supports queryCache and skipClean

rPraml:
Hello Rob, with #1303 I implemented the checkUniqueness method, that we use in validation to verify that a new or modified bean does not violate unique constraints of the database.

This is neccessary to provide good user feedback in our UI

As already mentioned in #1303, this could be a performance killer, especially if you do a mass import/update of some entities, where unique constraints are involved. To minimize this, I tried to do two improvements
- `skipClean` if you load a bean from database and do not touch the unique properties, they need not to be validated
- `useQueryCache` I use an `exists` query instead of `findCount` as this may be slightly faster. Adittionally I enabe the queryCache on that query. This eliminates subsequent DB hits if the bean gets validated multiple times

### DIFF
--- a/ebean-api/src/main/java/io/ebean/DB.java
+++ b/ebean-api/src/main/java/io/ebean/DB.java
@@ -458,8 +458,8 @@ public final class DB {
   /**
    * Same as {@link #checkUniqueness(Object)} but with given transaction.
    */
-  public static Set<Property> checkUniqueness(Object bean, Transaction transaction) {
-    return getDefault().checkUniqueness(bean, transaction);
+  public static Set<Property> checkUniqueness(Object bean, Transaction transaction, boolean useQueryCache, boolean skipClean) {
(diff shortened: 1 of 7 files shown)


rbygrave:
Looks good. The javadoc for the public API needs filling in for the new parameters and this is added in a breaking API way.

rPraml:
Hello @rob-bygrave 
I've fixed the breaking API here https://github.com/ebean-orm/ebean/pull/3041/commits/723c73e3d9cc6c2f489f6db4db336314bb272430

FYI: I also changed the default of "skipClean" to true [here](https://github.com/ebean-orm/ebean/pull/3041/commits/93128d9eb70271ab68c6b2d34b3f04c1d94f13e6), as it makes no sense to check clean properties against unique-violations (the property comes from the database and will not be updated anyway when saving that bean)
So, for people that use the existing checkUnique methods, there could be some performance benefits.

PR #3042, opened 2023-04-27
Transaction performance on logical empty transactions

rPraml:
## Actual behavior

When there is a logical empty transaction (e.g. when all queries hit the cache), ebean performs various "actions" like setAutoCommit/setIsolationLevel/commit/rollback against that transaction

Depending on the JDBC-driver, these actions may communicate over tcp/ip with the server and are 'slow' as data has to go over the network wire.

## Expected behavior

When you have transactional blocks and have effective no DB communication, there should be also no communication with the JDBC driver, as this may decrease performance

### Steps to reproduce

```java
    try (Transaction txn = DB.beginTransaction(TxScope.requiresNew())) {
      DB.find(Customer.class).setUseQueryCache(true).where().eq("name", "Rob").findOne();
      txn.commit();
    }
```
See provided test with measurements.
Depending on the DB backend, you may get runtimes, that depend on the factor 100 (or more if machine is really remote on LAN)

As Proof-Of-Concept I've written a small adapter for SpiTransaction, which initializes the transaction lazy and defers also connection retrieval from the connectionPool
I'm not sure, if this is really the right place. Possible other implementation ideas:

**Implement in PooledConnection**
It seems that commit/rollback affects the performance at most and JdbcTransaction.checkAutocommit a bit.
Retrieving connection from the pool seems to be fast 👍
PooledConnection could skip commit/rollback and may be cache the autocommit/isolation settings, if there was no statement executed

**Implement in JdbcTransaction**
JdbcTransaction detects already, if there were only read queries.
If it can detect `No-Query` / `Read-Query` / `Write-Query`, it can return the connection just to the pool.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionFactoryBuilder.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionFactoryBuilder.java
@@ -27,4 +27,8 @@ final class TransactionFactoryBuilder {
       }
     }
   }
+
+  static TransactionFactory lazy(TransactionFactory delegate) {
+    return new TransactionFactoryLazy(delegate);
+  }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionFactoryLazy.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionFactoryLazy.java
@@ -1,0 +1,553 @@
+package io.ebeaninternal.server.transaction;
+
+import io.ebean.ProfileLocation;
+import io.ebean.TransactionCallback;
+import io.ebean.annotation.DocStoreMode;
+import io.ebean.event.changelog.BeanChange;
+import io.ebean.event.changelog.ChangeSet;
+import io.ebeaninternal.api.SpiPersistenceContext;
+import io.ebeaninternal.api.SpiProfileTransactionEvent;
+import io.ebeaninternal.api.SpiTransaction;
+import io.ebeaninternal.api.TransactionEvent;
+import io.ebeaninternal.server.core.PersistDeferredRelationship;
+import io.ebeaninternal.server.core.PersistRequestBean;
+import io.ebeaninternal.server.persist.BatchControl;
+import io.ebeanservice.docstore.api.DocStoreTransaction;
+
+import javax.persistence.PersistenceException;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * Transaction factory used with multi-tenancy.
+ */
+class TransactionFactoryLazy extends TransactionFactory {
+
+  final TransactionFactory delegate;
+
+  TransactionFactoryLazy(TransactionFactory delegate) {
+    super(delegate.manager);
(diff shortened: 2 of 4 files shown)


rbygrave:
Hmm, ebean has readonly datasource support and this is what I would expect and hope people to use for this sort of case. 

So the PR code example is basically an anti-pattern and then looking to 'fix' the anti-pattern.  Why not use implicit transactions and the readonly datasource?

rPraml:
I agree, that this can be seen as an anti-pattern (why open a transaction and then do nothing) - but this was a workaround for the cache issues, we had in the past

rbygrave:
I feel like the focus should be on fixing the original issues first rather than the workaround per say.

rPraml:
We currently have issues with caches (but found workarounds for most of them, so it is ok to close this PR for now)

One question: Is it possible (or do you think it make sense) to configure when and how an implicit transaction is created?

See this example
```java
try (Transaction mainTxn = DB.beginTransaction()) {
    .... some code
    someFunction();
}

void someFunction() {
    cfg = DB.find(Configuration.class)
       .setUseQueryCache(true)
       .setImplicitTransactionScope(TxScope.requiresNew()) // if not found in cache - create new transaction and search
       .where() ....
       .findOne();
    ...
}
```

rbygrave:
> use an application managed cache instead of relying on Ebean

For me, I'd expect the answer to be approximately "cache invalidation". As I see it, generally filling and getting from any cache is typically pretty easy and the work / complexity etc comes when dealing with invalidating cache entries. For example, if invalidation is just "Time To Live" then any sort of cache is pretty easy. 

Apps will be using Ebean L2 caching because of the cache invalidation needs of the app are well matched to the built in cache invalidation that Ebean L2 caching has. At least, that is my expectation.


> See this example

I don't understand the example probably because key details are `...` there.  If I see `beginTransaction()` I do expect insert, update, or delete activity to occur. Based on READ_COMMITTED isolation level (expected default) there isn't otherwise a point to the explicit transaction via `beginTransaction()`.

So yeah, I'm pretty sure I don't understand the example BUT ....

> `.setImplicitTransactionScope(TxScope.requiresNew()) `

BUT ... If there is a database read replica [with replication lag] their can be the desire to do the opposite of this which is to _force_ an implicit query to use the master database and not implicitly use the read replica (but this is not your question).

rob-bygrave:
> updated the example code

Thanks. 

For AbstractImporter I don't see anything interesting. I'm going to skip that.

For ConfigProvider, yes I see.  Now if I refer you to the javadoc example for `Transaction.rollbackAndContinue()` we see an example use of catching DuplicateKeyException and handling it. The main thing to note is the use of 2 try blocks and not 1. The outer try for the transaction scope and the inner try for the catch of DuplicateKeyException.

```java
 try (Transaction txn = database.beginTransaction()) {
    try {      ...
      database.save(bean);
      database.flush();
    } catch (DuplicateKeyException e) {
      // carry on processing using the transaction
      txn.rollbackAndContinue();
      ...
    }
    txn.commit();
  }
```   

What I see with AbstractImporter is only 1 catch block when I expect to see 2 try blocks, and no use of rollbackAndContinue().

PR #3057, opened 2023-05-16
Property.value use intercept to trigger lazy load

rPraml:
Hello @rbygrave 

we use often `Property.value(bean)` in our code, as this is the common interface of all properties.
Today, we discovered, that this method does not trigger lazy load and just returns null in such case.

My question is now, is this a mistake or intended. If intended, why is there no `valueIntercept` in Property?

cheers
Roland


rob-bygrave:
Right, now that #3058 is merged we can merge this PR in as there is now no internal use of `Property.value()` [only used in tests] so effectively only code will be using `Property.value()` and it's reasonable that it SHOULD invoke lazy loading if required when used this way.

PR #3058, opened 2023-05-17
No functional change, change use of BeanProperty.value() to BeanProperty.getValue()

rob-bygrave:
Noting that value() currently calls through to getValue() hence this is no real change. This is done because value() is also a public method Property.value() [and this should probably change to use getValueIntercept() rather than getValue()]


rob-bygrave:
Refer to https://github.com/ebean-orm/ebean/pull/3057 which was the trigger for making this change

PR #3060, opened 2023-05-17
Improved Null-checking annotations

rPraml:
Hello @rbygrave, in the following example

```java
String value = query.findSingleAttribute()
if (value == null) ...
```
IntelliJ complains, that `value` could never be null, which is not true.

In this PR, I tried to improve the API and added more `@NonNullApi` / `@Nullable` annotations.

I moved some null checks, so this PR is only 99% no code change.

Please take a look, if the annotations makes sense this way.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/BeanFinder.java
+++ b/ebean-api/src/main/java/io/ebean/BeanFinder.java
@@ -2,7 +2,6 @@ package io.ebean;
 
 import io.avaje.lang.NonNullApi;
 import io.avaje.lang.Nullable;
-
 import java.util.List;
 import java.util.Optional;
 
@@ -64,7 +63,6 @@ public abstract class BeanFinder<I,T> {
   /**
    * Return the current transaction.
    */
-  @Nullable
   public Transaction currentTransaction() {
     return db().currentTransaction();
   }
--- a/ebean-api/src/main/java/io/ebean/BeanFinder.java
+++ b/ebean-api/src/main/java/io/ebean/BeanFinder.java
@@ -2,6 +2,7 @@ package io.ebean;
 
 import io.avaje.lang.NonNullApi;
 import io.avaje.lang.Nullable;
+
 import java.util.List;
 import java.util.Optional;
 
@@ -63,6 +64,7 @@ public abstract class BeanFinder<I,T> {
   /**
    * Return the current transaction.
    */
+  @Nullable
   public Transaction currentTransaction() {
     return db().currentTransaction();
   }
--- a/ebean-api/src/main/java/io/ebean/DB.java
+++ b/ebean-api/src/main/java/io/ebean/DB.java
@@ -126,13 +126,13 @@ public final class DB {
    * Return the next identity value for a given bean type.
    * <p>
    * This will only work when a IdGenerator is on this bean type such as a DB
-   * sequence or UUID. Otherwise, <code>null</code> is returned.
+   * sequence or UUID. (Otherwise <code>null</code> is returned. However, this
+   * method declared as <code>NonNull</code>, as this is the default use case)
    * <p>
    * For DB's supporting getGeneratedKeys and sequences such as Oracle10 you do
    * not need to use this method generally. It is made available for more
    * complex cases where it is useful to get an ID prior to some processing.
    */
-  @Nullable
   public static Object nextId(Class<?> beanType) {
     return getDefault().nextId(beanType);
   }
@@ -312,7 +312,7 @@ public final class DB {
    * Mark the current transaction as rollback only.
    */
   public static void setRollbackOnly() {
-    Objects.requireNonNull(currentTransaction(),"no currentTransaction").setRollbackOnly();
+    Objects.requireNonNull(currentTransaction(), "no currentTransaction").setRollbackOnly();
   }
 
   /**
--- a/ebean-api/src/main/java/io/ebean/DB.java
+++ b/ebean-api/src/main/java/io/ebean/DB.java
@@ -9,10 +9,7 @@ import io.ebean.text.json.JsonContext;
 
 import javax.persistence.OptimisticLockException;
 import javax.persistence.PersistenceException;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.Callable;
 
 /**
--- a/ebean-api/src/main/java/io/ebean/DB.java
+++ b/ebean-api/src/main/java/io/ebean/DB.java
@@ -129,12 +129,13 @@ public final class DB {
    * Return the next identity value for a given bean type.
    * <p>
    * This will only work when a IdGenerator is on this bean type such as a DB
-   * sequence or UUID.
+   * sequence or UUID. Otherwise, <code>null</code> is returned.
    * <p>
    * For DB's supporting getGeneratedKeys and sequences such as Oracle10 you do
    * not need to use this method generally. It is made available for more
    * complex cases where it is useful to get an ID prior to some processing.
    */
+  @Nullable
   public static Object nextId(Class<?> beanType) {
     return getDefault().nextId(beanType);
   }
@@ -232,6 +233,7 @@ public final class DB {
   /**
    * Returns the current transaction or null if there is no current transaction in scope.
    */
+  @Nullable
   public static Transaction currentTransaction() {
     return getDefault().currentTransaction();
   }
@@ -251,7 +253,7 @@ public final class DB {
    * </ul>
    */
   public static void flush() {
-    currentTransaction().flush();
+    getDefault().flush();
   }
 
   /**
@@ -313,7 +315,7 @@ public final class DB {
    * Mark the current transaction as rollback only.
    */
   public static void setRollbackOnly() {
-    getDefault().currentTransaction().setRollbackOnly();
+    Objects.requireNonNull(currentTransaction(),"no currentTransaction").setRollbackOnly();
   }
 
   /**
@@ -1085,7 +1087,9 @@ public final class DB {
 
   /**
    * Return the value of the Id property for a given bean.
+   * This can be null, if the bean has no id set or no id property at all.
    */
+  @Nullable
   public static Object beanId(Object bean) {
     return getDefault().beanId(bean);
   }
--- a/ebean-api/src/main/java/io/ebean/Database.java
+++ b/ebean-api/src/main/java/io/ebean/Database.java
@@ -35,7 +35,7 @@ import java.util.concurrent.Callable;
  * <h5>The 'default' Database</h5>
  * <p>
  * One Database can be designated as the 'default' or 'primary' Database
- * (see {@link DatabaseConfig#setDefaultServer(boolean)}. Many methods on DB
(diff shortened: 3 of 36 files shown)


ebean-api/src/main/java/io/ebean/DB.java
```diff
@@ -251,7 +250,7 @@ public static Transaction currentTransaction() {
    * </ul>
    */
   public static void flush() {
-    currentTransaction().flush();
+    getDefault().flush();
```
rPraml, line 253: Delegating to DefaultServer.flush, which may throw a "no currentTransaction" exception.

ebean-api/src/main/java/io/ebean/DB.java
```diff
@@ -313,7 +312,7 @@ public static void endTransaction() {
    * Mark the current transaction as rollback only.
    */
   public static void setRollbackOnly() {
-    getDefault().currentTransaction().setRollbackOnly();
+    Objects.requireNonNull(currentTransaction(),"no currentTransaction").setRollbackOnly();
```
rPraml, line 315: Note: There is no "setRollbackOnly" in DefaultServer. Intention?

ebean-api/src/main/java/io/ebean/Database.java
```diff
@@ -35,7 +35,7 @@
  * <h5>The 'default' Database</h5>
  * <p>
  * One Database can be designated as the 'default' or 'primary' Database
- * (see {@link DatabaseConfig#setDefaultServer(boolean)}. Many methods on DB
+ * (see {@link DatabaseConfig#setDefaultServer(boolean)}). Many methods on DB
```
rPraml, line 38: Fixed missing `)`

ebean-api/src/main/java/io/ebean/Database.java
```diff
@@ -166,13 +167,15 @@ public interface Database {
   /**
    * Return the BeanState for a given entity bean.
    * <p>
-   * This will return null if the bean is not an enhanced entity bean.
+   * This will throw an IllegalArgumentException if the bean is not an enhanced entity bean.
```
rPraml, line 170: checked impl. will never return null

ebean-api/src/main/java/io/ebean/ExpressionList.java
```diff
@@ -404,6 +404,7 @@
    *
    * }</pre>
    */
+  @Nullable
   default <A> A findSingleAttribute() {
```
rPraml, line 408: This was the root cause, why I created this PR

ebean-core/src/main/java/io/ebeaninternal/api/SpiEbeanServer.java
```diff
@@ -14,7 +15,6 @@
 import io.ebeaninternal.server.query.CQuery;
 import io.ebeaninternal.server.transaction.RemoteTransactionEvent;
 
-import javax.annotation.Nullable;
```
rPraml, line 17: Note: Wrong annotation was used here.

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -586,11 +586,7 @@ public void clearQueryStatistics() {
    */
   @Override
   public <T> T createEntityBean(Class<T> type) {
-    final BeanDescriptor<T> desc = descriptor(type);
-    if (desc == null) {
-      throw new IllegalArgumentException("No bean type " + type.getName() + " registered");
-    }
-    return desc.createBean();
+    return descriptor(type).createBean();
```
rPraml, line 589: Code change: I moved this check to `descriptor(class)` as it claims to never return null due `@NonNullApi`

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -630,7 +626,11 @@ private <T> T referenceFindOne(Class<T> type, Object id, BeanDescriptor<?> desc)
     }
     // we actually need to do a query because we don't know the type without the discriminator
     // value, just select the id property and discriminator column (auto added)
-    return find(type).select(idProp.name()).setId(id).findOne();
+    T bean = find(type).select(idProp.name()).setId(id).findOne();
+    if (bean == null) {
+      throw new NullPointerException("Could not find reference bean " + id + " for " + desc);
+    }
```
rPraml, line 632: Code change: I throw an NPE now.

ebean-api/src/main/java/io/ebean/BeanFinder.java
```diff
@@ -63,6 +64,7 @@ public Database db() {
   /**
    * Return the current transaction.
```
jnehlmeier, line 65: ... or null if there is no current transaction in scope.

ebean-api/src/main/java/io/ebean/Database.java
```diff
@@ -866,7 +871,7 @@ public interface Database {
    * the bean does not exist then this returns false indicating that nothing was
    * deleted. However, if JDBC batch mode is used then this always returns true.
    */
-  boolean delete(Object bean, Transaction transaction) throws OptimisticLockException;
+  boolean delete(Object bean, @Nullable Transaction transaction) throws OptimisticLockException;
```
jnehlmeier, line 874: Really? (haven't checked it). But if JavaDoc states `... with an explicit transaction.` on an API method it feels weird to me that this explicit transaction can actually be `null`. Could very well hide some coding problems in user code.

If that is really true, then JavaDoc should describe that you can pass in `null` as a transaction and what this means if you do so. Same for all the other methods taking an explicit transaction as input parameter. In that case JavaDoc of the equivalent method without transaction parameter could say something like `this method behaves exactly as passing a null transaction to delete(bean, transaction)` and link to the other method. Would avoid duplicating JavaDoc on all these methods.

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
jnehlmeier: NPE is more like for preconditions or if you really messed up your code, don't you think? How does this exception propagate? I would not expect to received a NPE in user code if an entity isn't found. Isn't there an EntityNotFoundException?

jnehlmeier:
Generally I feel like Ebean JavaDoc could make more use of `@return`, `@throws` and `@param` so you don't have to read through possibly lengthly JavaDoc text to get the details.

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
rPraml: Hmm. Unfortunaltely, this is not completely true.
While DefaultServer is annotated with `@NonNullApi`, `SpiEbeanServer` does not. And we have also some code, in our app, where we check if a class is registered. So it  might be better to annotate descriptor with @Nullable

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
rPraml: yes, thanks. EntityNotFoundException is the better choice here

ebean-api/src/main/java/io/ebean/Database.java
rPraml: > Really? (haven't checked it). But if JavaDoc states ... with an explicit transaction.

Yes, the implementation checks it, and the methods without transaction parameter just pass null there.

> If that is really true, then JavaDoc should describe that you can pass in null as a transaction and what this means if you do so. 

ACK

rPraml:
> Generally I feel like Ebean JavaDoc could make more use of @return, @throws and @param so you don't have to read through possibly lengthly JavaDoc text to get the details.

I partially agree, What I don't like, if you just repeat the documentation like in this example ¹
```java
/**
 * Returns the current transaction
 * @return the current transaction
 */ 
 Transaction currentTransaction();
```
But you are right, for longer javadocs, a small summary would not too bad.

(I don't want to (have to) revise the complete documentation in this PR, I just want to fix the NPE warnings in the IDE ;) )

¹) Personally, I read the documentation in 90% in the IDE by jumping to the source. (Or in doubt, reading the source itself, because code never lies, while documentation can be incorrect)

rob-bygrave:
Right so I will not merge this PR because it contains some changes that I don't like / see as being a negative.

Some changes look good but once they are included with other changes that went into behaviour changes and other changes that I believe are net negative.

ebean-api/src/main/java/io/ebean/Database.java
```diff
@@ -120,6 +120,7 @@ public interface Database {
   /**
    * Return the associated read only DataSource for this Database instance (can be null).
    */
+  @Nullable
```
rob-bygrave, line 123: I suspect net negative

ebean-api/src/main/java/io/ebean/Database.java
rob-bygrave: Good thanks.

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
rob-bygrave: Behaviour change means it should be in its own PR.  We don't want subtle behaviour changes in a PR that says "Improved Null-checking annotations"  and we do want behaviour changes in their own PR that justifies why the change is good.

rPraml:
Hello @rob-bygrave, @jnehlmeier,

thank you for your constructive feedback.

Rob already merged some changes with #3066

I'll try to summarize, what's left in this PR
- I will revert all code changes here and create extra PRs
- nextId: Yes - it makes no sense here for @Nullable
- currentTransaction: I'm unsure myself here. 
I understand Rob's argument that this could be a negative overall.
However, at least we have pieces of code where we make sure that this is not called with an active transaction.
For example, this is mostly code, that interacts with external APIs, which can't be roll backed.
```java
void deleteUser(String userId) {
   assert DB.currentTransaction() == null : "must not run in transaction";
   if (externalApi.deleteUser(userId)) {
       user = DB.find(User.class, userId);
       user.setDeleted(true);
       DB.save(user);
   }
}
```
It would be bad, if this code runs in a transaction, that is rolled back afterwards.
I can live with a potential IDE warning 'currentTransaction() is never null here' 
- what to do with methods like `boolean delete(Object bean, @Nullable Transaction transaction)`?

I will not continue with the pull request for the time being, but will make it available for discussion first, since it only solves cosmetic problems in the IDE for us.

Roland

rbygrave:
> `assert DB.currentTransaction() == null : "must not run in transaction";`

Ok, I see.  To me I'd say this use case is like `@Transaction(type=NEVER)` but in programatic form.

My gut says that the alternative is for `currentTransaction()` to always return a non nullable transaction and that a new  `boolean hasCurrentTransaction()` be added.  This is probably based on the thought that the most common use of `currentTransaction()` [that I was expecting] is it being used inside methods annotated with `@Transactional` and for these cases the code is always expects non-nullable transaction. 

Hmm, pondering.


> boolean delete(Object bean, @Nullable Transaction transaction)

For these cases I don't see any downside apart from verbosity. I don't think it negatively impacts existing code that is using those methods.

rbygrave:
I think we close this one and have a new fresh PR if desired.

PR #3082, opened 2023-05-31
Add testcase that tickles bug in history

tobias-:
This PR demonstrates a bug in history where if you do `findVersions()` on a history object that references a non-versioned and then a versioned object again, a `systime <@ null` will be added to the query (postgres) which will of course always fail.

This proves the bug reported as #3083


rbygrave:
Replaced by https://github.com/ebean-orm/ebean/commit/443b68a3b0b0607a50f15c44101c8feda81f5527 fixed in version 18.1.0

PR #3113, opened 2023-06-16
Bugfix for wrong batch statement ordering with many2many relations

jonasPoehler:
Hello Rob,

this PR fixes an issue, when a List representing a ManyToMany-relation is cleared and then filled with (some of) the original elements. Functionally this resulted in the delete and the following insert statements to be executed within the same batch queue, because with m2m intersection tables the same queue was used for both types of SqlUpdate. The end result was that in these cases the first insert statements were executed before all delete statements were flushed, resulting in a DuplicateKeyException. Additionally, I introduced constants for the delete and insert queue positions for better readability.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
@@ -83,6 +83,9 @@ public final class BatchControl {
 
   private final Queue[] queues = new Queue[3];
 
+  static final int DELETE_QUEUE = 1;
+  static final int INSERT_QUEUE = 2;
+
   /**
    * Create for a given transaction, PersistExecute, default size and getGeneratedKeys.
    */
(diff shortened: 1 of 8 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
```diff
@@ -267,8 +270,8 @@ public void clear() {
   private void flushBuffer(boolean reset) throws BatchedSqlException {
     flushQueue(queues[0]);
```
jonasPoehler, line 271: We were unable to find a use of the queue at position 0. Maybe this one is unused?

rbygrave: Ah ok ... hmm, I need to look at that.

ebean-core/src/main/java/io/ebeaninternal/server/persist/SaveManyBeans.java
```diff
@@ -294,7 +294,7 @@ private void saveAssocManyIntersection(boolean queue) {
         // build a intersection row for 'delete'
         IntersectionRow intRow = many.buildManyToManyMapBean(parentBean, otherDelete, publish);
         SpiSqlUpdate sqlDelete = intRow.createDelete(server, DeleteMode.HARD);
-        persister.executeOrQueue(sqlDelete, transaction, queue);
+        persister.executeOrQueue(sqlDelete, transaction, queue, BatchControl.DELETE_QUEUE);
```
rbygrave, line 297: This is the bug fix here.

rbygrave (approved):
Nice!!

PR #3115, opened 2023-06-19
Validate the lengths of Strings/Json, before they are written to database

rPraml:
So with #3107 it is now possible to use length limits across platforms.
Unfortunately, platforms that do not use varchars or fall back to a lob (like SqlServer) do not enforce that limit.

**Problem**
The JSON in an annotated field with `@DbJson(length = 5000)` is not size limited.
Due programming errors, it is possible to write gigabytes to the DB.

Due recent Jackson updates, these strings could not be read anymore. See https://github.com/ebean-orm/ebean/commit/ca3c02bc1cce66e1176b9b7bcf414ac074a25e6e

**Possible fix**
When a BeanProperty writes its value to the dataBind, it will pass the previously written value to a BindValidator.
The BindValidator can decide, if the value is OK or not.
It can validate string length, file sizes, array sizes and so on.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/DataBindException.java
+++ b/ebean-api/src/main/java/io/ebean/DataBindException.java
@@ -1,0 +1,24 @@
+package io.ebean;
+
+import javax.persistence.PersistenceException;
+
+/**
+ * Thrown when a bind validator detects a data bind error.
+ */
+public class DataBindException extends PersistenceException {
+  private static final long serialVersionUID = -1755215106960660645L;
+
+  /**
+   * Create with a message.
+   */
+  public DataBindException(String message) {
+    super(message);
+  }
+
+  /**
+   * Create with a message and cause.
+   */
+  public DataBindException(String message, Throwable cause) {
+    super(message, cause);
+  }
+}
--- a/ebean-api/src/main/java/io/ebean/config/PlatformConfig.java
+++ b/ebean-api/src/main/java/io/ebean/config/PlatformConfig.java
@@ -1,6 +1,7 @@
 package io.ebean.config;
 
 import io.ebean.annotation.Platform;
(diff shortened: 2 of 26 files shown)


rPraml:
**Explanation**

DataBind, respectively DataBinder have a new method `popLastObject'
This returns the last object, that was added with one of the setXXX methods (setObject, setString, setInt ....)

Performance costs due this stage: 
- Boxing for primitive types (setInt, setLong, setShort...)
- Special Boxing for setBinaryStream
- all other setters just copy the value in an additional field


BeanProperty.bind checks, if it has a bindValidator. If this is the case, it will pop the last object and pass it to the bindValidator

If there is no bindValidator, there are no additional performance costs.

The bindValidator itself is created by a BindValidatorFactory, defined by the DatabasePlatform and can be overridden by user in the PlatformConfig.

The default implementation just checks string lengths. But it is also possible to implement own validators
(e.g. check if precision loss may happen on Instants)

Note there is a difference, when UTF8 chars are used on certain platforms like DB2. So there is also an UTF8-BindValidatorFactory.

@rob-bygrave Since we've been talking past each other lately, is the description sufficient to understand what I'm trying to achieve? ;)

rbygrave:
The title says _"Validate the lengths of Strings/Json"_ but the PR does a _LOT_ more than that and impacts primitives etc - it goes into places it doesn't need to. So I question the design options. To me it is not clear that this is a good approach.

rPraml:
> The title says "Validate the lengths of Strings/Json" but the PR does a LOT more than that and impacts primitives etc - it goes into places it doesn't need to. So I question the design options. To me it is not clear that this is a good approach.

So you mean, that I should focus on the string problem. My approach was to introduce an API that can be extended for other use cases, but I agree, that this PR may be over-engineered ;)

rbygrave:
Close - replaced by #3121

PR #3119, opened 2023-06-20
JSON/String/Stream length validation (Take 2)

rPraml:
Hello Rob,

thanks for the feedback in #3115. First I want to explain what I want:

**What do I need**

When I save a JSON/Varchar/File that exceedes the defined db-length, I want an exception on every platform
While H2 enforces these constraints on DB level, other platforms do not and allow to store gigabytes in such fields

**Why do I need this** 

I need this to detect programming errors, misconfiguration and misuse early. 
1) A Json field, where I do not expect more than 4K should not be filled with some GB
2) Same for a field, where the user can write down some notes (expected < 32k) 
3) A file field, where I expect a file less than 1MB should not be misused by the user with a 700MB ISO image ;)

**Can this implemented in the application**

Partially. While 2 + 3 are implemented with `@Size` + BeanValidation in our application, I fail to implement this in our application, as I do not have access to the JSON, that ebean will write. Of course, I can serialize such beans twice, but I still see the problem to get BeanValidation working.

I tried different approaches the last days and also spent hours of implementing something that will not work ;-)

- Tried to change the return value from `ScalarType.bind(DataBinder binder, T value)` from void to Object, so that bind returns, what it has effectively bound. Problem: API change -> Breaks external modules
- Wondering, if I should implement the validation in ScalarType. Problem: Currently there is a singleton instance of ScalarTypeString. I did not want to change this
- So the most promising approach seems to "capture" the last value in the DataBinder. I am still wondering, if I should focus only on strings, arrays and files here.

The next problem is, when and how to enable this feature. While H2 performs all checks itself, I explicitly had to switch this on on other platforms. Oracle and DB2 are special cases, as they define the varchar limit on byte level (I assume, other platforms will do that also, depending on the underlying charset)

So maybe you can give me a hint in what direction I should go.
(I would also be satisfied if I could determine on the application side if a JSON object is getting too large.)

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/bind/DataBind.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/bind/DataBind.java
@@ -1,10 +1,14 @@
 package io.ebeaninternal.server.bind;
 
 import io.ebean.core.type.DataBinder;
+import io.ebean.core.type.InputStreamInfo;
 import io.ebeaninternal.api.CoreLog;
 import io.ebeaninternal.server.core.timezone.DataTimeZone;
 
-import java.io.*;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringReader;
 import java.math.BigDecimal;
 import java.sql.*;
 import java.util.ArrayList;
@@ -15,6 +19,7 @@ import static java.lang.System.Logger.Level.WARNING;
 
 public class DataBind implements DataBinder {
 
+  private static final Object UNBOUND = new Object();
   private final DataTimeZone dataTimeZone;
   private final PreparedStatement pstmt;
   private final Connection connection;
@@ -22,6 +27,8 @@ public class DataBind implements DataBinder {
   private List<InputStream> inputStreams;
   protected int pos;
   private String json;
+
+  private Object lastObject = UNBOUND;
 
   public DataBind(DataTimeZone dataTimeZone, PreparedStatement pstmt, Connection connection) {
     this.dataTimeZone = dataTimeZone;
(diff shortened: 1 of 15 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/bind/DataBind.java
```diff
@@ -117,36 +126,43 @@ public final PreparedStatement getPstmt() {
   @Override
   public void setString(String value) throws SQLException {
     pstmt.setString(++pos, value);
+    lastObject = value;
   }
 
   @Override
   public final void setInt(int value) throws SQLException {
     pstmt.setInt(++pos, value);
+    lastObject = value;
```
rob-bygrave, line 135: I think this should instead be `lastObject = null` - avoid the boxing + this should never be validated

I think we should do this on all the primitive types + dates + timestamps + BigDecimal - all the types we are not interested in validating.

ebean-core/src/main/java/io/ebeaninternal/server/bind/DataBind.java
```diff
@@ -211,21 +234,31 @@ public void setBinaryStream(InputStream inputStream, long length) throws SQLExce
...
   @Override
   public void setClob(String content) throws SQLException {
     pstmt.setCharacterStream(++pos, new StringReader(content), content.length());
+    lastObject = content;
   }
 
   @Override
   public void setArray(String arrayType, Object[] elements) throws SQLException {
     pstmt.setArray(++pos, connection.createArrayOf(arrayType, elements));
+    lastObject = elements;
```
rob-bygrave, line 255: This can also be `lastObject = null` as it is not going to be validated

rob-bygrave:
Ok, leave this with me for a little bit. I'll look to make a couple of minor adjustments.

rbygrave:
Close - replaced by #3121

PR #3121, opened 2023-06-25
Foconis string length validation

rbygrave:


ebean-api/src/main/java/io/ebean/config/LengthCheck.java
```diff
@@ -0,0 +1,22 @@
...
+   */
+  OFF,
+  /**
+   * When enabling length check, ebean validates strings/json strings and files before saving them to DB.
+   */
+  ON,
+  /**
+   * Same as "ON", but take the UTF8-bytelength for validation. This may be useful, if you have an UTF8 based charset (default for DB2)
+   */
+  UTF8
```
rPraml, line 21: #3115 would be more flexible here, if you have some odd charsets.

ebean-test/testconfig/ebean-db2.properties
```diff
@@ -5,3 +5,4 @@ ebean.test.username=admin
 ebean.test.password=admin
 datasource.default=db2-11
 ebean.db2-11.databasePlatformName=db2luw
+ebean.lengthCheck=utf8
```
rPraml, line 8: This is something, that maybe shoud be moved to databasePlatform.
On the other side, we cannot know the exact charset, the DB2 database is using.

rPraml (commented):
Thanks for looking on that PR, 

I wrote some comments about the differences and intentions between #3115 and #3119
#3115 is a more generic approach (custom validators support) while #3119 is restricted exactly to our problem (=String and file length validation for default and UTF8 charset)

I also see, that I've forgotten this change https://github.com/ebean-orm/ebean/pull/3115/files#diff-3647cd9abedddc222c44f401182e356fe5ce17e71800b9df38805c612c2a207d in #3119

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -554,6 +559,22 @@ public Object readSet(DbReadContext ctx, EntityBean bean) throws SQLException {
   @SuppressWarnings("unchecked")
   public void bind(DataBind b, Object value) throws SQLException {
     scalarType.bind(b, value);
+
+    if (needsLengthCheck()) {
+      LengthCheck lengthCheck = descriptor().config().getLengthCheck();
```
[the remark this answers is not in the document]
rbygrave: Out of date comment. This is changed.  I'll just mark this as resolved.

ebean-core/src/main/java/io/ebeaninternal/server/bind/DataBind.java
```diff
@@ -117,36 +126,43 @@ public final PreparedStatement getPstmt() {
   @Override
   public void setString(String value) throws SQLException {
     pstmt.setString(++pos, value);
+    lastObject = value;
   }
 
   @Override
   public final void setInt(int value) throws SQLException {
     pstmt.setInt(++pos, value);
+    lastObject = value;
```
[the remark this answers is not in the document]
rbygrave: This comment looks out of date. I've only just pushed up my actual changes to the original PR.

The `lastObject = null;` is used for all the types not expected to be validated for max length.

> Timestamps ... BigInteger ...

This is not expected to validate for those types.

ebean-api/src/main/java/io/ebean/config/LengthCheck.java
rPraml: Ack: let's leave utf8 for now, If other charsets are needed, we can implement it (maybe CUSTOM)

ebean-core-type/src/main/java/io/ebean/core/type/DataBinder.java
```diff
@@ -174,4 +174,9 @@ public interface DataBinder {
    */
   String popJson();
 
+  /**
+   * Returns the last bound object (e.g. for BindValidation). Note for InputStreams you'll get an InputStreamInfo.
```
rPraml, line 178: Add docs (and/or method name, e.g. 'popLastLengthObject' ), that only Strings & InputStreams are returned

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BindMaxLength.java
```diff
@@ -0,0 +1,68 @@
...
+
+  final class UTF8 implements BindMaxLength {
+
+    @Override
+    public long length(int dbLength, Object obj) {
+      if (obj instanceof String) {
+        String s = (String) obj;
+        int stringLength = s.length();
+        if (stringLength > dbLength) {
+          return stringLength;
```
rPraml, line 36: 👍 Yes this is an other fast path.

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -554,6 +562,18 @@ public Object readSet(DbReadContext ctx, EntityBean bean) throws SQLException {
   @SuppressWarnings("unchecked")
   public void bind(DataBind b, Object value) throws SQLException {
     scalarType.bind(b, value);
+    if (bindMaxLength != null) {
+      Object obj = b.popLastObject();
+      long length = bindMaxLength.length(dbLength, obj);
+      if (length > dbLength) {
+        b.closeInputStreams();
+        String s = String.valueOf(value); // take original bind value here.
```
rPraml, line 570: I'm still unsure, if original bind value should be taken.
- For strings, it does not matter 
- For JSON objects, you'll get the toString from the JSON object instead of the JSON itself (for Map e,g `[foo=a, bar=b]` instead of `{"foo":"a", "bar":"b"}`)
- The most important thing are files here. I'm more interested in the filename instead of some object id from a stream

rPraml (commented):
Thanks, this implementation fits our use case. 👍 

One more thing you could take a look at are SqlUpdates. See

> I also see, that I've forgotten this change https://github.com/ebean-orm/ebean/pull/3115/files#diff-3647cd9abedddc222c44f401182e356fe5ce17e71800b9df38805c612c2a207d in https://github.com/ebean-orm/ebean/pull/3119

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
rbygrave: How about we go with this until someone comes up with a better plan

ebean-test/testconfig/ebean-db2.properties
rbygrave: I think let have it explicit per database first and see.

ebean-test/testconfig/ebean-sqlserver19.properties
```diff
@@ -7,6 +7,7 @@ ebean.test.sqlserver.port=9434
 ebean.test.sqlserver.url=jdbc:sqlserver://localhost:9434;databaseName=test_ebean;sendTimeAsDateTime=false;integratedSecurity=false;trustServerCertificate=true
 datasource.default=sqlserver2019
 ebean.sqlserver2019.databasePlatformName=sqlserver17
+ebean.lengthCheck=on
```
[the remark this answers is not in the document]
rbygrave: Lets leave it off by default initially

rbygrave:
We can close this now.  Will be in 14.0.1

PR #3124, opened 2023-07-02
Add InTuples expression for multi-column IN expression

rbygrave:
Example

```java

    var c = QContact.alias();

    InTuples inTuples = InTuples.of(c.firstName, c.lastName)
      .add("Rob", "B")
      .add("Bob", "C")
      .add("Mob", "D");

    var query = new QContact()
      .firstName.isNotNull()
      .email.isNotNull()
      .inTuples(inTuples)
      .query();

    query.findList();
    assertThat(query.getGeneratedSql()).contains("(t0.first_name,t0.last_name) in ((?,?),(?,?),(?,?))");
```

**Note** this is currently not supported by SQLServer or DB2.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/DInTuples.java
+++ b/ebean-api/src/main/java/io/ebean/DInTuples.java
@@ -1,0 +1,62 @@
+package io.ebean;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ *
+ */
+final class DInTuples implements InTuples {
+
+  private final String[] properties;
+  private final List<Entry> entries = new ArrayList<>();
+
+
+  DInTuples(String[] properties) {
+    this.properties = properties;
+  }
+
+  /**
+   */
+  @Override
+  public InTuples add(Object... values) {
+    entries.add(new DEntry(values));
+    return this;
+  }
+
+  /**
+   * Return the first property name.
+   */
+  @Override
+  public String[] properties() {
+    return properties;
+  }
+
+  /**
+   * Return all the value pairs.
+   */
+  @Override
+  public List<Entry> entries() {
+    return Collections.unmodifiableList(entries);
+  }
+
+  /**
+   * A pair of 2 value objects.
+   * <p>
+   * Used to support inPairs() expression.
+   */
+  static final class DEntry implements InTuples.Entry {
+
+    private final Object[] vals;
+
+    DEntry(Object[] vals) {
+      this.vals = vals;
+    }
+
+    @Override
+    public Object[] values() {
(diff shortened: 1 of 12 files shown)


rPraml:
Hi @rbygrave 

I like that feature, unfortunately it will not work on all databases.

On DB2 you have to use `where (prop1, prop2) in ( VALUES (?,?),(?,?) ...)`  (Same syntax works on MariaDB)

On SqlServer, you have to do other tricks, like subqueries or just a simple conjunction of ANDs / ORs.
See: https://stackoverflow.com/questions/8006901/using-tuples-in-sql-in-clause

BTW: What do you think about this syntax?

```java
 var query = new QContact()
      .firstName.isNotNull()
      .email.isNotNull()
      .inTuples(c.firstName, c.lastName)
         .add("Rob", "B")
         .add("Bob", "C")
         .add("Mob", "D")
      .endTuple()
      .query();
```

Roland.

rbygrave:
> DB2

Oh, I didn't know DB2 supported it via that syntax with `VALUES` - interesting and maybe we can support that.  I was ok with putting this into the next version knowing it does not work with SQLServer and DB2.

> SQLServer

Yes. It's almost easier to use concat functions and use a plain IN which is what we have with `inPairs()` (noting that only supports 2 columns).


> BTW: What do you think about this syntax?

Well I expect the normal case is that we have a collection of things ... and iterate the collection calling `add(...)`. That is, I don't think we will do this in-line in the query much at all.

Hmmm, let me think about it.  `endTuple()` is consistent with `endOr()`.

rPraml:
> > SQLServer
> 
> Yes. It's almost easier to use concat functions and use a plain IN which is what we have with `inPairs()` (noting that only supports 2 columns).

Do you plan to make a fallback implementation for platforms that do not support tuples? I really like that, as we are often using this here (we use and/or queries), so it would be great, if we can get some benefits, if the platfrom supports it

Some thoughts:
- concat: easy to implement, will make problems on special data types where java.toString differs from server.toString (e.g. dates + times)
- combination of AND/OR: Should be also easy to implement, but makes complex queries
- some complex subquery: would make query less complex (=shorter) but difficult to implement

rbygrave:
> Do you plan to make a fallback implementation for platforms that do not support tuples?

Yes, that is the plan. I'll look to release as it is in 13.20.1 and add an implementation for DB2 & SQLServer later. [Rant, why is it that 2 of the largest tech companies on the planet can't implement this like everyone else ??? ]

> combination of AND/OR:

This is what I thought InTuples will have to do for SQLServer. In theory people can do the concat + IN today but perhaps need guidance / example to show them.

PR #3132, opened 2023-07-05
Adds configurable destination package for querybeans

dsa-ferreira:
added compiler arguments to allow custom destination package suffix for querybeans. motivation behind this is to allow Entities to be package protected. With the .query + . assoc hardcoded this can't compile because the querybeans cannot find the enitities. Kept the default by omission as it was (.query + .assoc). Only did base querybean-generator, no changes were made to kotlin generator.


rob-bygrave:
Noting that I don't believe this does fully works without some change to ebean-agent enhancement and how query beans are detected.

rbygrave:
Still waiting for comments on this.  I'll close this for now.

PR #3144, opened 2023-08-08
FEATURE: NEW: QueryPlanLogger for DB2

rPraml:
This adds a queryPlanLogger for DB2. The reported query-plan is based on this script: http://use-the-index-luke.com/s/last_explained

To generate query-plans, DB2 needs do have explain tables generated with `CALL SYSPROC.SYSINSTALLOBJECTS( 'EXPLAIN', 'C' , '', ${SCHEMA})`

By default, the selected schema is `CURRENT USER`. Depending on your DB2 setup, you can control this generation by an additional `queryPlan.options` config-parameter.
Examples:
- `ebean.db.queryPlan.options=create=false` do not create the explain tables
- `ebean.db.queryPlan.options=create=true;schema=SYSTOOLS` create the explain tables in the SYSTOOLS schema

### DIFF
--- a/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
+++ b/ebean-api/src/main/java/io/ebean/config/DatabaseConfig.java
@@ -501,6 +501,11 @@ public class DatabaseConfig {
   private boolean queryPlanEnable;
 
   /**
+   * Additional platform specific options for query-plan generation.
+   */
+  private String queryPlanOptions;
+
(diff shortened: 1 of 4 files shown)


rbygrave:
I think we need to have a service loading mechanism to load QueryPlanLogger / QueryPlanLoggerFactory ... such that this code can move into the ebean-db2 module instead of being in ebean-core.

PR #3146, opened 2023-08-08
Change to loose json parsing for boolean and number types to allow string values

rPraml:
We use `parser.getValueAsXXX` as it does some auto-detection, if the data type does not match exactly.
This allows to read jsons from other sources, where numeric values are encoded as string òr boolean values are encoded as `1` or `0` or `"true"`,`"false"`

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeBoolean.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/type/ScalarTypeBoolean.java
@@ -2,12 +2,7 @@ package io.ebeaninternal.server.type;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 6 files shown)


rob-bygrave:
This isn't strictly a BugFix, it is more a change to allow more loose json parsing for numbers and boolean to allow string values.

Although it might look benign enough this puts a limitation on future json parsing to similarly provide loose parsing for those types. For example, this might be something that is not supported by some future Vector API based json parser.

As such I'll change the title on this issue.

Additionally note that this is something that I feel is problematic in terms of future json parsing options so seems unlikely to be merged / adopted.

rPraml:
The problem was, that we have changed a datatype from "string" to "integer" (there were only numbers stored in the varchar column, unfortunately, exported jsons become unreadable after this change)

I'll talk with my team, if we can find an other solution @SimonBuettner FYI

rbygrave:
Can we close this PR?

PR #3152, opened 2023-08-09
[WIP] NEW: DefaultDbMigration is configurable from properties

rPraml:
This PR allows to configure the DbMigration completely from properties-files and it is a plugin now.
This means, ebean can automatically generate migration files. There is no programmatic code needed any more

TODO: Write some good documentation

### DIFF
--- a/ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/DbMigrationPlugin.java
+++ b/ebean-ddl-generator/src/main/java/io/ebeaninternal/dbmigration/DbMigrationPlugin.java
@@ -1,0 +1,56 @@
+package io.ebeaninternal.dbmigration;
+
+import java.io.IOException;
+
+import io.ebean.plugin.Plugin;
+import io.ebean.plugin.SpiServer;
+
+/**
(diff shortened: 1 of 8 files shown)


rob-bygrave:
Why?

rPraml:
I think the motivation to configure evertything from properties is clear

Q: Why a plugin:
A: You need no code. You just need to add the dbmigration to the classpath, set up the correct properties and ebean will generate the migrations automatically at startup. (Similar to ddl-generation)

**What I want**
Generationg dbmigration should work the same way as ddl-generation for tests.
so, if it is on classpath (and enabled in ebean.properties) it should automatically generate migration scripts.

**Current situation here**
We are doing this on our build server and this is controlled by special maven profiles. So a certain CI task runs this profile, generates the DDL and also auto commits it on the current branch. (And as additional feature, it runs DDL tests against various platforms to detect syntactical errors)

BTW: Ebean does a very good job in detecting model changes and generating the migration automatically. With the help of the `@DbMigration` annotation, we can cover nearly all of the use cases. It is extemely rare, that we have to hand craft migration scripts.

PR #3153, opened 2023-08-09
TEST: TestServerOffline

rPraml:
While DB recovery during run should be fine, this test checks, if ebean can start up with an offline DB.

This might be important, if you fire up a stack in kubernetes and the DB is not yet ready or the DB has a network outage. What we do NOT want is, that the (web)application fails completely to start (as we cannot display some error page to the user)


rPraml:
TODO: PR depends on other PR

PR #3156, opened 2023-08-11
BUG: Properties of inherited models can be used on different child models

rPraml:
Hello @rbygrave 

we currently are updating to 13.20 and noticed, that https://github.com/ebean-orm/ebean/pull/3057 has triggered a bug.

What we do is, that we fetch a property of an inherited root, but the property itself lives in one child node:
```java
Property prop = DB.getDefault().pluginApi().beanType(Animal.class).property("name");
```
As this property seems to be "animal" compatible, we use `prop.value(animal)` to read the property value of (any) animal.

But if the animal is a dog, and the property comes from cat, you are just reading the same property index from an other bean.
(Note `DB.getDefault().pluginApi().beanType(Animal.class).property("registrationNumber")` has the same property index as `name`)

before #3057 this works by luck in our use case, as the property was unloaded and so, it just returns null.
Now as a lazy load occurs, we were wondering, why we get a value, where we do not expect one

The PR provides a test and a suggested fix

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
@@ -669,6 +669,33 @@ public class BeanProperty implements ElPropertyValue, Property, STreeProperty {
   }
 
   /**
+   * By default, getIntercept and setIntercept will check, if the passed bean is an instance of descriptor type.
+   * <p>
+   * If the property is not part of the type hierarchy (i.e. is this property from this descriptor)
+   * an UnsuppoertedOperation is thrown.
+   * <p>
+   * If inheritance is involved, this method returns false instead of an exception, if the property might exist in
+   * one of the child beans. This is necessary for getIntercept, as it returns <code>null</code> in this case.
+   * @return true if property can be
+   */
+  private boolean checkPropertyAccess(EntityBean bean) {
+    if (bean == null || descriptor.type().isInstance(bean)) { // null = fall through - NPE is catched later.
+      return true;
+    }
+    InheritInfo inheritInfo = descriptor.inheritInfo();
+    if (inheritInfo == null || inheritInfo.isRoot() || !inheritInfo.getRoot().getType().isInstance(bean)) {
+      throw new IllegalArgumentException(propertyIncomatibleMsg(bean));
+    } else {
+      return false;
+    }
+  }
+
+  private String propertyIncomatibleMsg(EntityBean bean) {
+    String beanType = bean == null ? "null" : bean.getClass().getName();
+    return "Property " + name + " on [" + descriptor + "] is incompatible with type[" + beanType + "]";
+  }
+
+  /**
    * Set the value of the property without interception or
    * PropertyChangeSupport.
    */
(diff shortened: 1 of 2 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -759,12 +765,17 @@ public Object value(Object bean) {
     return getValueIntercept((EntityBean) bean);
   }
 
+  private boolean beanHasProperty(EntityBean bean) {
+    return descriptor.inheritInfo() == null // fast exit for non inherited beans
+      || owningType.isInstance(bean);
```
rPraml, line 770: PLEASE CHECK: 
the idea is, not to call `isInstance` when no inheritance is used (90% of use cases)
BUT there is no barrier somewhere that prevents the user to use a property from model A on model B.

So the question is, should we pay an additional price for the instance check here to make ebean more robust.

rbygrave: This seems reasonable.

Q: The only way to use an incorrect property is via `io.ebean.plugin.Property` correct?

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -759,12 +765,18 @@ public Object value(Object bean) {
...
+  }
+
   /**
    * Return the value of the property method.
    */
   public Object getValue(EntityBean bean) {
     try {
-      return getter.get(bean);
+      return beanHasProperty(bean) ? getter.get(bean) : null;
```
rbygrave, line 779: Technically I don't think this is needed on `getValue()` but instead on `getValueIntercept()` only

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
```diff
@@ -652,6 +652,9 @@ public void setTenantValue(EntityBean entityBean, Object tenantId) {
    */
   public void setValue(EntityBean bean, Object value) {
     try {
+      if (!beanHasProperty(bean)) {
```
rbygrave, line 655: Technically I don't think this is needed on `setValue()` but instead on `setValueIntercept()` only

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
rPraml: We sometimes use ebeaninternal.BeanProperty :-) but there is also ExpressionPath.pathGet/Set in the public API

ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
rPraml: Yes, setValueIntercept should be enough in both cases

rPraml:
@rbygrave I've updated this PR. The check is now perfomed in the setIntercept/getIntercept code paths, but is more strict:

**non inherited beans**
- The property must always be fetched from the descriptor of this bean class.
- There are slight functional changes (edge cases), like fetching `Customer.creTime` which is owned by the superclass `BasicDomain` 
  - **before:** it was possible to read/write all instances of `BasicDomain`
  - **now:** it will only work on customer:
  - if you fetch a property with `db.beanType(Customer.class).expressionPath("cretime")`, use it for `Customer` only (and not for `ContactGroup`....)
  - it is (currently) not possible to get a "universal" property with `db.beanType(BasicDomain.class).expressionPath("cretime")` as there is no descriptor.
  - I think this behaviour is acceptable.

**inherited beans**
- If there is a inherit hierarchy, you might get properties from the inheritance-root, which are not part of all child models.
- Reading such a (non existent) property returns `null`
- Writing to a property will fail

rPraml:
@rbygrave I review all my open PRs. This PR is a small one and IMHO ready to review and merge.

Issue #3133, opened 2023-07-05
"No matches for Entity._idClass the concatenated key columns where not found" when saving entity with composite primary key


Incanus3:
## Note

I actually just found the cause right before I wanted to submit, but still think it's worth reporting.

## Expected behavior

save the entity

## Actual behavior

throws the following exception:
```console
No matches for cz.sentica.qwazar.semantic.storage.DbConcept._idClass the concatenated key columns where not found? I expect that the concatenated key was null, and this bean does not have ManyToOne assoc beans matching the primary key columns?
javax.persistence.PersistenceException: No matches for cz.sentica.qwazar.semantic.storage.DbConcept._idClass the concatenated key columns where not found? I expect that the concatenated key was null, and this bean does not have ManyToOne assoc beans matching the primary key columns?
	at app//io.ebeaninternal.server.persist.dmlbind.BindableIdEmbedded.deriveConcatenatedId(BindableIdEmbedded.java:90)
	at app//io.ebeaninternal.server.persist.dml.InsertMeta.deriveConcatenatedId(InsertMeta.java:116)
	at app//io.ebeaninternal.server.persist.dml.InsertHandler.bind(InsertHandler.java:60)
	at app//io.ebeaninternal.server.persist.dml.DmlBeanPersister.execute(DmlBeanPersister.java:63)
	at app//io.ebeaninternal.server.persist.dml.DmlBeanPersister.insert(DmlBeanPersister.java:46)
	at app//io.ebeaninternal.server.core.PersistRequestBean.executeInsert(PersistRequestBean.java:1200)
	at app//io.ebeaninternal.server.core.PersistRequestBean.executeNow(PersistRequestBean.java:726)
	at app//io.ebeaninternal.server.persist.BatchControl.executeNow(BatchControl.java:232)
	at app//io.ebeaninternal.server.persist.BatchedBeanHolder.executeNow(BatchedBeanHolder.java:98)
	at app//io.ebeaninternal.server.persist.BatchControl.executeAll(BatchControl.java:321)
	at app//io.ebeaninternal.server.persist.BatchControl.flushInternal(BatchControl.java:298)
	at app//io.ebeaninternal.server.persist.BatchControl.flushBuffer(BatchControl.java:271)
	at app//io.ebeaninternal.server.persist.BatchControl.flushReset(BatchControl.java:257)
	at app//io.ebeaninternal.server.transaction.JdbcTransaction.batchFlushReset(JdbcTransaction.java:551)
[...]

PR #3158, opened 2023-08-14
Followup for #3133 - "Embedded id expansion on parentIds" for SQLServer and DB2

rbygrave:


rob-bygrave:
> my guts tells me, that we might run into the 2100 Parameter limit if there are too many dependant beans.

You gut is going to be right. This is going to need a follow up to deal with parameter limits.  We also could look at the same issue with InTuples as that is also a candidate for exceeding max parameter limits.

PR #3178, opened 2023-08-21
Draft: Refactored delete-by-id

rPraml:

PR #3179, opened 2023-08-21
Experimental elliminate not in with too many params

rPraml:
Hello Rob, in this PR I continue the work of #3178.

It is a work around for the "not in" query which might have too many parameter in a stateless update.
It may be useful also for the "in tuples workaround"

Please take a look,

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
@@ -1082,12 +1082,13 @@ public final class DefaultPersister implements Persister {
    * </p>
    */
   void deleteManyDetails(SpiTransaction t, BeanDescriptor<?> desc, EntityBean parentBean,
-                         BeanPropertyAssocMany<?> many, List<Object> excludeDetailIds, DeleteMode deleteMode) {
+                         BeanPropertyAssocMany<?> many, Set<Object> excludeDetailIds, DeleteMode deleteMode) {
     if (many.cascadeInfo().isDelete()) {
       // cascade delete the beans in the collection
       BeanDescriptor<?> targetDesc = many.targetDescriptor();
       if (deleteMode.isHard() || targetDesc.isSoftDelete()) {
-        if (targetDesc.isDeleteByStatement()) {
+        if (targetDesc.isDeleteByStatement()
+          && (excludeDetailIds == null || excludeDetailIds.size() <= 1000)) { // TODO wait for #3176
           // Just delete all the children with one statement
           IntersectionRow intRow = many.buildManyDeleteChildren(parentBean, excludeDetailIds);
           SqlUpdate sqlDelete = intRow.createDelete(server, deleteMode);
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
@@ -697,7 +697,7 @@ public final class DefaultPersister implements Persister {
             if (deleteMode.isHard() && targetDesc.isDeleteByStatement()) {
               executeSqlUpdate(sqlDeleteChildren(expOne), transaction);
             } else {
-              List<Object> childIds = findChildIds(expOne, true); // CHECMKE: do we need something for soft/hard delete here
+              List<Object> childIds = findChildIds(expOne, deleteMode.isHard());
               if (childIds != null && !childIds.isEmpty()) {
                 deleteChildrenById(transaction, targetDesc, childIds, deleteMode);
               }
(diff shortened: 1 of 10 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -642,146 +642,222 @@ public int delete(Class<?> beanType, Object id, Transaction transaction, boolean
...
-      BeanPropertyAssocOne<?>[] expOnes = descriptor.propertiesOneExportedDelete();
-      for (BeanPropertyAssocOne<?> expOne : expOnes) {
-        BeanDescriptor<?> targetDesc = expOne.targetDescriptor();
-        // only cascade soft deletes when supported by target
-        if (deleteMode.isHard() || targetDesc.isSoftDelete()) {
-          if (deleteMode.isHard() && targetDesc.isDeleteByStatement()) {
-            SqlUpdate sqlDelete = expOne.deleteByParentId(id, idList);
-            executeSqlUpdate(sqlDelete, t);
-          } else {
-            List<Object> childIds = expOne.findIdsByParentId(id, idList, t);
```
rPraml, line 712: old code: no "deleteHard" is passed for "ones"

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -642,146 +642,222 @@ public int delete(Class<?> beanType, Object id, Transaction transaction, boolean
...
           // only cascade soft deletes when supported by target
           if (deleteMode.isHard() || targetDesc.isSoftDelete()) {
             if (deleteMode.isHard() && targetDesc.isDeleteByStatement()) {
-              // we can just delete children with a single statement
-              SqlUpdate sqlDelete = many.deleteByParentId(id, idList);
-              executeSqlUpdate(sqlDelete, t);
+              executeSqlUpdate(sqlDeleteChildren(expOne), transaction);
             } else {
-              // we need to fetch the Id's to delete (recurse or notify L2 cache)
-              List<Object> childIds = many.findIdsByParentId(id, idList, t, null, deleteMode.isHard());
```
rPraml, line 733: old code: "deleteHard" is passed for "many"

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -642,146 +642,222 @@ public int delete(Class<?> beanType, Object id, Transaction transaction, boolean
...
-              if (!childIds.isEmpty()) {
-                delete(targetDesc, null, childIds, t, deleteMode);
+              List<Object> childIds = findChildIds(expOne, deleteMode.isHard());
```
rPraml, line 700: new code: "delteHard" is passed for ones. This could be a functional change, although I think this is correct. Please check

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -642,146 +642,222 @@ public int delete(Class<?> beanType, Object id, Transaction transaction, boolean
...
+          if (!many.isManyToMany()) {
+            BeanDescriptor<?> targetDesc = many.targetDescriptor();
+            // only cascade soft deletes when supported by target
+            if (deleteMode.isHard() || targetDesc.isSoftDelete()) {
+              if (deleteMode.isHard() && targetDesc.isDeleteByStatement()) {
+                // we can just delete children with a single statement
+                executeSqlUpdate(sqlDeleteChildren(many), transaction);
+              } else {
+                // we need to fetch the Id's to delete (recurse or notify L2 cache)
+                List<Object> childIds = findChildIds(many, deleteMode.isHard());
```
rPraml, line 720: new code: "deleteHard" is passed for manys -> same as before

rbygrave:
Ok, I'm good with this. I'm happy to merge this.

There might be some small tweaks (like I think there is a place with the id list can never be null).  Are you happy for us to merge and then apply the 3176 changes on top of this?

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
```diff
@@ -642,146 +642,222 @@ public int delete(Class<?> beanType, Object id, Transaction transaction, boolean
...
-    return delete(descriptor, null, idList, transaction, deleteMode);
+    return delete(descriptor, idList, transaction, deleteMode);
   }
 
-  /**
-   * Delete by Id or a List of Id's.
-   */
-  private int delete(BeanDescriptor<?> descriptor, Object id, List<Object> idList, Transaction transaction, DeleteMode deleteMode) {
+  private int delete(BeanDescriptor<?> descriptor, List<Object> idList, Transaction transaction, DeleteMode deleteMode) {
     if (idList == null || idList.size() <= maxDeleteBatch) {
```
rbygrave, line 655: pretty sure idList is never null here now, but we want to adjust this part anyway using number of id properties

ebean-core/src/main/java/io/ebeaninternal/server/persist/DefaultPersister.java
rbygrave: I think this is correct too.

PR #3184, opened 2023-08-23
Auto commit transactions after a certain amount of inserts/updates

rPraml:
Hello @rbygrave I would need feedback to this PR:

**Problem**
Our application creates huge transactions when importing certain data.
On DB2 we ran into the "Transaction log is full problem" (Sqlcode: -964, Sqlstate: 57011)

As temporary solution, the DBA tuned the log settings for this DB (increased LOGFILSIZ from 1024 to 80000, reduced LOGPRIMARY from 16 to 5 and increased LOGSECOND from 22 to 50)
They also told us, that they do not want to increase this to such high values, as they cannot guarantee a defined RPO, because the log files may not be archived in time.

**Suggested solution**
The advice of the DBA was to do imports in smaller transactions. We checked the `Transaction.commitAndContine()` function and this seems to do the job.

Unfortunately, this method has to be called in the application code (e.g. always after 1000 save's)
This would need a lot refactoring in our application. Also doing it without a transaction (and use only implicit) is not a good option, as this kills the performance

This PR adds a feature, that "commitAndContinue" is called from time to time.
My idea was, to set a `maxTransactionSize` and after that, the transaction will commit and continue.

**Questions**
- what is a good metric for `size`. The best unit would be size in bytes, but it is extremely difficult to compute this
- so I think, we should count the beans that were saved. Here is also the problem, that saving a simple bean counts the same as saving a complex object graph
- is `commitTransIfRequired` the correct place.
- currently, also non dirty beans will be counted.

**Further thoughts**
I think it is not a good idea, to count inserts/batch binds and do a commitAndContinue after a fixed amount of inserts (this means, if the transaction is then rollbacked, the "half" of the dependant beans could be saved.

Should we estimate the amount of data, that the transaction holds or should we be conservative here and set the max transaction size to <= 1000 beans, so that we are far away from the actual database limit

Rob, maybe you can drop a comment, if it is worth to contine the work on this idea or if there is a different approach to solve the issue.

### DIFF
--- a/ebean-api/src/main/java/io/ebean/Transaction.java
+++ b/ebean-api/src/main/java/io/ebean/Transaction.java
@@ -555,4 +555,13 @@ public interface Transaction extends AutoCloseable {
    * Get an object added with {@link #putUserObject(String, Object)}.
    */
   Object getUserObject(String name);
+
+  /**
+   * Sets the maximum transaction size. After this amount of inserts/updates or sql/orm calls,
+   * the transaction performs an autocommit.
+   * This prevents having huge transactions, which consumes a lot of log space.
+   * Note: saving an object graph could produce more than one insert/update statement,
+   * but this increases the transaction size only by one.
+   */
+  void setMaxTransactionSize(int maxTransactionSize);
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/api/SpiTransaction.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/SpiTransaction.java
@@ -357,4 +357,9 @@ public interface SpiTransaction extends Transaction {
    */
   void postRollback(Throwable cause);
 
+  /**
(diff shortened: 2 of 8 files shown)


rbygrave:
I don't really like this idea of automatically doing a commit and continue.

I really think this sort of use case is best dealt with explicitly. I expect when really large inserts are being processed there is something tracking the progress in either rows processed or bytes processed or something like that. In this sense I do not expect it to be difficult to make use of `commitAndContinue()`.

Once we start using `commitAndContinue()` we are explicitly buying into the idea that there might be some failure after that point and therefore maybe some "re-processing" and perhaps awareness of how far through the process got etc.

So generally I think this sort of thing is more interesting and more intricate and real design thought needs to go into these types of use cases. 

So no, I don't like this idea - I'm not sold on this PR.

rPraml:
Thanks for your feedback. We discard this idea and I will discuss in our team what alternatives there are and how we can fix this in the application code.

PR #3200, opened 2023-08-28
[15x] Remove read auditing feature

rbygrave:


rPraml:
We did use this feature long time ago, but the log output was enormous. (It logs all objects in the dependency graph, which was not end-user friendly)
We implemented a similar feature in the application for certain entities. So for us, it is ok to remove it.

PR #3212, opened 2023-08-30
[15x] Remove "Future queries" findFutureList, findFutureIds, findFutureCount

rbygrave:
Note that findFutureCount still exists as an internal implementation detail of findPagedList but not exposed via Query API for public use.


AntoineDuComptoirDesPharmacies:
Hi @rbygrave,

What will be the replacement for this methods ? There will be some new methods to get entity list and count in parallel ?

Yours faithfully,
LCDP

AntoineDuComptoirDesPharmacies:
I see, at the moment we are using `findFutureCount` to get total count without blocking the main thread.
This is pretty useful to mirror the behavior of `loadCount`, `getTotalCount` from `PagedList` class.

Our workflow look like this : 
1. Run findFutureCount()
2. Execute long process
3. Return a result pairing the count + long process

Yours faithfully,
LCDP

PR #3223, opened 2023-09-18
Regression of #3133 with formula properties

rPraml:
Hello Rob, we had found a regression due #3133

**Use case:**

We had the following dataModel in our application
```java
@Entity
public class DataMain {
  @Id  
  private UUID id; 
  private String title;
  @OneToMany(cascade = CascadeType.ALL, mappedBy = "main")
  private List<Data> metaData;
}

public class Data {
  private UUID mainId;
  private String metaKey;
  private Integer valueIndex;
  private String stringValue;
  @ManyToOne
  private DataMain main;
}
```
Logically, `Data` represents an attribute set on `DataMain`
So `DataMain` can have several attributes called `metaKey` and attributes with the same `metaKey` have a position (`valueIndex`)

Later we had the requirement, that it is guaranteed, that `mainId`, `metaKey` and `valueIndex` is unique.

So we were at the decision:
- to add a separate unique index on these 3 items 
- or extend the primary key.

We decided to extend the PK, as we think it is better for the DB than a separate index.
We did this in the way as it is implemented in the test. (Move mainId, metaKey + valueIndex to an embeddable object) and added some annotations, so that the mapping is handled correctly.

But now, we had a problem with existing code, as the properties for "metaKey" and "valueIndex" are only reachable over the embedded ID.

To be backwards compatible, we decided to add the old properties as `@Formula` that will compute directly on the DB field, so existing code can read them.

This works fine till commit #3155, issue #3133

It seems that MatchedImportedProperty will find the formula properties and then it produces a wrong insert statement:
```sql
insert into data_with_formula (main_id, ${}meta_key, ${}value_index, string_value) values (?,?,?,?)
```

This PR fixes ebean, so that we have no breaking change in our application.
Nevertheless, I would like to ask the question whether the decisions to model the data model this way or whether we did “bad things” here

Roland

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/dmlbind/MatchedImportedFactory.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/dmlbind/MatchedImportedFactory.java
@@ -41,7 +41,7 @@ final class MatchedImportedFactory {
     }
     // match on property name
     for (BeanProperty beanProperty : desc.propertiesBaseScalar()) {
-      if (prop.name().equals(beanProperty.name())) {
+      if (!beanProperty.isFormula() && prop.name().equals(beanProperty.name())) {
         return new MatchedImportedScalar(prop, beanProperty);
       }
     }
--- a/ebean-test/src/test/java/org/tests/model/composite/DataWithFormula.java
+++ b/ebean-test/src/test/java/org/tests/model/composite/DataWithFormula.java
@@ -1,0 +1,65 @@
+package org.tests.model.composite;
+
+
+import io.ebean.annotation.Formula;
+import io.ebean.annotation.Index;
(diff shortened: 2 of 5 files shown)


rbygrave:
> decisions to model the data model this way or whether we did “bad things” here

I think it is reasonable to use the `@Formula` columns this way. I think it could also have been done using an `@IdClass` mapping which I think would be viewed as a more "standard" approach than using the `@Formula`.

rbygrave:
IMO the more expected mapping would be to use `@Column(insertable = false, updatable = false)` rather than `@Formula` like:

```java
  @Column(insertable = false, updatable = false)
  // @Formula(select = "${ta}.meta_key")
  private String metaKey;

  @Column(insertable = false, updatable = false)
  // @Formula(select = "${ta}.value_index")
  private Integer valueIndex;
```

PR #3224, opened 2023-09-19
use glassfish jaxb instead of old com.sun jaxb

rPraml:
Hello Rob, as we are using ebean with spring 3 + jakarta in our application, we noted, that we will bundle the old jaxb-impl and the glassfish jaxb-runtime.

I did not spot out the exact difference, but https://mvnrepository.com/artifact/com.sun.xml.bind/jaxb-impl is called `Old JAXB Runtime module` - so I think it is not recommended to use any more.

I also do not know, if the glassfish-jaxb runs in javax mode. Maybe the dependency in ebean-ddl-generator could be removed completely and the user has to provide one of the two libraries.

### DIFF
--- a/ebean-ddl-generator/pom.xml
+++ b/ebean-ddl-generator/pom.xml
@@ -54,9 +54,9 @@
       <version>4.0.0</version>
     </dependency>
     <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-impl</artifactId>
-      <version>4.0.0</version>
+      <groupId>org.glassfish.jaxb</groupId>
+      <artifactId>jaxb-runtime</artifactId>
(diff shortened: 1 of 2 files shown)


jnehlmeier:
For Java 17 you need Glassfish jaxb-runtime anyways. The old one produces errors with Java 17.

rPraml:
> The old one produces errors with Java 17.

it was not the case here. Maybe, because both were on the classpath.

rbygrave:
Nice, this is a good fix.

> Maybe the dependency in ebean-ddl-generator could be removed completely and the user has to provide one of the two libraries.

My expectation is that ebean-ddl-generator is generally a test scope dependency and I think the right thing to do is to keep the dependency.

PR #3280, opened 2023-11-28
QueryBean generation - use full class names for TQRootBean, TQAssocBean

rbygrave:
Rather than imports, such that these are not possible name clashes with beans of the exact same name.


jnehlmeier:
Don't you think that all EBean API classes should be full qualified in generated code? As soon as there is an import to any Ebean API class, that class name can not be used as entity name.

Looking at the current file there is `Constants.AVAJE_LANG_NULLABLE` and `Constants.DB` also in `importTypes`. Not sure about Avaje Nullable as it is an annotation but DB could still be a problem.

PR #3288, opened 2023-12-08
FIX: nested NOT_SUPPORTED transaction with an inner REQUIRES transaction

rPraml:
Hello Rob, we discovered a strange behaviour, when using NOT_SUPPORTED transaction.

We have some import code in our application Our use case is as follow:

```java
try (Transaction txn1 = DB.beginTransaction()) {
      ImportStatus status = DB.find(ImportStatus.class).forUpdate()...findOne(); // find status entity and place lock

      try (Transaction txn2 = DB.beginTransaction(TxScope.notSupported())) {
        // pause current transaction and run import without an active txn
        runImport();
        txn2.commit(); // not really neccessary, it's the NO_TRANSACTION
        status.setSuccess(true);
      } catch (Throwable t) {
        status.setSuccess(false);
      }
      DB.save(status)
}
```
This works fine, as long the "runImport" will not open new transactions. In this case, the next `DB.beginTransaction()` will not detect the open stack of `txn1` and `txn2` and will effectively overwrite the existing txnContainer.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
@@ -28,8 +28,8 @@ import io.ebeaninternal.server.profile.TimedProfileLocationRegistry;
 import io.ebeanservice.docstore.api.DocStoreTransaction;
 import io.ebeanservice.docstore.api.DocStoreUpdateProcessor;
 import io.ebeanservice.docstore.api.DocStoreUpdates;
-
 import jakarta.persistence.PersistenceException;
+
 import javax.sql.DataSource;
 import java.sql.Connection;
 import java.sql.SQLException;
@@ -174,13 +174,6 @@ public class TransactionManager implements SpiTransactionManager {
   @Override
   public final SpiTransaction active() {
     return scopeManager.active();
-  }
-
-  /**
-   * Return the current active transaction as a scoped transaction.
-   */
-  private ScopedTransaction activeScoped() {
-    return (ScopedTransaction) scopeManager.active();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
```diff
@@ -176,13 +176,6 @@ public final SpiTransaction active() {
     return scopeManager.active();
   }
 
-  /**
-   * Return the current active transaction as a scoped transaction.
-   */
-  private ScopedTransaction activeScoped() {
-    return (ScopedTransaction) scopeManager.active();
-  }
```
rPraml, line 184: not longer in use

ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
```diff
@@ -499,7 +492,7 @@ public final ScopedTransaction externalBeginTransaction(SpiTransaction transacti
    */
   public final ScopedTransaction beginScopedTransaction(TxScope txScope) {
     txScope = initTxScope(txScope);
-    ScopedTransaction txnContainer = activeScoped();
+    ScopedTransaction txnContainer = (ScopedTransaction) inScope();
```
rPraml, line 495: is it possible (or allowed) that we will get an instance not implementing `ScopedTransaction` here?

rPraml: if a `DB.save(entityWithPersistAction)` creates an implicit transaction, we will fail here, if a callback or `@PrePersist` action tries to create a nested transaction, because implicit transactions are not nestable

rPraml:
Hmm. There are still some issues, when ebean tries to open some implicit transactions.
See last commit.

I've tried to start fixing it here: https://github.com/ebean-orm/ebean/commit/38b5ac23283459cc5063e72012e01d4a53b2cde0 but I think this will go into the wrong direction.

So the question is: Means "NOT_SUPPORTED" really: Pause current transaction and behave like never a transaction was opened

Note: As workaround we currently experiment with:
```java
  // find status
  Thread t = new Thread(()-> { /* runs without txn */});
  t.start();
  t.join();
  // update status
} 
```

PR #3289, opened 2023-12-08
Change semantic of userObjects in nested transaction

rPraml:
Hello Rob,
I want to discuss, if this change makes sense.

We often use `Transaction.putUserObject` to store some context information and use it in TransactionCallbacks
With the last PR, we discovered, that this does not work how we thought, when nested transactions are use.

We expected that user objects are stored at the current scope. So after a try/finally block, the information is gone, but it seems, if there is a transaction open, that the objects are stored at the upper most transaction.

Please take a look at this PR, which will change the behaviour, how userObjects are stored.

Roland

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/ScopeTrans.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/ScopeTrans.java
@@ -3,6 +3,8 @@ package io.ebeaninternal.api;
 import io.ebean.TxScope;
 
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


rbygrave:
Yes, makes good sense to me. Given the behaviour change we could include this in 14.x.  

My thought is that 14.x would be:
- this change
- the ebean-migration change to JdbcMigrations
- switch javax and jakarta so that 14.x is jakarta and 14.x-javax is javax

PR #3290, opened 2023-12-10
Update for postgis.jdbc version 2021.1.0

rbygrave:


rbygrave:
This can be removed now that we have ebean-net-postgis-types

Issue #3295, opened 2023-12-16
Updates are not seen inside transaction created using `beginTransaction()` (until commit)


Incanus3:
## Expected behavior

When I run update query (using `Database.update()` or `Query.update()`) and then read the value from db inside the same transaction created using `beginTransaction()` before calling `Transaction.commit()`, I would like to see the update value.

## Actual behavior

I get the old value.

### Steps to reproduce

```kotlin
@Entity
@DbName("ea")
@Table(name = "T_OBJECT")
data class EaObject(
    @Id
    @Column(name = "OBJECT_ID", nullable = false)
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    var id: Long = 0,

    @Column(name = "NAME")
    var name: String? = null,
) {
    override fun equals(other: Any?): Boolean = other is EaObject && other.id == this.id
    override fun hashCode(): Int = id.hashCode()
}

class DummyTestApplication

// this is needed for TestPropertySource to work
@SpringBootTest(classes = [DummyTestApplication::class])
@TestPropertySource(
    properties = [
        "logging.level.io.ebean.SQL=DEBUG",
        "logging.level.io.ebean.TXN=DEBUG",
    ],
)
class EbeanUpdateThenReadInTransactionTest {
    private val dsConfig = DataSourceConfig().also {
        it.url = "jdbc:h2:mem:eadb"
        it.driver = "org.h2.Driver"
        it.username = "user"
        it.password = "pass"

        it.addProperty("NON_KEYWORDS", "VALUE")
        it.addProperty("quoteReturningIdentifiers", false)
    }
    private val dbConfig = DatabaseConfig().also {
        it.name = "ea"
        it.isDdlRun = true
        it.isDdlGenerate = true
        it.isDefaultServer = false
        it.currentUserProvider = CurrentUserProvider { "test" }

        it.setDataSourceConfig(dsConfig)

        it.addClass(EaObject::class.java)
    }
    private val database = DatabaseFactory.create(dbConfig)

    @AfterEach
    fun tearDown() {
        // probably not supported by H2 - throws
        // org.h2.jdbc.JdbcSQLSyntaxErrorException: Cannot truncate "PUBLIC.T_OBJECT"; SQL statement:
        // truncate table T_OBJECT [90106-214]
        // database.truncate(EaObject::class.java)

        QEaObject(database).delete()
    }

    // this works
    @Test
    fun `using createTransaction()`() {
[...]

PR #3301, opened 2024-01-08
#3295 Clear PersistenceContext on execution of bulk updates or deletes

rbygrave:
Using SqlUpdate or an ORM Update query clear the appropriate part of the PersistenceContext. The effect is that ORM queries executed after a bulk update will effectively load a fresh copy of the data from the database and will not reuse an instance from the persistence context if the bean in question had already been loaded.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/OrmQueryRequest.java
@@ -338,7 +338,6 @@ public final class OrmQueryRequest<T> extends BeanRequest implements SpiOrmQuery
 
   private int notifyCache(int rows, boolean update) {
     if (rows > 0) {
-      beanDescriptor.contextClear(transaction.persistenceContext());
       beanDescriptor.cacheUpdateQuery(update, transaction);
     }
     return rows;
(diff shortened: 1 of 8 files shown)


jnehlmeier:
bean2 and bean5 should be the same which could be asserted as well. 

Before merging this we might want to think about JPA again. I am not 100% sure but I think this change could cause trouble for people who have migrated to ebean using your JPA mode. If the per transaction entity cache works differently in ebean as in JPA this could lead to hard to diagnose side effects.

rbygrave:
> which could be asserted as well.

Yes nice. I've added that.

> could cause trouble for people who have migrated to ebean using your JPA mode

So that mode is called `autoPersistUpdates` and effectively set on the transaction. I do believe this change does not really work when `autoPersistUpdates = true` or at least it could be very confusing. For now I've made a change to this PR so that this Persistence Context clearing only occurs when `autoPersistUpdates = false`.

Hmm.

rbygrave:
I'm feeling comfortable that this PR should be merged in with the next major version (so 14).

> JPA mode / autoPersistUpdates = true

I'm feeling some regret about adding that feature.  Also just to say Gavin is keen on getting "Stateless Session" functionality into JPA 4 which is all about explicit insert, update, delete etc.

PR #3302, opened 2024-01-12
Add support for Java 21 SequencedSet and SequencedMap

rbygrave:
This requires ebean-agent 13.26.0

Also requires JDK 21 to compile ebean-api with Multi-Release Jar for the Java 21 versions of BeanSet and BeanMap. That is, merging this in means we need JDK 21 to build.


PromanSEW:
Will this release work with Java 17?

rbygrave:
Yes.

The java 21 support is in MR-JAR (multi-release jar).

So technically ebean is supporting Java 11+

‐------
For folks contributing or building from source, we do need jdk 21 to build
ebean from source now. Users of those artifacts just need Java 11+


Er, not sure if I have made that clear, but the trick employed here is
multi-release jars.

Cheers, Rob.

PR #3308, opened 2024-01-18
fix TQRootBean.exists() when used with .or()

Incanus3:


rbygrave:
Great work. I've just adjusted that test there because we had a clash with `.raw("contact.customer_id = customer.id")` in that `customer.id` is a valid logical path from Contact (the ManyToOne to customer). So changing the alias so that it didn't clash there.

PR #3315, opened 2024-01-29
ADD: DB2 supports select .. for update queries

nPraml:
Hello @rob-bygrave ,

I've implemented the select - for update structure for DB2.

IBM documentation of select - for update: https://www.ibm.com/docs/en/db2/11.5?topic=statement-update-clause
nowait & skip locked are only for update queries: https://www.ibm.com/docs/en/db2/11.5?topic=statements-update#sdx-synid_nowait

Can you please take a look at it?

Kind regards
Noemi

### DIFF
--- a/ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
+++ b/ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
@@ -27,14 +27,14 @@ public class TestQueryForUpdate extends BaseTestCase {
     try (final Transaction transaction = DB.beginTransaction()) {
       query = DB.find(Customer.class)
         .forUpdate()
-        .order().desc("id");
+        .orderBy().desc("id");
 
       query.findList();
     }
 
     if (isSqlServer()) {
       assertThat(sqlOf(query)).contains("with (updlock)");
-    } else if (!isDb2()){
+    } else {
       assertThat(sqlOf(query)).contains("for update");
     }
   }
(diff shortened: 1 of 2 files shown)


ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
```diff
@@ -27,14 +27,14 @@ public void testForUpdate() {
     try (final Transaction transaction = DB.beginTransaction()) {
       query = DB.find(Customer.class)
         .forUpdate()
-        .order().desc("id");
+        .orderBy().desc("id");
```
nPraml, line 30: order() is deprecated, I changed it to orderBy() in this test class

platforms/db2/src/main/java/io/ebean/platform/db2/BaseDB2Platform.java
```diff
@@ -53,4 +54,9 @@ public PlatformIdGenerator createSequenceIdGenerator(BackgroundExecutor be, Data
     return new DB2SequenceIdGenerator(be, ds, seqName, sequenceBatchSize);
   }
 
+  @Override
+  protected String withForUpdate(String sql, Query.LockWait lockWait, Query.LockType lockType) {
+    // NOWAIT and SKIP LOCKED not supported with Db2
+    return sql + " for update";
```
nPraml, line 60: @rob-bygrave 

I have another question about this PR.

The test TestQueryFindPagedList.test_forUpdate throws an exception because a query with join cannot be executed.

"For update" cannot be executed in these constellations: https://www.sqlerror.de/db2_sql_error_-511_sqlstate_42829.html

Would it be a viable way to check the SQL to see if any of the "forbidden" keywords appear in the command and then decide whether for update is possible? (e.g.: join, distinct, order by, group by, etc.) Or does it work if ebean passes on the exception to the DB (like the existing solution)?

Something like this:

```
protected String withForUpdate(String sql, Query.LockWait lockWait, Query.LockType lockType) {
    // NOWAIT and SKIP LOCKED not supported with Db2
    if (checkKeywords(sql)) {
        return sql + " for update";
    } else {
        return super.withForUpdate(sql, lockWait, lockType);
    }
}

private boolean checkKeywords(String sql) { ... }
```

nPraml: Thanks for the feedback, we don't have a usecase that forces to check the keywords. We are also agree with the merged solution.

PR #3345, opened 2024-02-29
NEW: Allow control of jsonwrite-loadedprops

rPraml:
We have the following use case:

We want to convert a model to JSON, exactly, we want to convert the first level of an object graph.

We need the IDs of refrenced objects, but we do not need (and **must not** provide) any additional data. 
(The data we do not want to expose can contain secret informations: e.g. password hashes)

For example, when serializing a `Contact` then we expect all properties from contact, but only the ID from `Customer`:

```json
{
  "id":1,
  "firstName":"Jim1",
  "lastName":"Cricket",
  "isMember":false,
  "customer":{"id":1},
  "cretime":"2024-02-29T13:03:10.378Z",
  "updtime":"2024-02-29T13:03:10.378Z"
}
```
So we hoped a simple `PathProperties.parse("*")`  This works fine, as long the "customer" bean is not loaded. 
But if the bean is loaded, the result looks different, and the json contains all properties in customer.
(the presence in persistence-context and also in the bean-cache seems to affect the json output)

In our case, the data at our rest-api was a bit non deterministic, depending of the cache state of the reference bean.

I've tracked down the problem to [isIncludeProperty](https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/json/WriteJson.java#L504) as the fallback is to check if the property is loaded if no explicit path is configured for a certain graph path.

This PR allows to change this behaviour, so that only the ID is written.

@rbygrave maybe you have a better idea about naming or solving our issue?

cheers
Roland

### DIFF
--- a/ebean-api/src/main/java/io/ebean/text/json/JsonWriteOptions.java
+++ b/ebean-api/src/main/java/io/ebean/text/json/JsonWriteOptions.java
@@ -22,6 +22,8 @@ public class JsonWriteOptions {
 
   protected JsonConfig.Include include;
 
+  protected boolean includeLoadedImplicit = true;
+
   protected Map<String, JsonWriteBeanVisitor<?>> visitorMap;
 
   /**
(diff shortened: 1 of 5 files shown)


rbygrave:
> about naming or solving our issue?

I've got nothing better. Seems all good to me.

My only thought is that ... yes default to true for current behaviour but ... should this default to false at some point in the future? That is my only thought just at the moment.

PR #3361, opened 2024-03-16
ebean-test: Add PersistenceContextAccess.clear() for tests that need …

rbygrave:
…to clear the Persistence Context

To support tests that use a transaction that wraps test setup and running the test logic. The need for this case is to clear the persistence context after the test setup phase.


rbygrave:
Going to close this as unneeded at this stage.

Issue #3370, opened 2024-03-25
LimitOffsetPagedList add getTotalCount cache


spinachomes:
I suggest LimitOffsetPagedList.java cache  getTotalCount method result, so getList method may directly return empty list  to avoid an unnecessary query when getTotalCount had done and It's result is equal to Zero.

rbygrave:
Not sure about this in that it's designed to be able to run the 2 queries in parallel. So I don't think it's worth it.  If you want to prepare a PR for what you were thinking about I'd have a look?

PR #3373, opened 2024-03-26
#3370 LimitOffsetPagedList add getTotalCount cache

spinachomes:


ebean-core/src/main/java/io/ebeaninternal/server/query/LimitOffsetPagedList.java
```diff
@@ -85,22 +92,23 @@ public int getTotalPageCount() {
 
   @Override
   public int getTotalCount() {
+    // already fetched?
+    if (totalRowCount > -1) return totalRowCount;
```
rbygrave, line 96: This needs to move inside the lock ... (or be volatile? but I think just move inside the lock)

PR #3382, opened 2024-04-04
Fix for #3381 - Empty OneToMany when overflowing LazyLoadBatchSize an…

### LINKED ISSUE #3381: Empty OneToMany when overflowing LazyLoadBatchSize and mixed queries
##  Context
We never had this problem in 4 years. It start doing this since we upgrade from Ebean 13.22.0 to 13.26.1.
We did this upgrade in order to fix the following issue : https://github.com/ebean-orm/ebean/issues/3319

We tried to upgrade to 15.0.2 but the issue still exists.

## Expected behavior

The linked entites should be loaded from the database.

## Actual behavior

We randomly get an empty array when our process access a oneToMany lazy-loaded attribute.
However, the associated entities are well stored in the database and should be returned.

Note : We did not succeed ot reproduce on local development, but this happens once every two days on our production server.

### Steps to reproduce

Order entity :
```java
@Entity
public class Order [...] {
    [...]
    @OneToMany(cascade = CascadeType.ALL, mappedBy = "order", fetch = FetchType.LAZY)
    @Column(nullable = true)
    private List<OrderLine> orderLines;
    [...]
} 
```
OrderLine entity :
```java
public class OrderLine [...] {
  [...]
    @ManyToOne
    @Column(nullable = true)
    private Order order;
  [...]
}
```
Business code that produce the problem :
```java
        try (Transaction transaction = DB.beginTransaction()) {
            elementsToProcess = Order.find
                    .query()
                    .forUpdate()
                    .where()
                    .isNotNull("dirty")
                    .orderById(true)
                    .setMaxRows(15)
                    .findList();

            process(elementsToProcess.stream().toArray(Order[]::new));

            transaction.commit();
        }
```
```java
    public void process(Order... orders){
            for (Order order: orders) {
                UpdateQuery query = DB.update(Order.class);
                
                [...]
                

	        BigDecimal totalPriceHt = order.getOrderLines() // HERE, THE LINES ARE EMPTY
                                                          .stream()
                                                          .map(PaymentHelper::getTotalPriceHT)
                                                          .reduce(BigDecimal.ZERO, BigDecimal::add);

                query.set("dirty", null)
                        .set("totalPriceHt", totalPriceHt);

                query.where()
                    .eq("id", order.getId())
                    .eq("dirty", order.getDirty())
                    .update(txn);
            }

            txn.commit();
        }
    }
```

```console
No specific console, everything is running fine, no error, no warn.
```

Yours faithfully,
LCDP

rbygrave:
…d mixed queries

14.0.0 via #3295 introduced this bug.

 #3295 introduced a behaviour where bulk updates clear the persistence context. Now the lazy loading of a BeanCollection works with an assumption that the "parent bean" is in the persistence context (and this assumption is broken with that change for this test case). That is, the bulk update is clearing the persistence context - removing the parent bean BEFORE the lazy loading is invoked ... and that doesn't then work because the parent bean isn't in the persistence context.

 This fix means that when lazy loading many's, the parent beans are putIfAbsent into the persistence context.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
@@ -3,6 +3,7 @@ package io.ebeaninternal.api;
 import io.ebean.CacheMode;
 import io.ebean.bean.BeanCollection;
 import io.ebean.bean.EntityBean;
+import io.ebean.bean.PersistenceContext;
 import io.ebeaninternal.server.core.BindPadding;
 import io.ebeaninternal.server.core.OrmQueryRequest;
 import io.ebeaninternal.server.deploy.BeanDescriptor;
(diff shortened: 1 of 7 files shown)


ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
```diff
@@ -57,25 +58,30 @@ public String description() {
...
-        if (lazy && !originIncluded && bc == originCollection) {
-          originIncluded = true;
-        }
-        idList.add(many.parentId(bc.owner()));
+        final var parent = bc.owner();
+        final var parentId = descriptor.getId(parent);
+        idList.add(parentId);
         bc.setLoader(server); // don't use the load buffer again
+        if (lazy) {
+          descriptor.contextPutIfAbsent(pc, parentId, parent);
```
rbygrave, line 72: This is the fix here really - `descriptor.contextPutIfAbsent(pc, parentId, parent);`

Issue #3129, opened 2023-07-04
DbJsonB immediately dirty on load if multiple enum as property


AntoineDuComptoirDesPharmacies:
Hi everyone,

We are having a strange behavior with `DbJsonB` feature on POJO where multiple properties are Enum.
Ebean Version : 13.20.1

## Expected behavior

Having an entity `A` with a `DbJsonB` property composed of a POJO `B` with multiple fields (each fields is an enum).
After DB/find.byId of entity `A` , `DB.beanState` should answer that the entity is not dirty.

## Actual behavior

Having an entity `A` with a `DbJsonB` property composed of a POJO `B` with multiple fields (each fields is an enum).
After DB/find.byId of entity `A` , `DB.beanState` answer that the entity is dirty on the `DbJsonB` property even without change.

### Steps to reproduce

```java

@Entity
public class A{
    [...]
    @DbJsonB
    @Column(nullable = false)
    private A.B sources;
    [...]

    @JsonInclude(JsonInclude.Include.NON_NULL) // Do not encode null to "null" to gain space in DB
    static public class B {
        private DataProviderEnum field1;
        private DataProviderEnum field2;

        public B() {
           // NOTHING TO DO
        }

        @JsonCreator
        public B(
                @JsonProperty("field1") DataProviderEnum field1,
                @JsonProperty("field2") DataProviderEnum field2
                                   ) {
            this.field1= field1;
            this.field2 = field2;
        }


        public DataProviderEnum getField1() {
            return field1;
        }

        public void setField1(DataProviderEnum field1) {
            this.field1= field1;
        }

        public DataProviderEnum getField2() {
            return field2;
        }

        public void setField2(DataProviderEnum field2) {
            this.field2= field2;
        }

        @Override
        public String toString() {
            return "ProductFieldSources{" +
                    "field1=" + field1+
                    ", field2=" + field2+
                    '}';
        }
    }
}

public enum DataProviderEnum {
    @EnumValue("VALUE1")
    VALUE1,
    @EnumValue("VALUE2")
    VALUE2,
    @EnumValue("VALUE3")
    VALUE3
}
```

Here is the value for this property in database : 
```
{"field1": "VALUE1", "field2": "VALUE2"}
```

When executing :
`DB.beanState(A.find.byId(1L)).dirtyValues()`
If mutationDetection = SOURCE :
=> Return the property `sources` with `Old` being `null` and `New` being the value from database.

If mutationDetection = DEFAULT :
[...]

PR #3402, opened 2024-05-14
#3129 Fix for 15.x to verify DBJson dirtiness using same field orderi…

AntoineDuComptoirDesPharmacies:
This pull request aim to fix a problem (#3129) where `@DbJsonB` fields are always considered dirty on load if the order of their properties do not match fields ordering from data stored in Postgres.

This is due to the fact that Postgres is re-ordering the fields if column type is a JsonB in order to optimize storage  / index / etc...

To solve this problem, we parse/format the value coming from `readSet` method in order to be sure that comparison will be done using Java Jackson encoded JSON on both side.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanProperty.java
@@ -713,9 +713,9 @@ public class BeanProperty implements ElPropertyValue, Property, STreeProperty {
   }
 
   /**
-   * creates a mutableHash for the given JSON value.
+   * creates a mutableHash for the given Object value.
    */
-  public MutableValueInfo createMutableInfo(String json) {
+  public MutableValueInfo createMutableInfo(Object value) {
     throw new UnsupportedOperationException();
   }
 
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/BeanPropertyJsonMapper.java
@@ -39,22 +39,22 @@ public final class BeanPropertyJsonMapper extends BeanPropertyJsonBasic {
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


AntoineDuComptoirDesPharmacies:
Hi @rbygrave,

We do not know why the pipeline is not successful on kotlin build.
On our computer, everything compile fine and tests pass.

You will certainly have a better insight about this failure.

Thanks in advance.
Yours faithfully,
LCDP

rbygrave:
Ah right, a couple of things here.

1. I broke the build - I broke the test kotlin module part of the build which needed a bump to the kotlin-querybean-generator
2. That test / test-kotlin module is conditionally run **_ONLY_** when **_JDK 11_** is used (so we can not notice the issue locally when we are not using JDK 11 which for me is most of the time as I'm 99% of the time locally on JDK 21 now). 
3. I fixed the build in **_master_** (but **_not_** in **_ebean-15x_** branch yet)

So your PR should be all good in the sense that the CI didn't pass because of my mistake.


Now, currently we operate with all changes going to master branch and then those changes are merged back to ebean-15x branch.

That is, any and all changes are applied to master and thus go into 14.x, 14.x-javax and 15.x ... so all new features and bug fixes go into master. When we release we release all 3 of these versions and they are all aligned in terms of features / bug fixes / version numbers ... with the exception that 15.x has some functionality removed.

So in general we want all PRs to be applied back to master. As part of the release I merge those changes to ebean-15x branch and resolve any conflicts as needed and release.

I'll look to merge that fix into the ebean-15x branch, but I think we really want this change to be merged back to master. I don't think we can just re-target this PR to master because it will then look to bring in the changes in ebean-15x, so I think we need a new PR based off master with these changes in it.

Can I get you to create a new PR based off master branch with these changes?  (and looking to merge those changes into master instead of ebean-15x).

AntoineDuComptoirDesPharmacies:
Hi @rbygrave,

This make perfect sense ! In the future we will be careful using `master` only for contributions.
Here is the new PR starting and targeting `master` branch : #3405 

Yours faithfully,
LCDP

PR #3422, opened 2024-06-18
Mark transaction as inactive in JtaTransactionManager during AfterCom…

VictorJimenezKwast:
…pletion to invalidate transaction ThreadLocal


rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3424

Issue #3423, opened 2024-06-18
JtraTransactionManager keeps reference to already closed scope (and Transaction/connection ThreadLocal) when JtaTxnListener#afterCompletion is called by a different thread


VictorJimenezKwast:
## Expected behavior
When the afterCompletion is called on a JtaTxnListener it should invalidate the corresponding transaction regardless of which thread registered this transaction.

## Actual behavior
We're running Ebean 14.3.1 on a JEE container (WildFly) with the JtaTransactionManager enabled. WildFly includes a process that keeps track of (long) running transactions called the [TransactionReaper](https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java). This background process aborts transactions and kills the underlying connection when it exceeds a certain pre-configured timeout.

When the Reaper detects a timeout it:
- calls the afterCompletion on the transaction (from the TransactionReaper thread)
- returns the connection to the pool
- causing the container to not call afterCompletion on the original thread since the Reaper has already done so.

This causes issues because the JtaTransactionManager stores the current transaction in a ThreadLocal (via TransactionScopeManager). This ThreadLocal is only cleaned up in the afterCompletion which does not work in this case because the TransactionReaper runs in a different Thread. After this has occurred all subsequent queries with Ebean will remain broken on that thread.

### Relevant logging:

TransactionReaper killing a transaction:
```
15:44:33,167 WARN  (Transaction Reaper) [com.arjuna.ats.arjuna] ARJUNA012117: TransactionReaper::check processing TX 0:ffff7f000101:-18aac15b:66718ee9:31f in state  RUN
15:44:33,168 WARN  (Transaction Reaper Worker 0) [com.arjuna.ats.arjuna] ARJUNA012095: Abort of action id 0:ffff7f000101:-18aac15b:66718ee9:31f invoked while multiple threads active within it.
15:44:33,171 WARN  (Transaction Reaper Worker 0) [com.arjuna.ats.arjuna] ARJUNA012381: Action id 0:ffff7f000101:-18aac15b:66718ee9:31f completed with multiple threads - thread default task-1 was in progress with java.base@11.0.20.1/java.net.SocketInputStream.$$YJP$$_jr$jni$socketRead0(Native Method)
java.base@11.0.20.1/java.net.SocketInputStream.socketRead0(SocketInputStream.java)
[...]

PR #3424, opened 2024-06-20
#3423 JtaTransactionManager keeps reference to already closed scope (…

rbygrave:
…and Transaction/connection ThreadLocal) when JtaTxnListener#afterCompletion is called by a different thread

Bugfix for "Long/slow transaction reaper" in Wildfire that can close/rollback a transaction in a different thread. Fix is to use active flag to inactivate the transaction and detect that case in JtaTransactionManager.getCurrentTransaction()


VictorJimenezKwast (approved):
:+1: Changes look good. Also tested it against our stack and problem scenario and this works well. Thanks!

PR #3446, opened 2024-07-15
Db2 forupdate fix

nPraml:
Hello Rob,

In our application we assume that the forUpdate() query locks the selected rows (that read access from another thread is no longer possible). It works on MariaDb and SqlServer, but not on Db2.

The DB2 documentation states that the rows are locked but are not protected against write access: https://www.ibm.com/docs/en/db2/11.5?topic=statement-update-clause 

We found out that if you add an orderBy("id") to the query, the locking also works under DB2. In our opinion it is just a workaround and not the solution. (And we do not understand why)

We found another way for DB2 forUpdate() to lock the rows: with `rs use and keep update locks`

Can you please look at the PullRequest and give us feedback?

Best regards
Noemi

### DIFF
--- a/ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
+++ b/ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
@@ -40,6 +40,56 @@ public class TestQueryForUpdate extends BaseTestCase {
   }
 
   @Test
+  public void testConcurrentForUpdate() throws InterruptedException {
+
+    ResetBasicData.reset();
+
+    Thread t1 = new Thread() {
+      @Override
+      public void run() {
+        try (final Transaction transaction = DB.beginTransaction()) {
+          System.out.println("Thread: before find");
+          DB.find(Customer.class)
+            .forUpdate()
+           // .orderBy().desc("1") // this would help by the locks in DB2
+            .findList();
+
+          System.out.println("Thread: after find");
+          try {
+            Thread.sleep(3000);
+          } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+          }
+          System.out.println("Thread: done");
+        }
+      }
+    };
+
+    t1.start();
+
+    Thread.sleep(100);
+
+    long start = System.currentTimeMillis();
+    try (final Transaction transaction = DB.beginTransaction()) {
+      if (isH2()) {
+        DB.sqlUpdate("SET LOCK_TIMEOUT 5000").execute();
+      }
+      System.out.println("Main: before find");
+      DB.find(Customer.class)
+        .forUpdate()
+        //.orderBy().desc("1") // this would help by the locks in DB2
+        .findList();
+
+      System.out.println("Main: after find");
+    }
+
+    start = System.currentTimeMillis() - start;
+
+    assertThat(start).isGreaterThan(2800);
+
+  }
+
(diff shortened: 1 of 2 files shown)


rPraml:
Did we use "forUpdate" in a wrong way?
We used it for synchronizing two or more application-instances when accessing a certain piece of code. In the special use case we have a shared NFS directory  where only one instance should access this directory.

This was done with a 
```java
try(Transaction txn = DB.beginTransaction()) {
   List configs = DB.find(DirectoryScanConfig).where(...).forUpdate().findList(); // other thread will block here
   // scan the directory in the config(s)
}
```
Now we found out, that this might be the wrong way. Reasons:
- it does not work in DB2 as we expected. And we cannot explain, why it would work with `order by id` - does DB2 lock the index?
- BTW: it also does not work on H2 as there is a default lock timeout of 2s. So an other thread will block only for this time (see test)

Nevertheless, we need something to do a synchronization over DB.
@rbygrave - if we misunderstand the purpose of `forUpdate`, what do you think, if we extend the API with a `query.lockRows()` or optionally `query.lockRows(timeout)` that can be used to implement this kind of locking. It can be implemented with `for update` by default, maybe with an infinity timeout on H2 or with `rs use and keep lock` on DB2

FYI: This locking is also used on DbMigration for some platforms

ebean-test/src/test/java/org/tests/basic/TestQueryForUpdate.java
```diff
@@ -39,6 +39,56 @@ public void testForUpdate() {
...
+        .forUpdate()
+        //.orderBy().desc("1") // this would help by the locks in DB2
+        .findList();
+
+      System.out.println("Main: after find");
+    }
+
+    start = System.currentTimeMillis() - start;
+
+    assertThat(start).isGreaterThan(2800);
```
rPraml, line 88: here we check, if the second thread blocks, while the first thread locks the data for at least 3000 ms.

platforms/db2/src/main/java/io/ebean/platform/db2/BaseDB2Platform.java
```diff
@@ -57,6 +57,6 @@ public PlatformIdGenerator createSequenceIdGenerator(BackgroundExecutor be, Data
   @Override
   protected String withForUpdate(String sql, Query.LockWait lockWait, Query.LockType lockType) {
     // NOWAIT and SKIP LOCKED not supported with Db2
-    return sql + " for update";
+    return sql + " with rs use and keep update locks";
```
rPraml, line 60: rs = read stablility
Maybe relevant doc links:
https://www.ibm.com/docs/en/db2/11.5?topic=statement-update-clause
https://www.ibm.com/docs/en/db2/11.5?topic=statement-isolation-clause
https://www.ibm.com/docs/en/db2/11.5?topic=issues-isolation-levels
https://www.ibm.com/docs/en/db2/11.5?topic=statement-lock-request-clause

rbygrave:
> Did we use "forUpdate" in a wrong way?

No, you used it in the way I would expect.

> it does not work in DB2 as we expected

Yes, DB2 is taking its own approach here imo.

> This locking is also used on DbMigration for some platforms

Not for DB2 though, DB2 is using "LogicalLocking" for DbMigration.


So yes, I'm happy with this PR.

PR #3449, opened 2024-07-30
ebean-test: Modify postgis platform to support both net.postgis and og.postgis

rbygrave:
Use Class.forName() to test existence of `net.postgis.jdbc.DriverWrapperLW` and if available use that driver rather than the older `org.postgis.DriverWrapperLW` one.

Also support explicitly specifying the driver to use via test property - ebean.test.postgis.driver


rbygrave:
Hi @Ryszard-Trojnacki , can you check if you are happy with this.

Ryszard-Trojnacki:
I have checked this and:
1. Works as expected. I added log to this and driver was selected correctly.
2. You should probably remove `NetPostgisSetup` now because it is not needed and remove it registration from `PlatformAutoConfig`: `KNOWN_PLATFORMS.put("netpostgis", new NetPostgisSetup());`.

rbygrave:
Great thanks.

> You should probably remove NetPostgisSetup now ...

Ah yes, good point. I'll do that.


Thanks !!

PR #3457, opened 2024-08-13
Update README.md fixing broken Foconis logo

focbenz:
Updated the Foconis AG logo with the new Foconis Analytics URL which was broken because of a deep link to a Yoomla logo that is no longer available.


focbenz:
Could also be changed to a static link to https://ebean.io/images/sponsor_foconis_logo.png

Followed with an update of the logo in the following website-source pull request,
https://github.com/ebean-orm/website-source/pull/65

Issue #3461, opened 2024-08-26
Add also() for query beans


artemik:
## Feature Request
Please add `also()` function (unconditional version of `alsoIf()`) into query beans. Example usage would be:
```
new QOrder()
    ...
    .also(isNotDeleted());
```

## Current Workaround
Hardcode `true` for conditional `alsoIf()`:
```
new QOrder()
    ...
    .alsoIf(() -> true, isNotDeleted());
```

### Why
Better readability. Code reuse. It would be very convenient for applying group of conditions that form a single meaning, especially when such logic could be shared between multiple query bean creations.

Possible examples are:
```
.also(isNotDeleted()) // some custom soft deletion logic
.also(withinDateRange(start, end)) // some date range check
.also(withPaging(pageParams)) // setFirstRow(...), setMaxRows(...) cursor paging
```

rbygrave:
I've created a PR that uses `also(Consumer<QueryBuilder<?,?>> apply)`.  So this uses QueryBuilder rather than `SELF` as my thinking is that this is for general functions to apply query features like paging. 

An alternative would be to use the more specific `Consumer<SELF>` instead and then adapt that noting that SELF is specific to each query bean type. This could be a touch more cumbersome when we would look to apply generic logic.   


The `Consumer<QueryBuilder<?,?>>` use looks like:

```java
  @Test
  void also() {
    var myPager = new MyPager(10);
    var q = new QCustomer()
      .select(name)
      .also(myPager)
      .query();

    q.findList();
    assertThat(q.getGeneratedSql()).contains("select /* QueryAlsoIfTest.also */ t0.id, t0.name from be_customer t0 limit 10");
  }

  static class MyPager implements Consumer<QueryBuilder<?,?>> {

    final int maxRows;

    MyPager(int maxRows) {
      this.maxRows = maxRows;
    }

    @Override
    public void accept(QueryBuilder<?, ?> queryBuilder) {
      queryBuilder.setMaxRows(maxRows);
    }
  }
```

artemik:
I'm afraid `Consumer<QueryBuilder<?,?>> apply`, won't allow to set something specific to a given query bean like `.also(withinDateRange(start, end)) // some date range check` which I think is very valuable.

Another concern is that such signature would differ from `alsoIf()`.

rbygrave:
> Is there some solution to support both?

[...]

PR #3465, opened 2024-08-27
#3461 - Add also() for query beans that takes QueryBuilder

rbygrave:


rbygrave:
Putting this PR with the generic QueryBuilder consumer function on hold.  Lets see how we go with the `also()` that takes the specific query bean first and see if we also desire this support for the QueryBuilder.

rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3497

PR #3485, opened 2024-10-01
Fix orderBy for cached many requests

nPraml:
Hello @rbygrave,

@rPraml and I have a test case in which we change the order of the ordered many properties. 
We then load the parent with a find() and the details with getDetails().

The master bean is loaded from the cache, but the details are loaded directly from the database because they are not in the cache. However, @OrderColumn is not taken into account and the details are delivered in the wrong order. We have an easy fix for this in this PR.

We have observed that the queries are different when you get the parent bean from cache or from the database. We tried to refactor so that the same query is produced in both cases, see our next PR.

Can you please take a look and give us feedback?

Best regards
Noemi

### DIFF
--- a/ebean-test/src/test/java/org/tests/cascade/TestOrderedList.java
+++ b/ebean-test/src/test/java/org/tests/cascade/TestOrderedList.java
@@ -188,4 +188,51 @@ public class TestOrderedList extends BaseTestCase {
     assertThat(masterDb.getDetails()).containsExactlyInAnyOrder(detail3, detail1);
 
   }
+
+  @Test
+  public void testModifyListWithCache2() {
+    final OmCacheOrderedMaster master = new OmCacheOrderedMaster("Master");
+    final OmCacheOrderedDetail detail1 = new OmCacheOrderedDetail("Detail1");
+    final OmCacheOrderedDetail detail2 = new OmCacheOrderedDetail("Detail2");
+    final OmCacheOrderedDetail detail3 = new OmCacheOrderedDetail("Detail3");
+    DB.save(detail1);
+    DB.save(detail2);
+    DB.save(detail3);
+    master.getDetails().add(detail1);
+    master.getDetails().add(detail2);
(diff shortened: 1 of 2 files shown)


nPraml:
the refactored version is in https://github.com/ebean-orm/ebean/pull/3486

ebean-test/src/test/java/org/tests/cascade/TestOrderedList.java
```diff
@@ -188,4 +188,51 @@ public void testModifyListWithCache() {
...
+    masterDb.getDetails().remove(2);
+    DB.save(masterDb);
+
+    LoggedSql.start();
+    OmCacheOrderedMaster masterDbNew = DB.find(OmCacheOrderedMaster.class, master.getId());
+    masterDbNew.getDetails().size();
+    assertThat(masterDbNew.getDetails()).containsExactly(detail2, detail1);
+    List<String> sql = LoggedSql.stop();
+    assertThat(sql).hasSize(1).first().asString()
+      .startsWith("select t0.id, t1.id, t1.name, t1.version, t1.sort_order, t1.master_id from om_cache_ordered_master t0 left join om_cache_ordered_detail t1 on t1.master_id = t0.id where t0.id = ? order by t1.sort_order;");
```
rPraml, line 226: Note: if the master bean is in cache, we take a different code path (defaultBeanLoader) instead of LoadManyRequest and this produces different queries.
IMHO this query is more complex, as it queries two tables.
In the refactored version (See [here](https://github.com/ebean-orm/ebean/pull/3486/files#diff-860cba4c5ae07ba9d4643752bb11ec9f2043a11f4604eec055ca3dcafaa9028eR229) we got the same query

rbygrave:
Yes I am happy with this PR, the change looks correct to me.

rPraml:
Hello Rob, please merge this solution first, as it fixes our problem. We will update the refactored version later with more details etc.

PR #3486, opened 2024-10-01
Fix orderBy for cached many requests - v2

nPraml:
Hello @rbygrave,

this is the refactored solution, the easy one is in https://github.com/ebean-orm/ebean/pull/3485

Please take a look and give us feedback.

Cheers
Noemi

@rPraml FYI

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
@@ -93,19 +93,8 @@ public final class LoadManyRequest extends LoadRequest {
 
   public SpiQuery<?> createQuery(SpiEbeanServer server) {
     BeanPropertyAssocMany<?> many = many();
-    SpiQuery<?> query = many.newQuery(server);
+    SpiQuery<?> query = many.newLoadManyQuery(server, onlyIds);
     query.usingTransaction(transaction);
-    String orderBy = many.lazyFetchOrderBy();
-    if (orderBy != null) {
-      query.orderBy(orderBy);
-    }
-    String extraWhere = many.extraWhere();
-    if (extraWhere != null) {
-      // replace special ${ta} placeholder with the base table alias
-      // which is always t0 and add the extra where clause
-      query.where().raw(extraWhere.replace("${ta}", "t0").replace("${mta}", "int_"));
-    }
-    query.setLazyLoadForParents(many);
     final var pc = loadContext.persistenceContext();
     many.addWhereParentIdIn(query, parentIdList(server, many, pc), loadContext.isUseDocStore());
     query.setPersistenceContext(pc);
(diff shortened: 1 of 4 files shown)


ebean-core/src/main/java/io/ebeaninternal/api/LoadManyRequest.java
```diff
@@ -117,15 +106,7 @@ public SpiQuery<?> createQuery(SpiEbeanServer server) {
...
     loadContext.configureQuery(query);
-    if (onlyIds) {
-      // lazy loading invoked via clear() and removeAll()
-      String mapKey = many.mapKey();
-      if (mapKey != null) {
-        query.select(mapKey);
-      } else {
-        query.select(many.targetIdProperty());
-      }
-    }
```
rPraml, line 128: CHECKME: Is ist ok, if the `onlyIds` code path is in `newLoadManyQuery` or may there be cases, where configureQuery overwrites the select?

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultBeanLoader.java
```diff
@@ -87,26 +92,12 @@ private void loadManyInternal(EntityBean parentBean, String propertyName, boolea
...
-      query.fetch(many.name());
-    }
-    if (filterMany != null) {
-      query.setFilterMany(many.name(), filterMany);
-    }
-
-    query.where().idEq(parentId);
-    query.setBeanCacheMode(CacheMode.OFF);
-    query.setMode(Mode.LAZYLOAD_MANY);
-    query.setLazyLoadManyPath(many.name());
```
rPraml, line 103: CHECKME: what do we need here? 
- I think we've lost the `CacheMode.OFF` for refresh? - see: https://github.com/ebean-orm/ebean/pull/3486/files#diff-5ce08d7970ceac437b60d46bbe9a40b0a4893f2211045684f0c29342c3322541R105
- Do we need `Mode.LAZYLOAD_MANY`

rPraml:
Would be good if we have some feedback from Rob, because I am unsure about the change ;)

rbygrave:
Hmmm, when we say "refactored solution" it can mean different things to different people. Generally a code reviewer wants to know what code was changed and why, and what code was simply moved but not actually changed [and maybe why it was moved / why include it in the PR?]. Looking to distill the "changes to the logic" and confirm there were no "Accidental changes" or "Changes with unknown impacts". The easiest PRs only change what they absolutely need to and nothing more.

In this PR I see some changes that don't look the same but I don't know if that was actually the intention or something else?

Can you state what the changes here are and why they were maed? For example, why change findOne into findList? Why was that done etc.

You can do this by putting your own comments into the PR - 1 per change. Like "This code was moved over to other location with no other changes", "findOne changed to findList because ...". Add those comments in and then I can review this again.

rbygrave:
This PR can be closed now right? as we merged in the other alternative PR.

nPraml:
Hello @rbygrave,

sorry I was on vacation, hence the late reply.

I probably wouldn't have time to work on this PR for the next few weeks.
I would still leave the PR open because we noticed in https://github.com/ebean-orm/ebean/pull/3485/files#r1783081211 that the queries that should actually look the same are different (difference: the master bean is in the cache or not). For this reason, we started refactoring so that queries are only built in a central location.

Is it ok for you?

rbygrave:
Closing this as alternate PR https://github.com/ebean-orm/ebean/pull/3485 was merged

PR #3487, opened 2024-10-02
Bug: Batch ordering - Regression introduced in version 13.26.1, insertAll() leaves batchMode=true enabled ... when it should not

nPraml:
Hello @rbygrave,

we are currently in the process of updating from ebean 13 to 14 and we encountered a bug, see the test case.

The entity `Contact` has a `@ManyToOne` property (without cascadeType) `ContactGroup`.

In a transaction we store some elements using `DB.insertAll()`. Then we save a contact (without a group), we create a group with a certain id, we save it and then we put the group in the contact.

With transaction flush, however, the order in the queue is not correct: contact (with group) is saved first, but group does not yet exist in the DB.

What we discovered during debugging: after `DB.insertAll()`, txn.batchMode = true is set (this was not the case with ebean 13) and during flush the objects contact and group are in the wrong order in the queue.

With ebean 13 you have to explicitly call `txn.setBatchMode(true)` for the test to fail.

Could you please take a look and help us fix this?

Cheers,
Noemi

@rPraml FYI

### DIFF
--- a/ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
+++ b/ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
@@ -15,6 +15,7 @@ import io.ebean.xtest.base.DtoQuery2Test;
 import io.ebeaninternal.api.SpiTransaction;
 import org.junit.jupiter.api.Test;
 import org.tests.model.basic.*;
+import org.tests.query.cache.Acl;
 
 import java.sql.Timestamp;
 import java.util.ArrayList;
@@ -357,6 +358,24 @@ public class TestBatchInsertFlush extends BaseTestCase {
 
     } finally {
       txn.end();
+    }
+  }
+
+  @Test
+  public void testBatchEscalationInsert() {
+    try (Transaction txn = DB.beginTransaction()) {
+      assertThat(txn.isBatchMode()).isFalse();
+      DB.insertAll(List.of(new Acl("test")));
+      assertThat(txn.isBatchMode()).isFalse();
+    }
+  }
+
+  @Test
+  public void testBatchEscalationSave() {
+    try (Transaction txn = DB.beginTransaction()) {
+      assertThat(txn.isBatchMode()).isFalse();
+      DB.saveAll(List.of(new Acl("test")));
+      assertThat(txn.isBatchMode()).isFalse();
     }
   }
 }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -1689,6 +1689,7 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
       for (Object bean : beans) {
         persister.insert(checkEntityBean(bean), options, txn);
       }
+      txn.flushBatchOnCollection();
       return 0;
     }, transaction);
   }


nPraml:
if I modify `insertAll` on the same way as `saveAll`:

image

it fixes the test, but the test case fails if I use instead `insertAll` a `trx.setBatchMode(true)`

rbygrave:
> `txn.flushBatchOnCollection()`

Yes, this bug is a regression that was introduced in version `13.26.1` by #3324 line - https://github.com/ebean-orm/ebean/pull/3324/files#diff-0804968369579e408eea483a9e582485cd7faf94e5d78dc8f7a918c729962a70R1695

It should have that `txn.flushBatchOnCollection()`


> but the test case fails if I use ... trx.setBatchMode(true)

That is expected by me.

The thinking here is that this isn't a case of "Cascading" persistence [where ebean is determining the "depth" and ordering based on the "depth"] but **_instead_** a case where both Contact and ContactGroup are explicitly saved as "Top Level" via `database.save()`

So when we have:

```java
...
database.save(contact);
...
database.save(contactGroup);

```
The contact was saved before the contactGroup ... and these are both "Top Level" ... and so the ordering is then down to the order in which those are saved, which is contact first.

Does that make sense?

rPraml:
Hello Rob, unfortunately the places in our code are a bit spread out. One part of the code creates the contacts, the other the contact groups. A third then takes care of mapping both.
Imagine there are 3 CSV files (`contacts.csv`, `groups.csv` and `mapping.csv`) I tried to model this in a unit test

ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
```diff
@@ -359,4 +369,79 @@ public void testFlushOnSetProperty() {
...
+
+
+  @Test
+  public void testWithBatch() {
+
+    ResetBasicData.reset();
+
+    Map<Integer, String> contactCsv = Map.of(123, "Noemi", 124, "Roland");
+    Map<Integer, String> contactGroupCsv = Map.of(42, "Praml", 43, "Max");
+    Map<Integer, Integer> mappingCsv = Map.of(123, 42);
```
rPraml, line 412: 3 csv files.

ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
```diff
@@ -359,4 +369,79 @@ public void testFlushOnSetProperty() {
...
+      Customer someCustomer = DB.find(Customer.class).setMaxRows(1).findOne(); // just to get no error
+      // Save contact without group assigned
+
+      Map<Integer, Contact> createdContacts = new HashMap<>();
+      contactCsv.forEach((id, name) -> {
+        Contact contact = new Contact();
+        contact.setFirstName(name);
+        contact.setId(id);
+        contact.setCustomer(someCustomer); // customer is mandatory
+        DB.save(contact);
```
rPraml, line 425: Here we create the contacts and same them in a map for further mapping (I know, there are better ways, but this is what our code does. There are up to 100 objects in a temporary map)

ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
```diff
@@ -359,4 +369,79 @@ public void testFlushOnSetProperty() {
...
+        group.setName(name);
+        group.setId(id);
+        DB.save(group);
+      });
+
+
+      mappingCsv.forEach((contactId, groupId) -> {
+        Contact contact = createdContacts.get(contactId);
+        contact.setGroup(DB.reference(ContactGroup.class, groupId));
+        DB.save(contact);
```
rPraml, line 441: The last code part does some post-processing on some of the objects (the objects may not be saved, yet due ebean batching)

Yes, there are several ways, how to solve this issue
- do a `DB.saveAll(createdContacts.value())`
- disable batching
- do a flush
- write a better import-code 😉

I'm unsure, if ebean can do much here. It must know or adjust the save-order, but this may rise up new problems

rbygrave: > I'm unsure, if ebean can do much here

I know its an example but to me its pretty clear that import code should look to first materialise and save the [ContactGroup] beans that will be later _referenced_ by other beans [Contact].

Ebean gives the application **_exact control_** over this. With that control, this import can be done as batch inserts of ContactGroup followed by batch inserts of Contact with no actual updates required [optimal handling of this from the database perspective].

You are going to have a hard time convincing me that ebean should change behaviour here.

What we want is for ebean to give developers the ability to control **_exactly_** how this works because at certain levels of scale we absolutely need that control. Trying to get "fancy" here would not end well.



> do a DB.saveAll(createdContacts.value())

So, do that and rely on ebean cascade persisting from Contact to ContactGroup [and then ebean will determine the ordering because we are now using cascade persist and ebean will get it right]. Yes, a good option.


> disable batching

At some scale people regret this. I say invest the time and think about ordering and stay with jdbc batch [and get orders of magnitude performance benefit from that investment]. I'd personally never go without batching here.


> do a flush

If we really must. Doing so, implies that we'll see _extra_ updates being used when if the ordering was correct the app would avoid those extra updates. So generally not optimal.


> write a better import-code

IMO this is actually the best answer. Actually spend some time thinking about the ordering and design import processing with that in mind, looking to use JDBC batch and avoiding extra updates when possible.

rPraml:
> You are going to have a hard time convincing me that ebean should change behaviour here.

Everything is fine, we just noticed this "regression" because with ebean 14 some unit tests in our application failed because batching was active after insertAll.

We have the situation here that many people work on our code base and (unfortunately) often produce suboptimal code, as in this case.

Fortunately, we also have a lot of unit tests, so we can find most of these errors quickly.

If you say it "works as planned", then I also have good arguments in our company that maybe some parts of the code need to be revised.

rPraml:
Stripped down the testcase

PR #3491, opened 2024-10-09
Add setArrayParameter to SqlQuery

ajcamilo:
The `DB.sqlQuery` was the only API that did not support array parameters.

This PR adds the feature.

```
DB.sqlQuery("select * from table where id in (?)")
  .setArrayParameter("ids", List.of(1, 2, 3, 4))
  ...
```


rbygrave:
Awesome, thanks.

rbygrave (approved):

PR #3496, opened 2024-10-16
fixed flaky test in ToStringBuilderTest.java

mumbler6:
The order of items in the `map` variable for the `ParamStore` class is indeterministic, causing the `map_recurse()` test case to sometimes fail. Since `map` is a HashMap and the items inside are unordered, the conversion to string doesn't always guarantee the same string.

This PR proposes to change the `map` variable from a HashMap to a LinkedHashMap, preserving the ordering of insertion and removing the flakiness of the test. 

This change was confirmed by running the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, which explores and reports errors in different behaviors of under-determined Java APIs.

To reproduce this problem, you can run the test with NonDex using this command:

```
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -pl ebean-api -Dtest=io.ebean.bean.ToStringBuilderTest#map_recurse
``` 
Here are screenshots of the outputs produced by NonDex before and after the fix: 

image

image

Please let me know if you'd like to discuss the fix.

### DIFF
--- a/ebean-api/src/test/java/io/ebean/bean/ToStringBuilderTest.java
+++ b/ebean-api/src/test/java/io/ebean/bean/ToStringBuilderTest.java
@@ -213,7 +213,7 @@ class ToStringBuilderTest {
 
     final int id;
 
-    final Map<String, ParamValue> map = new HashMap<>();
+    final Map<String, ParamValue> map = new LinkedHashMap<>();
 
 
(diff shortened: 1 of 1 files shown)


rbygrave:
Nice, thanks !!

rbygrave (approved):

PR #3498, opened 2024-10-18
Migrate to JSpecify NullMarked/Nullable from avaje.lang NonNullApi/Nullable

rbygrave:


rPraml:
Can you write 1-2 sentences, what's the long term plan here and why you have chosen jspecify and not one of the other 123 null annotations 😉 (e. g. jetbrains or findbugs)

rbygrave:
>  why you have chosen jspecify and not one of the other 123 null annotations

Jspecify is a project that has support from Jetbrains, Spring, but also Oracle, Google, Microsoft, Meta - https://jspecify.dev/about/
... it's goal is to be that set of unifying null annotations. So yes I expect Spring to adopt it at some point. The Google projects like Guava are adopting it. It will take time but I do hope for it to unify things (has a chance of this given the companies and projects that support it). 

The Lead was  Kevin Bourrillion Google, Guava etc
 
Maybe this video helps: https://www.youtube.com/watch?v=HS_kA42YNkU

PR #3504, opened 2024-10-22
Fix: dynamic property group by with alias

AlexWagner:
If I use formula in dto queries e.g. concat(...) and i want count them,
I have to specify an alias to map it to the dto.
Then this alias will be put into the group by clause additionally to the formula, this leads to an incorrect query.
see the test

I have implemented an easy fix for this.


rbygrave (approved):

rbygrave:
Nice work !!

PR #3506, opened 2024-10-24
Fix/Remove some deprecated warnings

rPraml:
This PR removes some (not yet all) deprecated warnings during compilation.

I've organized this into separate commits.

There should be no effective code changes in this PR

Roland

### DIFF
--- a/ebean-api/src/main/java/io/ebean/DB.java
+++ b/ebean-api/src/main/java/io/ebean/DB.java
@@ -605,7 +605,7 @@ public final class DB {
    *   // find orders and their customers
    *   List<Order> list = DB.find(Order.class)
    *     .fetch("customer")
-   *     .order("id")
+   *     .orderBy("id")
    *     .findList();
    *
    *   // sort by customer name ascending, then by order shipDate
--- a/ebean-api/src/main/java/io/ebean/Database.java
+++ b/ebean-api/src/main/java/io/ebean/Database.java
@@ -400,7 +400,7 @@ public interface Database {
    *   // find orders and their customers
    *   List<Order> list = database.find(Order.class)
    *     .fetch("customer")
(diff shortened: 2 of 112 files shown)


rbygrave:
So generally I only do the order() -> orderBy() changes when we remove the deprecated method in question. The reasons are that it keeps our test coverage on the deprecated methods, it shows we actually had test coverage in the first place, the PR that removes the deprecated method provides examples of the impact [for people looking at the PR].

So to me these should have been separate PRs and then the ones like order() -> orderBy() only get merged with the actual removal of the deprecated method. Removing deprecated methods is significant and so they deserve their own PR.

I realise this is a "grey area" and I appreciate the effort but I don't want to merge this PR in this form.

rPraml:
Can you cherry pick that commit: https://github.com/ebean-orm/ebean/pull/3506/commits/c379abda50d7bdbad12275496adaabeaf92e9c56
Or would it be better if I do another PR?

rPraml:
> for people looking at the PR because they didn't look at the warnings and now their code doesn't compile

Haha, I know this problem too good. 🤣🤣

We've adjusted our compiler settings few months ago, that build will fail, if you use a deprecated forRemoval code. The only way is to mark that code part with `@SupressWarinings("removal")` 

I've currently the issue, that an ebean build prints out hundreds of warnings on a build (Starting with `'build.plugins.plugin.version' for org.apache.maven.plugins:maven-javadoc-plugin is missing` - javadoc generation also fails on my machine)

I would be willing to support you a little here and clear these things up. That's why I startet with the (at least for me) most annoying thing, that IntelliJ underlines `order()` in red in every second test class.

I've exctracted the orderby thing in a separate PR: https://github.com/ebean-orm/ebean/pull/3508 and I will close this for now

PR #3512, opened 2024-10-29
Invalid SQL for exist queries with many to many relation.

jskrobko:
Hello @rbygrave ,
We have identified the following issue: when performing an exists query with a One-To-Many relationship, correct SQL is generated regardless of whether an id is present or not. (The generated SQL does not include unnecessary joins).
However, when executing the same type of query with a Many-To-Many relationship, different SQL is being generated:
- without id: the SQL is incorrect. It contains `sq1.id = t0.null`.
- with id: the SQL works, but includes an unnecessary join: `left join mt_permission t1 on t1.id = t1z_.mt_permission_id`.

Perhaps you could take a look at this. We have 2 tests with asserts where we are expecting different results.

Juri

### DIFF
--- a/ebean-test/src/test/java/org/tests/m2m/TestM2MQueryAndIds.java
+++ b/ebean-test/src/test/java/org/tests/m2m/TestM2MQueryAndIds.java
@@ -56,7 +56,8 @@ public class TestM2MQueryAndIds extends BaseTestCase {
     Query<Role> query = DB.find(Role.class).select("id").where().exists(sq1).query();
     List<Role> models = query.findList();
     assertThat(models).hasSize(1);
-    assertThat(query.getGeneratedSql()).isEqualTo("select distinct t0.id from mt_role t0 left join mt_role_permission t1z_ on t1z_.mt_role_id = t0.id left join mt_permission t1 on t1.id = t1z_.mt_permission_id where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t1.id))");
+    assertThat(query.getGeneratedSql()).isEqualTo("select distinct t0.id from mt_role t0 left join mt_role_permission t1z_ on t1z_.mt_role_id = t0.id where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t1z_.mt_permission_id))");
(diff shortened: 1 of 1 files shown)


jskrobko:
Error:    TestM2MQueryAndIds.testM2MWithId:59 
expected: "select distinct t0.id from mt_role t0 left join mt_role_permission t1z_ on t1z_.mt_role_id = t0.id where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t1z_.mt_permission_id))"
 but was: "select distinct t0.id from mt_role t0 left join mt_role_permission t1z_ on t1z_.mt_role_id = t0.id left join mt_permission t1 on t1.id = t1z_.mt_permission_id where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t1.id))"
Error:  Errors: 
Error:    TestM2MQueryAndIds.testM2MWithoutId:74 » Persistence Query threw SQLException:Syntax error in SQL statement "select t0.id from mt_role t0 where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t0.[*]null))"; expected "identifier"; SQL statement:
select t0.id from mt_role t0 where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t0.null)) [42001-220] Bind values:[null] Query was:select t0.id from mt_role t0 where exists (select 1 from mt_permission sq1 where sq1.name = ? and (sq1.id = t0.null))

rbygrave:
> without id is `.raw("(sq1.id = permissions)")` and with id is `.raw("(sq1.id = permissions.id)")`

The "without id" expression isn't valid. 

I am thinking that is confusing due to how ManyToOne paths get translated to the foreign key column. In this case,  `permissions` is a ToMany path and so it literally only gets translated into a table alias and not a table alias + column. Hence we end up with `t0.null`.


> with id: the SQL works, but includes an unnecessary join

That is because this path is a ManyToMany, so we get the join to the "hidden" intersection table and the join to the other side of the ManyToMany.

jskrobko:
Hello Rob,
We discussed this again as a team. We were not aware that a query without 'id' is not valid.
We can confirm that the result is correct in both O2M, M2M, and M2O when using '.id'. The only point here is that the query in M2M can still be optimized (remove additional 'join').

Do you still want to pursue this further? Otherwise, the issue can be closed.
Thank you.
Juri

rbygrave:
> Do you still want to pursue this further? Otherwise, the issue can be closed.

Let just close this one at this stage.

PR #3519, opened 2024-11-05
No productive code change: cleanup basic test data

nPraml:
Hello @rbygrave,

our ebean fork has a green build in github and local, but not on our Bamboo server. 

Reason: one test is based on the 4 basic customers in the DB and other tests create customers that are never cleaned up. The execution order is different on different servers, so it is a matter of luck whether the build goes through.

I have adjusted the affected tests to clean up their test data, so there is no longer a warning in the build log with "WARNING: basic test data was modified. Current content:"

Can you take a look at it?

Cheers
Noemi

### DIFF
--- a/ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
+++ b/ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
@@ -55,6 +55,9 @@ public class TestBatchInsertFlush extends BaseTestCase {
       DB.saveAll(customers);
 
       txn.commit();
+    } finally {
+      DB.find(Customer.class).where().startsWith("name", "BatchFlush").delete();
+      DB.find(Contact.class).where().startsWith("firstName", "Fred").startsWith("lastName", "Blue").delete();
     }
   }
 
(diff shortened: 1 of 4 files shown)


ebean-test/src/test/java/org/tests/batchinsert/TestBatchInsertFlush.java
```diff
@@ -42,19 +42,22 @@ public void batchFlush() {
...
         Customer customer = ResetBasicData.createCustomer("BatchFlushPreInsert " + i, null, null, 3);
-        customer.addContact(new Contact("Fred" + i, "Blue"));
+        customer.addContact(new Contact("BatchFlush" + i, "Blue"));
         customers.add(customer);
       }
 
       for (int i = 3; i < 6; i++) {
         Customer customer = ResetBasicData.createCustomer("BatchFlushPostInsert " + i, null, null, 3);
-        customer.addContact(new Contact("Fred" + i, "Blue"));
+        customer.addContact(new Contact("BatchFlush" + i, "Blue"));
```
nPraml, line 51: I renamed it, because ResetBasicData.reset() creates also contacts with `Fred` (So the delete in `finally` is easier

rbygrave (approved):

rbygrave:
Looks great, thanks, merging.

Issue #56, opened 2013-11-04
Using findIterate with MySQL streams the entire result set at once


Ronnie76er:
Using findIterate streams the entire result set.  If it's particularly large, this could cause a Java OutOfMemoryError when calling findIterate.

According to the MySQL JDBC documentation, it seems like the statement needs to be constructed like this for streaming to work correctly: 

``` java
stmt = conn.createStatement(java.sql.ResultSet.TYPE_FORWARD_ONLY,
              java.sql.ResultSet.CONCUR_READ_ONLY);
stmt.setFetchSize(Integer.MIN_VALUE);
```

MySQL JDBC doc: http://dev.mysql.com/doc/refman/5.5/en/connector-j-reference-implementation-notes.html
Google Group Topic: https://groups.google.com/forum/#!topic/ebean/PaMDPEetbJI

PR #3523, opened 2024-11-27
set defaultFetchBuffer in findEach/findList in SQL and DTO Queries

rPraml:
Hello Rob,

I've made some tests and found out, that MariaDb (and maybe other DBs, too) do not use streaming and cause an OOM

As this is very related to #56, I would like to ask, if you think, we need the forward_only here. (or if the forward-only is no longer needed as all)


rbygrave:
This does not apply to ORM queries - applies to SqlQuery and DtoQuery only.

> forward-only is no longer needed 

I'm not sure. I'd need to look again at how MySql works there.

rbygrave (approved):

PR #3528, opened 2024-12-14
Use Avaje Prisms in QueryBean APT

SentryMan:
- Uses Avaje Prisms to extract annotation data
- Make Processing Context static
- Validates when a module-info does not have the `provides` clause for the generated register


SentryMan:
@rbygrave consider me baffled, it works on my local

querybean-generator/src/main/java/io/ebean/querybean/generator/ModuleMeta.java
```diff
@@ -1,21 +0,0 @@
-package io.ebean.querybean.generator;
-
-import java.util.List;
-
-class ModuleMeta {
```
rbygrave, line 5: How did this get removed?

querybean-generator/src/main/java/io/ebean/querybean/generator/PomPluginWriter.java
```diff
@@ -0,0 +1,59 @@
+package io.ebean.querybean.generator;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+
+final class PomPluginWriter {
```
rbygrave, line 7: This isn't a minimalist change ... it's an "Extra Feature" ... these changes make the PR bigger and increase the chance of it not getting approved.  Try and stay minimalist with a view to iterating rather than doing "all things" ... it takes away attention from the issues that stop this PR from being approved.

querybean-generator/src/main/java/io/ebean/querybean/generator/ProcessingContext.java
```diff
@@ -16,118 +33,68 @@
...
+    private final Set<String> services = new TreeSet<>();
 
-  private final Set<String> otherClasses = new TreeSet<>();
+    private final PropertyTypeMap propertyTypeMap = new PropertyTypeMap();
 
-  /**
-   * The DB name prefixed entities.
-   */
-  private final Set<String> prefixEntities = new TreeSet<>();
+    /** All entity packages regardless of DB (for META-INF/ebean-generated-info.mf). */
```
rbygrave, line 49: If you put the formatting back to the original ... then this does not become a DIFF ... and that helps me review it. Adding these types of DIFFs does not help a big PR.

querybean-generator/src/main/java/io/ebean/querybean/generator/Processor.java
```diff
@@ -1,123 +0,0 @@
...
-import javax.lang.model.element.Element;
-import javax.lang.model.element.TypeElement;
-import java.util.Arrays;
-import java.util.LinkedHashSet;
-import java.util.Set;
-
-/**
- * Process compiled entity beans and generates 'query beans' for them.
- */
-public class Processor extends AbstractProcessor implements Constants {
```
rbygrave, line 16: Doing the rename here means the DIFF isn't showing now ...

querybean-generator/src/main/java/io/ebean/querybean/generator/ModuleMeta.java
SentryMan: because we can extract the data directly from the annotation instead of putting it in a proxy class

querybean-generator/src/main/java/io/ebean/querybean/generator/ProcessingContext.java
SentryMan: all of these fields are going into a nested class class, so formatting has to change no?

querybean-generator/src/main/java/io/ebean/querybean/generator/Processor.java
SentryMan: should be good now

querybean-generator/src/main/java/io/ebean/querybean/generator/PomPluginWriter.java
SentryMan: removed

SentryMan:
@rbygrave please re-review when you get time

SentryMan:
this will have to be redone later

PR #3530, opened 2024-12-18
FIX: IllegalAccessError on modular DTO queries

SentryMan:
When using a dto query on a JPMS application, I get an IllegalAccessError even though I export and open my package.

This change allows for the lookup of public method/constructor method handles in JPMS


rbygrave:
Nice thanks!! Gee, I read the javadoc again and I still wouldn't have understood it correctly without this issue/pr.

rbygrave (approved):

PR #3534, opened 2024-12-20
FIX: export io.ebeaninternal.api to support `@Transactional` on the module path 

### LINKED ISSUE #3533: BUG: @Transactional doesn't work when running on the module-path
## Expected behavior
The transactional annotation should work like manually creating a transaction with `DB.beginTransaction()`

## Actual behavior
```
java.lang.IllegalAccessError: 
class com.avaje.jdk.realworld.web.ProfilesController (in module avaje.realworld) 
cannot access class io.ebeaninternal.api.HelpScopeTrans (in module io.ebean.core) 
because module io.ebean.core does not export io.ebeaninternal.api to module avaje.realworld
```
### Steps to reproduce
Have a module-info and a method annotated with transactional 
```java
  @Transactional
  void useTransactional() {
  //contents don't matter
  }
```

SentryMan:
resolves #3533 by exporting the inaccessible module

### DIFF
--- a/ebean-core/src/main/java/module-info.java
+++ b/ebean-core/src/main/java/module-info.java
@@ -47,7 +47,7 @@ module io.ebean.core {
   exports io.ebeanservice.docstore.api.support to io.ebean.elastic, io.ebean.test;
   exports io.ebeanservice.docstore.api.mapping to io.ebean.elastic;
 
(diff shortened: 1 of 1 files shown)


SentryMan:
internals should really stay internal, will figure something else out

Issue #3531, opened 2024-12-19
Enhancement: Service Load `MethodHandles.Lookup`


SentryMan:
When using JPMS, `MethodHandles.publicLookup()`/`MethodHandles.lookup()` as used by ebean currently requires a user to add an `exports my.package` directive to allow ebean to access the fields/methods/constructors of entities/DTOs. 

The way around this is to pass a `MethodHandles.Lookup` from the entity/DTO class's module for ebean to use. I'm thinking we could create a new interface:
```java
import java.lang.invoke.MethodHandles.Lookup;

public interface LookupProvider {

  Lookup provideLookup();

}
```
And have the querybean generator generate an implementation that can be service loaded.

```java
public class Example implements LookupProvider {

  private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();

  @Override
  public Lookup provideLookup() {
    return LOOKUP;
  }
}
```

In this way, ebean can gain free reign to access the fields/methods/constructors of entities/DTOs without needing `exports`/`opens` directives from the user.

Related to #3530

Would like to get #3528 merged before I start on this

SentryMan:
Next step is to enhance the generator to generate a class implementing the new interface

PR #3535, opened 2024-12-20
Feature: Add a `MethodHandles.Lookup` SPI

SentryMan:
- Adds new SPI interface to retrieve `MethodHandles.Lookup`
- Adds a new class to contain all the lookups by module name
- replace reflective bean construction with MethodHandles

Part of #3531


ebean-core/src/main/java/io/ebeaninternal/server/type/DefaultTypeManager.java
```diff
@@ -597,19 +599,28 @@ private ScalarTypeEnum<?> createEnumScalarType(Class enumType, Map<String, Strin
...
             // first try objectMapper constructor
-            scalarType = cls.getDeclaredConstructor(ObjectMapper.class).newInstance(objectMapper);
+            scalarType =
+                (ScalarType<?>)
+                    lookup
+                        .findConstructor(cls, MethodType.methodType(ObjectMapper.class))
+                        .invoke(objectMapper);
           } catch (NoSuchMethodException e) {
-            scalarType = cls.getDeclaredConstructor().newInstance();
+            scalarType =
```
rbygrave, line 617: I think this is supposed to be  `Lookups.newDefaultInstance(foundType)` ?

ebean-core/src/main/java/io/ebeaninternal/server/type/DefaultTypeManager.java
```diff
@@ -597,19 +599,28 @@ private ScalarTypeEnum<?> createEnumScalarType(Class enumType, Map<String, Strin
   private void initialiseCustomScalarTypes(BootupClasses bootupClasses) {
     for (Class<? extends ScalarType<?>> cls : bootupClasses.getScalarTypes()) {
       try {
+        var lookup = Lookups.getLookup(cls);
         ScalarType<?> scalarType;
         if (objectMapper == null) {
-          scalarType = cls.getDeclaredConstructor().newInstance();
+          scalarType =
```
rbygrave, line 605: I think this is supposed to be  `Lookups.newDefaultInstance(foundType)` ?

ebean-api/src/main/java/io/ebean/lookup/Lookups.java
```diff
@@ -0,0 +1,30 @@
+package io.ebean.lookup;
```
rbygrave, line 1: I think this could live in the `io.ebean.plugin` package, hmm.

ebean-core/src/main/java/io/ebeaninternal/server/type/DefaultTypeManager.java
```diff
@@ -599,17 +601,17 @@ private void initialiseCustomScalarTypes(BootupClasses bootupClasses) {
       try {
         ScalarType<?> scalarType;
         if (objectMapper == null) {
-          scalarType = cls.getDeclaredConstructor().newInstance();
+          scalarType = Lookups.newDefaultInstance(cls);
         } else {
           try {
             // first try objectMapper constructor
-            scalarType = cls.getDeclaredConstructor(ObjectMapper.class).newInstance(objectMapper);
+            scalarType = Lookups.newDefaultInstance(cls);
```
rbygrave, line 608: This one is not the default constructor but takes an ObjectMapper

rbygrave:
Ok, looks pretty good. Are you happy with it / no more expected changes?

SentryMan:
I don't expect anything else

Issue #3541, opened 2025-01-16
[querybean-generator] FilerException trying to write EntityClassRegister with Quarkus dev mode / hot reload


rbygrave:
When using Quarkus dev mode / hot reload with the querybean-generator we can get the compilation error below.

This is because with the dev mode / hot reload the annotation processor tries to write the EntityClassRegister which already exists.

The fix for this issue is to not log this at ERROR level (which effectively means the compilation failed) but instead log it at WARN or INFO/NOTE level.

### Steps to reproduce

Using Quarkus dev mode / hot reload with the querybean-generator 
- `mvn Quarkus:dev`
- Trigger the "resume testing" `r`
- Wait for tests to all run
- Make some reasonable change to an entity bean

```console
Compilation Failed:
error: FilerException trying to write EntityClassRegister error:javax.annotation.processing.FilerException: Attempt to recreate a file for type org.acme.domain.EbeanEntityRegister stack:[jdk.compiler/com.sun.tools.javac.processing.JavacFiler.checkNameAndExistence(JavacFiler.java:741), jdk.compiler/com.sun.tools.javac.processing.JavacFiler.createSourceOrClassFile(JavacFiler.java:498), jdk.compiler/com.sun.tools.javac.processing.JavacFiler.createSourceFile(JavacFiler.java:435), io.ebean.querybean.generator.ProcessingContext.createWriter(ProcessingContext.java:399), io.ebean.querybean.generator.SimpleModuleInfoWriter.<init>(SimpleModuleInfoWriter.java:31), io.ebean.querybean.generator.Processor.writeModuleInfoBean(Processor.java:90), io.ebean.querybean.generator.Processor.process(Processor.java:53), jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:1021), ...

```
After this fix is applied, and this is now logged as `WARN` we get:

```console
2025-01-16 15:49:13,221 INFO  [io.quarkus] (Quarkus Main Thread) Profile dev activated. Live Coding activated.
2025-01-16 15:49:13,221 INFO  [io.quarkus] (Quarkus Main Thread) Installed features: [cdi, rest, smallrye-context-propagation, vertx]
[...]

Issue #3539, opened 2025-01-09
Build fail with IDEA 2024.3.x


spinachomes:
Build fail with IDEA 2024.3.x,but it's ok with IDEA 2024.2.x

### Steps to reproduce
use IDEA build project with jdk21 and ebean 15.8.1
```java
  // some java code
```

```console
[...]

PR #3542, opened 2025-01-16
[querybean-generator] FilerException trying to write EntityClassRegister with Quarkus dev mode / hot reload #3541

rbygrave:
Drop to log at WARN level when we catch FilerException such that Quarkus dev mode / hot reload works

The effective change here is to catch FilerException and log at WARN rather than ERROR:
```java
} catch (FilerException e) {
  processingContext.logWarn(null, "FilerException trying to write EntityClassRegister: " + e);
}
```

### DIFF
--- a/kotlin-querybean-generator/src/main/java/io/ebean/querybean/generator/ProcessingContext.java
+++ b/kotlin-querybean-generator/src/main/java/io/ebean/querybean/generator/ProcessingContext.java
@@ -415,9 +415,10 @@ class ProcessingContext implements Constants {
     messager.printMessage(Diagnostic.Kind.ERROR, String.format(msg, args), e);
   }
 
-  /**
-   * Log a info message.
-   */
+  void logWarn(String msg, Object... args) {
+    messager.printMessage(Diagnostic.Kind.WARNING, String.format(msg, args));
+  }
+
   void logNote(String msg, Object... args) {
     messager.printMessage(Diagnostic.Kind.NOTE, String.format(msg, args));
   }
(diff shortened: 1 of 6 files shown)


spinachomes:
same error #3539，first build(command+ F9) build normal,but second build will throw FilerException,after upgrade to 15.8.2, FilerException is resovled,but get new error:java: java.lang.NullPointerException: Cannot invoke "String.length()" because "s" is null

rbygrave:
Sorry, can you be a lot more explicit about what is going on so that it's really really obvious to me. You mention 2 JDKs, and 2 exceptions.  Also include the exact compiler output, also state explicitly if you are using Quarkus dev mode.


> but second build will throw FilerException

Include the output / error.


> after upgrade to 15.8.2, FilerException is still exist with jdk 17,

Include the output / error.  We are looking to see if this is now a WARNING and not an ERROR [which is what this change was].

> get new error:java: java.lang.NullPointerException: Cannot invoke "String.length()" because "s" is null with jdk 21,

Include all the output and be explicit as to whether this is related to Quarkus dev mode or not.


Thanks.

spinachomes:
ok,I test agian,I am using springboot without quarkus and spring boot devtools，and disable jrebel pluign，idea: 2024.3.2.2
1、use jdk21 or jdk17 and ebean 15.8.1,after mvn clean,first time build run normal,second time build appear error:
image
2、use jdk21 or jdk17 and ebean 15.8.2,after mvn clean,first time build run normal,second time build appear error:
image
This test shows that the performance of different JDKs is the same.

spinachomes:
same actions with idea 2024.2.5 run normal

spinachomes:
oh, mark target/generated-sources/annotations as source root can resolve the problom,now everything is ok

spinachomes:
I resolved this problom, the generated source root dir is wrong after upgrade latest idea,I remark the target/generated-sources/annotations as generated source root resolved the problom

PR #3550, opened 2025-01-29
Pass create/configOptions to DB2 test container

rPraml:
This allows to pass collation/pagesize settings via the `ebean.test.createOptions=` property to the container

This is a quick workaround, so that I can set up DB2 containers with certain pagesize.
Please see also https://github.com/ebean-orm/ebean-test-containers/issues/116, if this is right way for long term?

### DIFF
--- a/ebean-test/src/main/java/io/ebean/test/config/platform/Config.java
+++ b/ebean-test/src/main/java/io/ebean/test/config/platform/Config.java
@@ -19,7 +19,7 @@ class Config {
   /**
    * Common optional docker parameters that we just transfer to docker properties.
    */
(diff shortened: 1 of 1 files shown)


rbygrave (approved):

rbygrave:
Yes, happy with this PR, we can merge it in.

> if this is right way for long term?

So I've got a couple of higher priority things to sort out first before coming back to that one.

_Background note:_ 
ebean-test is cool and useful but for a recent project that only used 1 database [Postgres] the testing setup/configuration was all programmatic. So it used ebean-test-containers and all programmatic test setup and didn't actually use ebean-test.

That is, imo there are some interesting pros & cons with using ebean-test and "properties" stringy based test setup. For example, you had to dive into the code there to figure this out, and that sort of thing is tougher for the "average developer". So I've got background thoughts going on about that - basically ebean-test needs great documentation and full alignment to the underlying functionality in ebean-test-containers ... but even then it is that one extra layer of indirection and that has downsides. 

Need to come back to this in more depth later. Maybe its just a matter of merging this PR and then we are all good ...

PR #3565, opened 2025-02-13
#3564 Fix for nested transaction with batch-mode does not flush

rbygrave:
- Adds flush() into ScopeTrans.commitTransaction()
- Related fixes in DefaultServer execute() and executeCall() where it was performing extra unnecessary nesting on PersistenceException

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,11 +60,6 @@ Work at the highest level of abstraction and drop down levels as needed.
         </a>
       </td>
       <td align="center" valign="middle">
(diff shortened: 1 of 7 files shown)


rbygrave:
Are you happy with this PR @jonasPoehler ?

jonasPoehler:
Yes, this looks good to me! Thank you for cleaning this up and the extensive explanation in #3564. We can just close my original PR and merge this instead.

PR #3570, opened 2025-02-21
Unmodifiable - Improved read only immutable entity beans / graphs

rbygrave:
## Removes query.setReadOnly() 
query.setReadOnly() is completely removed. Code need to migrate explicitly to query.setUnmodifiable() [which is tighter / more restrictive / truely immutable etc] or migrate to "normal" mutable beans.

### Unmodifiable entities are:
- immutable, will throw on attempts to mutate via setters
- disabled lazy loading [don't mutate via lazy loading]
- can be partially populated [with some properties unloaded]
- throw on attempts to read unloaded properties including unloaded collections
- collections "freeze" to unmodifiable collections
- mutable types (e.g. DbArray) also "freeze" to unmodifiable _if supported_
- Uses PersistenceContextScope.QUERY [must do so in order to avoid connecting to modifiable entities in PC]

### Changes
- Adds new LazyInitializationException and UnmodifiableEntityException
- Changes from using IllegalStateException to UnloadedPropertyException, UnmodifiableEntityException, or UnsupportedOperationException
- A query that is read only + disabled lazy loading "upgrades" to unmodifiable

### Ugh
"Read Only" and "Unmodifiable" should probably be exactly the same thing/concept but read only existed before and so it slightly different [it does not include/imply disabled lazy loading + PersistenceContextScope.QUERY ]. 

Ultimately it feels like the best approach is to explicitly remove query.setReadOnly() altogether because Unmodifiable is more restrictive. Code that currently uses setReadOnly() needs to explicitly opt-in to being unmodifiable [no lazy loading].


rbygrave:
FYI @nedge

ebean-api/src/main/java/io/ebean/bean/InterceptReadOnly.java
```diff
@@ -1,39 +1,35 @@
...
  * For the read only use this intercept doesn't need to hold any state that is normally
- * required for updates such as per property changed, loaded, dirty state, original values
+ * required for updates such as per property changed, dirty state, original values
  * bean state etc.
  */
-public class InterceptReadOnly implements EntityBeanIntercept {
+public final class InterceptReadOnly extends InterceptBase {
 
-  private final EntityBean owner;
+  private final boolean[] flags;
```
rPraml, line 18: Maybe rename to "loaded" or "loadedFlags" .

For beans with <=64 properties, we even may introduce an memory optimizes intercept, that packs the loaded flags into a 64bit long value

ebean-api/src/main/java/io/ebean/bean/InterceptReadOnly.java
```diff
@@ -416,12 +414,14 @@ public void preGetId() {
...
-
+    if (!flags[propertyIndex]) {
+      throw new UnloadedPropertyException("Property not loaded: " + property(propertyIndex));
+    }
   }
 
   @Override
   public void preSetterMany(boolean interceptField, int propertyIndex, Object oldValue, Object newValue) {
-
+    throw new UnmodifiableEntityException();
```
rPraml, line 424: Should we add the property name (and maybe also the value) to the exception message?

rPraml (commented):
Hello Rob, I took a short look on this PR. 
I'm looking forward to this change.

We've always used a cumbersome combination of persistence-scope-query + readonly + cache etc.

I have to say that we also use the BeanCache.

I haven't looked at your PR in detail about what happens if data is partially present in the BeanCache and a partial bean is created and the missing data then has to be reloaded using lazy load

ebean-api/src/main/java/io/ebean/bean/InterceptReadOnly.java
nedge: Adding the property name makes sense - then it would be easier to tune the query or fix what is trying to access the non-loaded property

ebean-api/src/main/java/io/ebean/bean/InterceptReadOnly.java
rbygrave: Renamed to "loaded", that will do for now.

ebean-api/src/main/java/io/ebean/bean/InterceptReadOnly.java
rbygrave: Added the property name. We will not include the value as that can be a security issue.

rPraml:
> "Read Only" and "Unmodifiable" should probably be exactly the same thing/concept but read only existed before and so it slightly different [it does not include/imply disabled lazy loading + PersistenceContextScope.QUERY ]. Ideally we change query.setReadOnly(true) to be _exactly_ the same as query.setUnmodifiable(true).

I would agree that. (if really necessary, we couldbadd a config flag for backward compatibility)

But I think it still needs to be specified what should happen when caches are involved, as they already check the readOnly flag. 

**QueryCache:**

Beans/IdLists in the queryCache should be always unmodifiable by default. (we use setUseQuerCache=true always together with setReadonly=true)
I can imagine only very few use cases, where this is not necessary (we have exacly one) 

It is also necessary to set the persistence-scope to QUERY when beans are put into the queryCache, otherwise we got memory leaks.

So I have to pay attention always to use all 3 properties together correctly in my code. 

Therefore, I would appreciate it if setUseQuerCache(true) would set everything correctly (or maybe an other convenient method like `withQueryCaching()`) 


**BeanCache**

Imagine if I do the query `DB.find(...).select("prop1").setId(42).setUseBeanCache(true).findOne()`

This will put partial data to the bean cache. 

Next I do a `select("*").setImmutable(true)` and will have a beanCache hit. 

What will we expect? 
- got a partial loaded bean that throws a LazyLoadException on access to an unloaded property (that would be bad) 
- or can a bean from the cache do lazy loads anyway
- or has the cache to validate, if there are all properties present and treat this situation as cache miss

**mutable types**

There are mutable types like Calendar or `@DbJson` properties. The later we use a lot. (you may remember JsonMutationDetection) 

How do we handle this and protect from modification? 

I think there is not much we can do about this, which would be totally OK for me, if it is specified. 
(But I could imagine that custom json objects also can implement the `freeze` method later) 

Roland

ebean-api/src/main/java/io/ebean/UnloadedPropertyException.java
```diff
@@ -0,0 +1,18 @@
+package io.ebean;
+
+/**
+ * Thrown when trying to access a property that isn't loaded on a read only entity.
+ * <p>
+ * On a normal mutable entity accessing the property would invoke lazy loading. On
+ * a read only entity with lazy loading disabled, accessing an unloaded property
+ * throws this UnloadedPropertyException instead.
+ */
+public class UnloadedPropertyException extends UnmodifiableEntityException {
```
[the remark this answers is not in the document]
rbygrave: > CHECKME: should it extend from UnmodifiableEntityException or better from UnsupportedOperationException?

It already does in that ...

UnmodifiableEntityException extends UnsupportedOperationException

So:

UnloadedPropertyException extends UnmodifiableEntityException extends UnsupportedOperationException


> Maybe LazyInitialisationException could even be a better name

Yes it might. We could consider renaming UnloadedPropertyException to LazyInitialisationException.

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultOrmQueryEngine.java
```diff
@@ -129,22 +129,21 @@ public <T> BeanCollection<T> findMany(OrmQueryRequest<T> request) {
...
       request.descriptor().cacheBeanPutAll(result.actualDetails());
     }
 
     request.mergeCacheHits(result);
-    if (request.isQueryCachePut()) {
-      // load the query result into the query cache
-      result.setReadOnly(true);
-      request.putToQueryCache(result);
-      if (Boolean.FALSE.equals(query.isReadOnly())) {
-        result = result.shallowCopy();
```
rbygrave, line 144: **_Breaking change:_** No longer support readOnly = FALSE ... producing a shallow copy.

Applications using query qache + readOnly False ... were getting back a shallow copy that was mutable. This now throws an IllegalStateException.

rbygrave:
My current plan is to release this as a series of Release Candidate releases. I think we can release the first RC release from this change.

- There are some behaviour changes, do we protect well enough against those changes?
- query.setReadOnly(false) was previously able to be used [to return shallow copies of results from query cache that are mutable] but currently I think it makes sense to make it invalid and query cache always returns fully immutable results?
- Should we completely remove query.setReadOnly() from the API, to indicate that this is where the behaviour changes are?
- Using unmodifiable(true) with bean caching and partially loaded beans in the cache looks like an issue. How best to deal with it? Maybe unmodifiable(true) can't be used for bean cache partials or bean caching at all initially?
- This change has got bigger than I thought it would be

rbygrave:
Right now I am thinking that we should effectively remove the old ReadOnly specific behaviours and have it fully replaced by Unmodifiable.

This would mean we should **_bump the major version_**, and I believe the best approach is to **_remove query.setReadOnly() from the API_**. This then forces people to review the application use of ReadOnly and explicitly opt in to Unmodifiable which has more restrictions (no lazy loading, fully immutable - so no readOnly(false) support).

That is, imo the problem with keeping ReadOnly is that the differences are too subtle and we can have runtime behaviour changes and I think that will be a bad experience - we are better off breaking API and people explicitly deciding to go looser [normal mutable etc] or tighter [fully immutable]

That is, people upgrading will be forced to remove use of query.setReadOnly() and make the decision of changing to just normal mutable lazy loading etc ... or explicitly opt in for the more strict world of query.setUnmodifiable(true).

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultOrmQueryEngine.java
rPraml: Returning an immutable graph here, would be OK here. We have exactly one (theoretical) use case, where we modify a bean in the cache. We can convert the bean to a modifiable one manually - or use a stateless update...

But:
```java
 if (request.query().isUnmodifiable()) {
      finalResult = result == null ? null : result.freeze();
      if (request.isQueryCachePut()) {
        // load the query result into the query cache
        request.putToQueryCache(finalResult);
      }
    }
```
does this mean, that if I set queryCache to true, I have also set unmodifiable to true, otherwise the cache will not work properly?
I would prefer if a `query.setQueryCache(true)` automatically sets the unmodifiable flag (maybe this is done somewhere and I haven't discovered it yet)

ebean-test/src/test/java/org/tests/basic/TestLoadBeanCache.java
```diff
@@ -35,14 +35,15 @@ void testLoad() {
     Map<String, Country> map = DB.find(Country.class)
       .setBeanCacheMode(CacheMode.PUT)
       .setUseQueryCache(true)
-      .setReadOnly(true)
+      //.setReadOnly(true)
+      .setUnmodifiable(true)
```
rPraml, line 39: This shouldn't be neccessary (see other comment) `setUseQueryCache(true)` should be enough.

ebean-test/src/test/java/org/tests/basic/TestQueryWithCache.java
```diff
@@ -50,15 +51,13 @@ public void testCountryDeploy() {
     Country nz4 = DB.find(Country.class).setId("NZ").setAutoTune(false).setUseCache(false)
       .findOne();
 
-    assertTrue(nz2 == nz2b);
-    assertTrue(nz2 == nz3);
-    assertTrue(nz3 != nz4);
-
+    assertThat(nz2).isNotSameAs(nz2b); // Changed behaviour with unmodifiable
```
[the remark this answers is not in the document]
rPraml: > Reference beans of @Cache entity beans that don't have ANY relationships

JFYI: This case practically never occurs in our code. We have a reference (whoCreated/whoModified) to a User object in almost every entity.

rPraml (commented):
>    query.setReadOnly(false) was previously able to be used [to return shallow copies of results from query cache that are mutable] but currently I think it makes sense to make it invalid and query cache always returns fully immutable results?

yes, and the result should also be persistenceScope=QUERY - otherwise the PC gets into the cache and may cause memory leaks.

>    Should we completely remove query.setReadOnly() from the API, to indicate that this is where the behaviour changes are?

This means a breaking API. But It would not be a big problem for us to review the changes where we used readonly (Strictly speaking, we had the rule to set readOnly and persistenceScope=QUERY in all queries with queryCache - this breaking change would make our code more readable)

>    Using unmodifiable(true) with bean caching and partially loaded beans in the cache looks like an issue. How best to deal with it? Maybe unmodifiable(true) can't be used for bean cache partials or bean caching at all initially?

That would be a showstopper for us. We have some beans where we keep a few properties in the cache, but then reload lobs via lazy loading.
We might also need to consider database systems like DB2, which, for example, can't load lobs with distinct queries. These wouldn't be properly cached.

>    This change has got bigger than I thought it would be

and my comments make them even bigger ;)

Can't we make it so that, in cached beans, calling the setter throws an Unmodifiable exception, but lazyLoad is still possible?
I might have to think about that again...

nedge:
I am in favour of removing the query.setReadOnly as it wasn't consistent anyway around lazy loading (not throwing an Exception but instead returning null when lazy loading was invoked)
I think we need to keep in mind what the typical use case is for querying/using unmodifiable objects. In our app we have a load of configuration data that is very seldom changed. These objects can are used in so many REST get calls so caching them as unmodifiable objects should give a huge performance boost and reduce GC.  
Generally, most apps will load Beans by ID when updating (probably also using so lazy loading) so you never really have a mixture of modifiable and unmodifiable in the same calls although we may want to set references to unmodifiable objects which is fine.

ebean-core/src/main/java/io/ebeaninternal/server/query/DefaultOrmQueryEngine.java
rbygrave: > query.setQueryCache(true) automatically sets the unmodifiable flag

That is how its done in DefaultOrmQuery

```java
  @Override
  public final Query<T> setUseQueryCache(CacheMode useQueryCache) {
    this.useQueryCache = useQueryCache;
    if (CacheMode.OFF != useQueryCache) {
      unmodifiable = true;
    }
    return this;
  }
```

rbygrave:
FYI: There is a followup PR that removes readOnly from Query, BeanCollection, EntityBeanIntercept and related places - https://github.com/ebean-orm/ebean/pull/3592

Currently I think these 2 PRs would go together in a major version bump.

ebean-api/src/main/java/io/ebean/LazyInitialisationException.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Thrown when trying to access a property that isn't loaded on an entity
+ * that is unmodifiable or has disabled lazy loading.
+ * <p>
+ * On a normal mutable entity accessing the property would invoke lazy loading. On
+ * a unmodifiable entity with lazy loading disabled, accessing an unloaded property
+ * throws this LazyInitialisationException instead.
+ */
+public class LazyInitialisationException extends UnmodifiableEntityException {
```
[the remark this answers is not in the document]
rbygrave: > This means a breaking API. ... not be a big problem for us ... would make our code more readable

Very good. I think everything is pointing towards that approach of removing setReadOnly().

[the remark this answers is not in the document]
rbygrave: > That would be a showstopper for us. We have some beans where we keep a few properties in the cache, but then reload lobs via lazy loading.

Can you clarify why it would be a showstopper? As in, I presume those are currently query.setReadOnly(true) that are hitting the bean cache ... and we'd just not be able to make them query.setUnmodifiable(true) because we are actually relying on lazy loading for that use case [and unmodifiable beans don't support lazy loading].

So why is that a show stopper? What is the issue with using normal modifiable beans for this case?

ebean-test/src/test/java/org/tests/basic/TestLoadBeanCache.java
rbygrave: > We have exactly one (theoretical) use case, where we modify a bean [that has come out of the query cache]

The query cache results are always unmodifiable. So say we have a use case where we've got an unmodifiable bean (from either a query cache result or just a query.setUnmodifiable()) and we need to perform some mutation on it and perform an update. How can we do it? 

Options:
- 1: Manually create a mutable copy of the bean, mutate, then database.update() it
- 2: Add some **_new API_** that creates a mutable copy of an entity bean like `<T> T database.mutableCopy(T source);`

Interestingly enough, a mutable copy instance could actually allow lazy loading (makes me think of the lazy loading lob case mentioned above). Hmmm.

PR #3578, opened 2025-02-26
Work around `mysqldump` bug by moving comment one line up

### LINKED ISSUE #3577: mysqldump backups broken with #3408
@rbygrave @mkurz The change in #3408 breaks backups, that are made with "mysqldump" / "mariadbdump"
(We use `mysqldump --hex-blob --routines --single-transaction --skip-extended-insert --complete-insert --no-autocommit`)

When backing up a DB, "mysqldump" creates the following output:
```sql
DELIMITER ;;
CREATE DEFINER=`user`@`%` PROCEDURE `usp_ebean_drop_column`(IN p_table_name VARCHAR(255), IN p_column_name VARCHAR(255))
BEGIN
CALL usp_ebean_drop_foreign_keys(p_table_name, p_column_name);
SET @sql = CONCAT('ALTER TABLE `', p_table_name, '` DROP COLUMN `', p_column_name, '`');
PREPARE stmt FROM @sql;
EXECUTE stmt;
END
-- play-ebean-end ;;
DELIMITER ;
```

Before, the penultimate line was `END;;`. Now it is `-- play-ebean-end ;;` and the backup cannot be restored, because the delimiter is behind the comment. You may have to replace `-- play-ebean-end ;;` by `;;` in these backups.

I think this is a bug in mysqldump, but I still wanted to mention that we have a problem here now.

an easy workaround could be, that after `-- play-ebean-end` a newline is added or that it is ignored by the ebean migration-runner - but we have not verified that, yet.

mkurz:
It seems like `mysqldump` just adds `;;` to the last line but unfortunately does not check if that line is a comment. To work around that it's good enough to move the comment one line up - as long as the END does not end with ; (which here it does not so that is ok)

- Fixes #3577

### DIFF
--- a/ebean-ddl-generator/src/main/resources/io/ebeaninternal/dbmigration/builtin-extra-ddl.xml
+++ b/ebean-ddl-generator/src/main/resources/io/ebeaninternal/dbmigration/builtin-extra-ddl.xml
@@ -155,8 +155,8 @@ EXECUTE stmt;
 END LOOP;
 
 CLOSE curs;
-END
 -- play-ebean-end
+END
 $$
 
 DROP PROCEDURE IF EXISTS usp_ebean_drop_column;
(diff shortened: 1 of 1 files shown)


rPraml:
I don't know much about the play-framework, so if the `play-ebean-start` and `play-ebean-end` will do their job correctly, I'm fine with this change.

As far as I understand #3408 you will try to switch off statement splitting between the two comments. Just for my interest: It would be nice, if you explain in 2-3 sentences, how this splitting is avoided. (or is it private code?)

rbygrave (approved):

Issue #3573, opened 2025-02-25
OneToMany does not support initialisation via List.of() - Was - Fetching multiple ToMany eagerly doesn't work on transitive associations


apflieger:
Underlying issue here:

```java
@OneToMany
private List<Attendance> attendances = List.of(); // we can't use List.of() 
````

A collection initialised with `List.of()` isn't supported. It should be left uninitialised or initialised as `new ArrayList<>()` [which are treated exactly the same, ebean enhancement removes that initialisation code as it must control the initialisation of collections to support orphanRemoval and ManyToMany adds/removes]
 

----------------

Hi,

Thanks a lot for your work, that's always worth saying.

## Context

I am trying to load an entity that has multiple ToMany relations, but transitively, in a single EBean query. The model look like this:
```java
class Payment {
    @ManyToOne
    Ticket ticket;
}

class Ticket {
    @OneToMany
   List<Payment> payments;

    @OneToMany
   List<Attendance> attendances;
}
```
I am fetching Payments entities with a FetchGroup that queries the two paths `ticket.payments` and `ticket.attendances`. 

## Expected behavior
I expect to have one query with a join for the first relation, and a secondary query for the second. 

## Actual behavior
What I see is that secondary queries are not executed, everything is then lazy loaded one row at a time.

I didn't setup a reproduction case, let me know if that's necessary. But I spent some time to debug and I landed [here](https://github.com/ebean-orm/ebean/blob/40637bd68ccea867525a857756355fd6021e2dd2/ebean-core/src/main/java/io/ebeaninternal/server/loadcontext/DLoadManyContext.java#L96C13-L96C22)
[...]

PR #3581, opened 2025-02-28
OneToMany initialisation via List.of() - was Tests only - new test on multiple many joins

apflieger:
This test reproduce the problem of #3573
The field initialization `= List.of();` prevents eager secondary query.

### DIFF
--- a/ebean-test/pom.xml
+++ b/ebean-test/pom.xml
@@ -309,7 +309,7 @@
         <extensions>true</extensions>
         <configuration>
           <tiles>
-            <tile>io.ebean.tile:enhancement:14.9.0</tile>
+            <tile>io.ebean.tile:enhancement:14.10.0</tile>
           </tiles>
         </configuration>
       </plugin>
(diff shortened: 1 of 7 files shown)


rbygrave:
While this test is Ok'ish in that it shows the issue with `List.of()`, it does actually rely **_incorrectly_** on **_field access_** ... which you can't do unless you explicitly enable a special enhancement flag `entity-field-access`.

That is the assert:

```java
    assertThat(list2.get(0).orderInvoices).hasSize(2);
``` 
... is relying on the _GETFIELD_ call for `orderInvoices`, and if that changes to an expected method call like `getOrderInvoices()` then the test fails to reproduce an error (because lazy loading is invoked and replaces the empty collection with a populated one via lazy loading).

So that means I'm concerned that your application code is incorrectly using field access? [Or why did the test case get written using field access? Did you just get lucky?]

apflieger:
Heh, yes I was aware of that. Our production code doesn't access fields directly. I was trying to reproduce the case in a simple way and I had in mind that this test wouldn't get merged anyway, so I judge it to be good enough for a disposable test.

Issue #3188, opened 2023-08-24
Consistency of count value in background thread according to current transaction


AntoineDuComptoirDesPharmacies:
## Expected behavior

When calling `getTotalRowCount`, the value should be correct in the current transaction scope even if `loadRowCount` have been called before.

## Actual behavior

If we call `loadRowCount` prior to `getTotalRowCount`, the count query is executed outside of current transaction.

This is due to the fact that `DefaultServer.findFutureCount` (https://github.com/ebean-orm/ebean/blob/13.21.0/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java#L1273) ignore the given parameter (which is null) and create a new transaction.

### Steps to reproduce

```java
try (Transaction outerTransaction = DB.beginTransaction()) {
  DB.update()
      .set("status", "enabled")
      .where()
      .eq("reference", reference)
      .update();

  PagedList<Order> pagedList = DB.find(Order.class)
       .where()
       .eq("status", "enabled")
       .setFirstRow(0)
       .setMaxRows(10L)
       .findPagedList();

  // initiate the loading of the total row count   
  // in a background thread
  pagedList.loadRowCount();

  // fetch and return the list in the foreground thread
  List<Order> orders = pagedList.getList();

  LOG.info(orders.size());

  // get the total row count (from the future)
  int totalRowCount = pagedList.getTotalRowCount();

  LOG.info(totalRowCount);

  transaction.commit();
}

try (Transaction outerTransaction = DB.beginTransaction()) {
  DB.update()
      .set("status", "enabled")
      .where()
      .eq("reference", reference)
      .update();
  
  PagedList<Order> pagedList = DB.find(Order.class)
       .where()
       .eq("status", "enabled")
       .setFirstRow(0)
       .setMaxRows(10L)
       .findPagedList();

  // fetch and return the list in current thread
  List<Order> orders = pagedList.getList();

  LOG.info(orders.size());

  // get the total row count (immediate call, without future)
  int totalRowCount = pagedList.getTotalRowCount();

  LOG.info(totalRowCount);

  transaction.commit();
}
```

```console
First Call :
   orders.size() : 1
   totalRowCount : 0

Second Call : 
   orders.size() : 1
   totalRowCount : 1
```

Yours faithfully,
LCDP

PR #3593, opened 2025-03-18
Add feature findFutureMap

thomas-lcdp:
Add feature to query in background as findFutureList

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
@@ -1319,6 +1319,33 @@ public final class DefaultServer implements SpiServer, SpiEbeanServer {
   }
 
   @Override
+  public <K, T> FutureMap<K, T> findFutureMap(SpiQuery<T> query) {
+    SpiQuery<T> spiQuery = query.copy();
+    spiQuery.usingFuture();
+    // FutureMap query always run in it's own persistence content
+    spiQuery.setPersistenceContext(new DefaultPersistenceContext());
+    if (!spiQuery.isDisableReadAudit()) {
+      BeanDescriptor<T> desc = descriptorManager.descriptor(spiQuery.getBeanType());
+      desc.readAuditFutureList(spiQuery);
+    }
+    // Create a new transaction solely to execute the findMap() at some future time
+    boolean createdTransaction = false;
+    SpiTransaction transaction = query.transaction();
+    if (transaction == null) {
+      transaction = currentServerTransaction();
+      if (transaction == null) {
+        transaction = (SpiTransaction) createTransaction();
+        createdTransaction = true;
+      }
+      spiQuery.usingTransaction(transaction);
+    }
+    QueryFutureMap<K, T> queryFuture = new QueryFutureMap<>(new CallableQueryMap<>(this, spiQuery, createdTransaction));
+    backgroundExecutor.execute(queryFuture.futureTask());
+    return queryFuture;
+  }
+
+
+  @Override
   public <T> PagedList<T> findPagedList(SpiQuery<T> query) {
     int maxRows = query.getMaxRows();
     if (maxRows == 0) {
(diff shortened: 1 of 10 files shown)


rbygrave:
Yes, this all looks pretty good.

ebean-core/src/main/java/io/ebeaninternal/server/core/DefaultServer.java
```diff
@@ -1318,6 +1318,33 @@ public <T> FutureList<T> findFutureList(SpiQuery<T> query) {
...
+    SpiQuery<T> spiQuery = query.copy();
+    spiQuery.usingFuture();
+    // FutureMap query always run in it's own persistence content
+    spiQuery.setPersistenceContext(new DefaultPersistenceContext());
+    if (!spiQuery.isDisableReadAudit()) {
+      BeanDescriptor<T> desc = descriptorManager.descriptor(spiQuery.getBeanType());
+      desc.readAuditFutureList(spiQuery);
+    }
+    // Create a new transaction solely to execute the findMap() at some future time
+    boolean createdTransaction = false;
```
rbygrave, line 1332: FYI: I realise this is consistent with the other find futures ... but I'll remove this creating transaction part, the reason is that it prevents us using a "Read Only DataSource" for executing the query. [so this is actually an issue with findFutureList, findFutureIds, findFutureCount ... I'll remove it from all of them]

So removing this part, means that the query will instead get and use a "read only transaction" to run the query.

Just putting this comment here so that its not a surprise when I remove that part.

rbygrave: > Do you think it may create a regression 

I know it does not produce a regression because ... we gotta test for that :)

More specifically, at https://github.com/ebean-orm/ebean/pull/3596/files#diff-0804968369579e408eea483a9e582485cd7faf94e5d78dc8f7a918c729962a70R1261 ... we get the find future query to use a currently active transaction (if there is one).

```java
  private <T> SpiQuery<T> configureForFuture(SpiQuery<T> query) {
    query.usingFuture();
    if (query.transaction() == null) {
      // use a current transaction if active
      query.usingTransaction(currentServerTransaction());   //   <<---- HERE !!! 
    }
    return query;
  }
```


> removed code that will produce "read only transaction".

So specifically we removed the line https://github.com/ebean-orm/ebean/pull/3596/files#diff-0804968369579e408eea483a9e582485cd7faf94e5d78dc8f7a918c729962a70L1266

_removed line:_
```java
 transaction = (SpiTransaction) createTransaction();
```
... and that's because this isn't actually a **_read only_** transaction [that might use the read only DataSource] but instead this is a "normal" transaction and this will always use the main DataSource.

So we remove this and we've now got:
- If the query has as explicit transaction via query.usingTransaction() it will use that transaction
- If there is a current active transaction, it will use this (that was the bug #3188 ) 
- If the future query has Query.usingMaster() it will use the main DataSource 
- If a Read Only DataSource is configured, it will use that

rbygrave: Side note: I improved the docs for Read Replicas just yesterday - https://ebean.io/docs/read-replicas/

rbygrave: > indexes from write replica are not forwarded to read replica

There is "Physical replication" [driven of a WAL / transaction log] and then there is "Logical replication" [of selected tables / columns].

Generally a "Read Replica" means its using "Physical replication" [or equivalent] and so all indexes etc are all present on the replica. [AWS Aurora uses a shared networked storage but again all indexes etc are present on the replica].


>  any feature in EbeanDynamicEvolution that permit having specific DB Evolution files applied to the read replica ?

No, and I don't see there being a need there.


However, with logical replication we can see similar but different models (some columns but not all, extra columns, different indexes etc) - but then it would have its own distinct DDL / migrations.

AntoineDuComptoirDesPharmacies: Just checked out our replica and you are right, indexes are "propagated", nice !
We checked the replica lag metric and we are around 20ms of max replica lag, which is quite not bad.

Perfect, thanks for these insights !

PR #3594, opened 2025-03-18
Modify disableLazyLoad to throw LazyInitialisationException instead o…

rbygrave:
…f returning null

Modify such that accessing an unloaded property throws a LazyInitialisationException. This includes ToMany collections that where previously initialised as empty collections (that didn't lazy load).

This change brings the disableLazyLoad behaviour in line with the unmodifiable behaviour.

It does mean that the use case of bulk mapping from entities to DTOs via something like MapStruct will no longer work nicely (mapping nulls and empty lists versus LazyInitialisationException).

### DIFF
--- a/ebean-test/src/test/java/org/tests/batchload/TestQueryDisableLazyLoad.java
+++ b/ebean-test/src/test/java/org/tests/batchload/TestQueryDisableLazyLoad.java
@@ -1,18 +1,18 @@
 package org.tests.batchload;
 
+import io.ebean.LazyInitialisationException;
 import io.ebean.xtest.BaseTestCase;
 import io.ebean.DB;
 import io.ebean.test.LoggedSql;
 import org.junit.jupiter.api.Test;
 import org.tests.model.basic.Order;
-import org.tests.model.basic.OrderDetail;
 import org.tests.model.basic.ResetBasicData;
 
+import java.sql.Date;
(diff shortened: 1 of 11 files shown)


ebean-test/src/test/java/org/tests/batchload/TestQueryDisableLazyLoad.java
```diff
@@ -83,7 +82,29 @@ public void onAssocOne_when_partial() {
...
+      .select("status, orderDate")
+      .orderBy().asc("id")
+      .findList();
+
+    assertThat(l0).isNotEmpty();
+
+    Order order = l0.get(0);
+
+    // normally invokes lazy loading
+    assertThrows(LazyInitialisationException.class, () -> order.setShipDate(new Date(System.currentTimeMillis())));
```
rPraml, line 107: I'm unsure, if this should fail.., but I understand your concern, that this would break changelog etc.
I think I have no use case, so lets's also fail on setter

rPraml (commented):
I think this change will hurt people, that are using lazy load and their code will fail now.
But you can probably say in many cases (probably in all) that they used ebean wrong and also wrote wrong code.

BTW: I like exceptions thrown as early as possible, like here. What good is an NPE if the error already occurred 10 method calls earlier... 👍

PR #3597, opened 2025-03-20
#3515 Bug in QueryCache invalidation 

rbygrave:
Effectively the bug fix is in SqlTreeNodeRoot.dependentTables() to include the baseTable in the set of dependent tables

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeRoot.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeRoot.java
@@ -76,6 +76,7 @@ final class SqlTreeNodeRoot extends SqlTreeNodeBean {
 
   @Override
   public void dependentTables(Set<String> tables) {
+    tables.add(baseTable);
     for (SqlTreeNode child : children) {
       child.dependentTables(tables);
     }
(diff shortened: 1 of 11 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/query/SqlTreeNodeRoot.java
```diff
@@ -76,6 +76,7 @@ public SqlJoinType appendFromBaseTable(DbSqlContext ctx, SqlJoinType joinType) {
 
   @Override
   public void dependentTables(Set<String> tables) {
+    tables.add(baseTable);
```
rbygrave, line 79: This is effectively the bug fix here. We need the baseTable to be included in the set of dependent tables

ebean-test/src/test/java/org/tests/cache/TestQueryCacheConcurrent.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestQueryCacheConcurrent extends BaseTestCase {
+
+  private volatile boolean running;
+  private volatile boolean failed;
+  private volatile int  step;
+
+  @ForPlatform(Platform.POSTGRES) // H2 isn't good enough for this test
```
rbygrave, line 20: This test is not reliable on H2

Issue #3625, opened 2025-04-22
filterMany uses id property name instead of id column name on idNullOr


goncalo-andrade:
Hello,

I believe I found a bug, happening since version 14.5.2/15.5.2. On a bean where the id property name differs from the corresponding column, the `id is null or` part of the query generated by `filterMany` uses the property name, instead of the column name, leading to errors if no column exists with that name, or inconsistent behaviour if it does. 

I'm happy to help on reproducing or fixing this bug.

rbygrave:
Can you provide actual code that reproduces the issue, and the resulting
sql that is produced. Thanks.

goncalo-andrade:
Hi! Thank you for the quick response. Sorry for not providing code right away. My classes look something like this:

```java
// ParentEntity.java
import jakarta.persistence.CascadeType;
import jakarta.persistence.Entity;
import jakarta.persistence.Id;
import jakarta.persistence.OneToMany;
import java.util.List;

@Entity
public class ParentEntity {

    @OneToMany(mappedBy = "parentEntity", cascade = CascadeType.ALL)
    private final List<ChildEntity> children;

    @Id
    private long id;

    public ParentEntity(List<ChildEntity> children) {
        this.children = children;
    }

    public long getId() {
        return id;
    }

    public List<ChildEntity> getChildren() {
        return children;
    }
}

// ChildEntity.java
import io.ebean.annotation.Index;
import jakarta.persistence.Column;
import jakarta.persistence.Entity;
import jakarta.persistence.GeneratedValue;
import jakarta.persistence.Id;
import jakarta.persistence.ManyToOne;
import jakarta.validation.constraints.NotBlank;
import jakarta.validation.constraints.NotNull;
import jakarta.validation.constraints.Size;

@Entity
public class ChildEntity {

    @Id
    @GeneratedValue
    @NotNull
    @Column(name = "id")
    private long idPrimaryKey;

    @Index
    @NotNull
    @NotBlank
    @Column(unique = true, length = 64)
    @Size(max = 64, min = 64)
    private String someOtherId;

    @ManyToOne(optional = false)
    private ParentEntity parentEntity;

    public ChildEntity(String someOtherId) {
        this.someOtherId = someOtherId;
    }

    public String getSomeOtherId() {
        return someOtherId;
    }

    public long getIdPrimaryKey() {
        return idPrimaryKey;
    }

    public ParentEntity getParentEntity() {
        return parentEntity;
    }
}

// Repository.java
[...]

PR #3626, opened 2025-04-22
Fix incorrect usage of bean property name instead of corresponding column

goncalo-andrade:
This is a possible fix for issue #3625. It fixes the issue for my particular context, but it might not be a general fix.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/deploy/id/IdBinderSimple.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/deploy/id/IdBinderSimple.java
@@ -264,6 +264,6 @@ public final class IdBinderSimple implements IdBinder {
 
   @Override
   public String idNullOr(String prefix, String filterManyExpression) {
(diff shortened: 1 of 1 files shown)


rbygrave:
Thanks, excellent.  Note I've added some tests in a followup commit.

PromanSEW:
This commit breaks my code, nothing helps
```console
Caused by: java.lang.UnsupportedOperationException: null
	at io.ebeaninternal.server.el.ElPropertyChain.idNullOr(ElPropertyChain.java:291)
	at io.ebeaninternal.server.query.CQueryPredicates.prepare(CQueryPredicates.java:194)
```
Was: `fetch("clan", "buildings")`
Tried `fetch("clan.buildings", "id, type, level"`, `fetch("clan.buildings", (String) null)`
Where:
```java
// ClanQuest.java (parent)
@Basic(fetch = FetchType.LAZY)
@ManyToOne(optional = false)
public Clan clan;
```
```java
// Clan.java
@OneToMany(cascade = CascadeType.ALL)
public List<Building> buildings = new ArrayList<>();
```

Where in building class:
```java
//Building.java
@Id
private int id;
```
Please help

PR #3634, opened 2025-05-14
[native-image] Modify querybean-generator to include AttributeConverters in generated reflect-config.json

rbygrave:


rbygrave:
Include this in RC3

Issue #3643, opened 2025-06-25
Simplify the select id of the *ToOne properties


nPraml:
Hello @rbygrave ,

We have a feature in our application that exports all data to JSON files so that it can be imported into other applications (such a backup & restore feature).
For this, we always need the linked object relations (e.g., *ToOnes).
We use `PathProperties` for this, which precisely define what needs to be loaded for each entity during export.

We always create a dependency graph that shows the order in which the objects need to be exported and imported for it to work.
For example: `Customer` is exported before `Order`, and for `Order`, it's sufficient if we only export the Customer ID; this way, the relations remain intact.

Some of our objects are very large, and MariaDB throws an exception stating that only 61 tables can be used in the query:

```
Query threw SQLException:(conn=14) Too many tables; MariaDB can only use 61 tables in a join
```

For this example, we wrote the following test:

```java
 @Test
  void test_withToOne() {
    ResetBasicData.reset();

    PathProperties root = PathProperties.parse("*,customer(id)");
    LoggedSql.start();
    Query<Order> query = DB.find(Order.class).apply(root);
    query.findList();
    List<String> sql = LoggedSql.stop();
    assertThat(sql).hasSize(1);
    // test gives this sql:
    // select t0.id, t0.status, t0.order_date, t0.ship_date, t1.name, t0.cretime, t0.updtime, t1.id from o_order t0 join o_customer t1 on t1.id = t0.kcustomer_id
    // we would assume this:
    assertThat(sql.get(0)).contains("select t0.id, t0.status, t0.order_date, t0.ship_date, t1.name, t0.cretime, t0.updtime, t0.kcustomer_id from o_order t0");
  }
```

We discovered that some IDs trigger an unnecessary join (`from o_order t0 join o_customer t1 on t1.id = t0.kcustomer_id`), and then only the ID is finally selected from `t1`.
In our opinion, the customer_id could simply come from the "main" table, and the join would be unnecessary (`select ..., t0.kcustomer_id`.
We analyzed how we could reduce our query from the current 109 tables to 40-50 tables.

Can you please give us feedback on whether this simplification of the joins would make sense in your opinion?
What kind of SQL would be expected in the test case?

@rPraml , @jonasPoehler FYI

rbygrave:
Can you repeat the example without using PathProperties and instead just use a query?

[...]

PR #3644, opened 2025-06-26
Test fetch ID only

nPraml:
Hello @rbygrave ,

I wrote the tests for issue https://github.com/ebean-orm/ebean/issues/3643. 
In all three cases, I would expect the same SQL query to be output: `select t0.id, t0.status, t0.kcustomer_id from o_order t0`


nPraml:
Hi @rbygrave ,

We prefer this solution over https://github.com/ebean-orm/ebean/pull/3646.
We've adopted this in our fork and successfully tested it in our application.
Could you please take a look?

rbygrave:
Copied / rebased to https://github.com/ebean-orm/ebean/pull/3858

PR #3645, opened 2025-06-27
Query cache oom for too large beans

Wosch96:
Hi @rob-bygrave,

we have a problem with the query cache that holds whole beans inside the memory. If these beans are too big, a potential OOM can occur.
I created a test that uses a 'big' bean in an '.eq' expression without the specific id call.

The easy solution would be the usage of '.id'
`.eq("child.id", DB.reference(MemleakChild.class, id).id)`.

I added a potential draft to fix this. What default would you expect in this case? Should I always use .id or should this be handled automatically if possible. Let me hear what you think about this.


rbygrave:
> or should this be handled automatically if possible

Well if it's easy enough for people to "get the wrong result" ... then we should address it if we can.  So yes this looks like the good and right thing to do.

I'll have another look to see if there are any side effects or if the PR is good to go.

PR #3646, opened 2025-07-02
Draft: Fetch id only - v1

nPraml:
Hello @rbygrave ,

I've been trying to find a solution to the problem in https://github.com/ebean-orm/ebean/pull/3644.

I see two ways to approach the problem:
- either recognize when calling fetch that you only need the ID of the OneProperties and convert it immediately to a select (in `OrmQueryDetail` in the select clause instead of fetchPath)
- or recognize later, when assembling the `SqlTree`, when a join would be unnecessary.

I've programmed a sketch of the initial idea, which only shows roughly what I envision for the solution. I only provisionally built the extension of the selects in `OrmQueryDetail` and `OrmQueryProperties` because we would then need such a mechanism; it's probably not 100% correctly implemented.

I've disabled one test, which I can look at later. In the subsequent tests, the queries have become simpler.

Can you please take a look at this PR to see if it's moving in the right direction? Or should we pursue the second idea that the simplification only happens later in the `SqlTree`?

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
@@ -1447,6 +1447,21 @@ public class DefaultOrmQuery<T> extends AbstractQuery implements SpiQuery<T> {
   }
 
   private Query<T> fetchInternal(String path, String properties, FetchConfig config) {
+    // look at all One Props, see if we can select only the ID of one?
+    for (BeanPropertyAssocOne<?> beanPropertyAssocOne : beanDescriptor.propertiesOne()) {
+      if(beanPropertyAssocOne.name().equals(path)) {
+        // only if we select the ID do we prevent the join, otherwise we need something from the other table anyway...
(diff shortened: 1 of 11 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/querydefn/DefaultOrmQuery.java
```diff
@@ -1447,6 +1447,21 @@ public final Query<T> fetchLazy(String path, String properties) {
   }
 
   private Query<T> fetchInternal(String path, String properties, FetchConfig config) {
+    // look at all One Props, see if we can select only the ID of one?
+    for (BeanPropertyAssocOne<?> beanPropertyAssocOne : beanDescriptor.propertiesOne()) {
+      if(beanPropertyAssocOne.name().equals(path)) {
+        // only if we select the ID do we prevent the join, otherwise we need something from the other table anyway...
+        if(beanPropertyAssocOne.targetIdProperty().equals(properties)) {
```
nPraml, line 1454: this if branch does not work properly yet, you can also fetch other OnProperty, not just the id

rbygrave:
Replaced by https://github.com/ebean-orm/ebean/pull/3858

PR #3656, opened 2025-07-17
Non-working modify-tracking in certain cases

rPraml:
We found an issue, when O2M relations are not correctly persisted to the DB.

This happens, when
- a bean is saved and the O2M property is empty
- something is added and cleared again in two subsequent saves.
- the same master-bean object has to be used

The issue here is, that the BeanCollection is lazily initialized with no modifyListenMode set after the first save.
This happens only for O2M relations with no order column. (Others work fine) See: https://github.com/ebean-orm/ebean-agent/blob/d4c40f1ce85c58f99cb0a85152aaa0a0075a9c01/ebean-agent/src/main/java/io/ebean/enhance/entity/FieldMeta.java#L469

And we need also a save, where the bean is saved with an empty/null value in the O2M property. 
Subsequent saves will not update the modifyListenMode. See SaveManyBeans
```java
      if (insertedParent) {
        // after insert set the modify listening mode for private owned etc
        c.setModifyListening(many.modifyListenMode());
      }
```

We found this in one of our unit-tests, where we've configured a bean for different states. It is probably something, that should not be too critical in real code, as you normally save a bean only once (When the bean was retrieved from DB, it should not occur)

### DIFF
--- a/ebean-test/src/test/java/org/tests/cascade/TestOrderedList.java
+++ b/ebean-test/src/test/java/org/tests/cascade/TestOrderedList.java
@@ -1,9 +1,14 @@
 package org.tests.cascade;
 
+import io.ebean.bean.BeanCollection;
 import io.ebean.xtest.BaseTestCase;
 import io.ebean.DB;
 import io.ebean.test.LoggedSql;
 import org.junit.jupiter.api.Test;
+import org.tests.model.basic.Order;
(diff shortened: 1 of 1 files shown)


rbygrave:
Fixed and replaced by PR - https://github.com/ebean-orm/ebean/pull/3793

PR #3657, opened 2025-07-23
Lazy load in BeanPersistController.preDelete does not work in batch mode

rPraml:
We found an issue, where lazyLoad in preDelete of does not work in batch mode.

**Example:**
```java
class BeanPersistController {
      public boolean preDelete(BeanPersistRequest<?> request) {
        assertThat(((EBasicVer) request.bean()).getDescription()).isEqualTo("MyDescription");
        return true;
      }
}


try (Transaction txn = db.beginTransaction()) {
      txn.setBatchMode(true);
      EBasicVer bean = db.find(EBasicVer.class).setUseCache(false).select("name").setId(id).findOne();
      db.delete(bean); // will remove bean from PC
      txn.commit(); // will invoke preDelete
}
```

**What happens here:**
- you need batch mode = true
- db.delete(bean) will delay the persist, but will call removeFromPersistenceContext immediately (came with https://github.com/ebean-orm/ebean/pull/2213/files#diff-681bf3a1feb3d0e7a61fd3697ebd41a0080e2efe80ecfd39294763cf1f0a7334R884)
- txn.commit will invoke the PersistController.preDelete
- lazy load occurs
- SqlTreeLoadBean.readIdBean will not find the bean in the PC and will insert the `localBean` instead. The lazyLoad happens on the wrong bean

**Workaround**
- do not rely on lazyload of `request.bean()` and fetch data explicit with a query in preDelete (`fromDb = DB.find(cls, request.bean().getId())`)

**Possible fix:**
- Do not really remove bean from persistence-context (This PR)
- Re-add bean temporary to PC (#3657)

### DIFF
--- a/ebean-test/src/test/java/io/ebean/xtest/event/BeanPersistControllerTest.java
+++ b/ebean-test/src/test/java/io/ebean/xtest/event/BeanPersistControllerTest.java
@@ -2,13 +2,15 @@ package io.ebean.xtest.event;
 
 
 import io.ebean.Database;
+import io.ebean.DatabaseBuilder;
 import io.ebean.DatabaseFactory;
 import io.ebean.Transaction;
-import io.ebean.DatabaseBuilder;
 import io.ebean.config.DatabaseConfig;
 import io.ebean.event.BeanDeleteIdRequest;
 import io.ebean.event.BeanPersistAdapter;
+import io.ebean.event.BeanPersistController;
 import io.ebean.event.BeanPersistRequest;
+import io.ebean.test.LoggedSql;
 import org.junit.jupiter.api.Test;
 import org.tests.model.basic.EBasicVer;
 import org.tests.model.basic.UTDetail;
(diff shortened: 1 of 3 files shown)


ebean-test/src/test/java/io/ebean/xtest/event/BeanPersistControllerTest.java
```diff
@@ -132,15 +134,124 @@ public void testInsertUpdateDelete_given_stopPersistingAdapter() {
...
+      txn.setBatchMode(true);
+      db.delete(master);
+      txn.commit();
+    }
+
+    master = db.find(UTMaster.class, id);
+    assertThat(master).isNotNull();
+    // CHECKME: Deleting of master was denied by the PersistListener
+    // What about detail? Is this intended, that it will be deleted?
+    assertThat(master.getDetails()).hasSize(0);
```
rPraml, line 177: CHECKME:
Can you tell me if the PersistController is intended to protect a bean from deletion? If so, then I would expect that if the deletion process was prevented by UtMaster, its details wouldn't be deleted.

But if you say the PersistController only takes care of the physical persistence, then that would be OK (although perhaps this should be explicitly mentioned again in the documentation for preInsert/Update/Delete).

rbygrave:
Closing this in favour of https://github.com/ebean-orm/ebean/pull/3658

PR #3658, opened 2025-07-23
Lazy load in BeanPersistController.preDelete does not work in batch mode (2)

rPraml:
See #3657 - this is an other approach to fix that bug


rbygrave:
This was effectively merged in via https://github.com/ebean-orm/ebean/pull/3748 

I believe this PR can be closed now @rPraml

rbygrave:
Closing as merged via https://github.com/ebean-orm/ebean/pull/3748

PR #3692, opened 2025-10-24
FEATURE: Add support for Generic mapped superclass

dsa-ferreira:
Extends DeployCreateProperties and ProcessingContext to allow generic types in MappedSuperclasses. This is intended to allow generic types to be used as column types in Entities and QEntities. The main use-case we are trying to achieve is to allow typification of Ids and Id references.

Example of what we want to achieve:

```
@MappedSuperclass
public abstract class DomainObject<T> {

    @Id
    private T id;

...

    public T getId() {
        return this.id;
    }

...

}

@Entity
public class Client extends DomainObject<ClientId> {

...

}

```


rbygrave:
There are no tests yet, nothing added to `ebean-querybean/src/test` yet.

dsa-ferreira:
added some tests, not sure if these are enough or not. please let me know!

PR #3705, opened 2025-11-22
Problem with filterMany

PromanSEW:
~Unfortunately I could not create a working test, but I provided minimal workaround with my problem~
Test is working.
See https://github.com/ebean-orm/ebean/pull/3626#issuecomment-3562488621
This code worked with Ebean 15.1.0, but crashes with `UnsupportedOperationException` on Ebean 17.1.0
My full working query with other fields (don't make sense anyway):
```java
ClanQuest quest = finder.query().setId(id).select("conditions, fishes, time, ended")
					.fetch("questtype", "id").fetch("clan", "buildings")
					.filterMany("clan.buildings").eq("type", Building.CAFE).findOne();
```

### DIFF
--- a/ebean-test/src/test/java/org/tests/model/basic/Building.java
+++ b/ebean-test/src/test/java/org/tests/model/basic/Building.java
@@ -13,6 +13,7 @@ public class Building extends Model {
   public int id;
   @Column(nullable = false)
   public String type;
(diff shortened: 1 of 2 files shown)


PromanSEW:
Why this PR is closed? I don't do that...

PR #3741, opened 2026-04-09
Alternative Fix for @Column on timestamp defined as timestamp(255)

rbygrave:
As per https://github.com/ebean-orm/ebean/discussions/3720

Fix:
Detect the default value of `@Column.length()` which will be 255 when the JPA API is used rather than the Ebean supplied one.


rbygrave:
Closing this PR as another approach was taken which was to make the timestamp type (that does have precision but not 255 worth of precision) more robust in terms of how it handles precision.

PR #3750, opened 2026-04-21
Change [implicit] Lazy loading to be by TYPE rather than by PATH

rbygrave:
We have some graph models where a common type (in the tests it is Label) is used in many different *paths* of the graph. When we are loading via *path* then all the loading of the Labels isn't in a single batch / load context but instead split into different load contexts PER PATH.

This change, means that lazy loading operates by TYPE instead of by PATH.

In the tests, all the Labels are read via a single lazy loading query rather that one lazy loading query per PATH, and this is more efficient.


rbygrave:
FYI @nedge  - lazy loading all the Labels together.  Lazy loading by **_TYPE_** rather than by PATH.

rbygrave:
This PR was included in https://github.com/ebean-orm/ebean/pull/3753

PR #3760, opened 2026-05-05
FEATURE: Support for Map<Enum, Object> in @DbJson(B) fields

dragkes:
Addresses issue #3735


rbygrave:
So I'm pondering if there is some design option to reduce the number of combinations ebean has internally. This change looks sound but I am thinking about doing a deeper review the design pattern ebean has here and if and when we can improve that.

rbygrave:
This fix is effectively also included in https://github.com/ebean-orm/ebean/pull/3789

rbygrave:
I'll close this PR, fixed in versions 18.0.0 and 19.0.0

PR #3762, opened 2026-05-05
Performance improvement in OrmQueryProperties with immutable hash pre…

rbygrave:
…fix and suffix

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/querydefn/OrmQueryProperties.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/querydefn/OrmQueryProperties.java
@@ -34,6 +34,8 @@ public final class OrmQueryProperties implements Serializable {
   private final String path;
   private final boolean allProperties;
   private final Set<String> included;
+  private final String immutableHashPrefix;
+  private final String immutableHashSuffix;
   private final FetchConfig fetchConfig;
   private final boolean cache;
   /**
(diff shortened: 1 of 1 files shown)


ebean-core/src/main/java/io/ebeaninternal/server/querydefn/OrmQueryProperties.java
```diff
@@ -141,6 +155,23 @@ private static Set<String> immutableIncluded(Set<String> included) {
...
+      builder.append(path);
+    }
+    if (included != null) {
+      builder.append("/i");
+      appendSet(builder, included);
+    }
+    return builder.toString();
+  }
+
+  private static String buildImmutableQueryPlanHashSuffix(FetchConfig fetchConfig) {
```
zhangshifeng, line 171: Should we put `}` in the suffix?

rbygrave: Yes, although an alternative is to just include the immutableHashSuffix into the immutableHashPrefix. Hmm.

rbygrave: Ah, scratch my thought ... the immutableHashSuffix needs to be separate, so yeah that ending `}` could go in there.

rbygrave: PR https://github.com/ebean-orm/ebean/pull/3768

PR #3786, opened 2026-06-12
Test for GetchGroup

nedge:
I'm trying to force the loading of OneToMany using FetchGroups in a query....setUnmodifiable(true)
My goal is to be able to load parts of the object graph and use the ImmutableBeanCache functionality but I get 

 ResourceEntityTest.a_find_fetchQuerySecondary_inheritsImmutableCache_expect_noLabelLazyLoadSql:662->lambda$a_find_fetchQuerySecondary_inheritsImmutableCache_expect_noLabelLazyLoadSql$19:663 » LazyInitialisation Property not loaded: name

### DIFF
--- a/ebean-spring-txn/src/test/java/org/example/EbeanSpringModuleTest.java
+++ b/ebean-spring-txn/src/test/java/org/example/EbeanSpringModuleTest.java
@@ -117,6 +117,24 @@ public class EbeanSpringModuleTest {
     logger.info("nonTransactional find user1:{} user2:{}", user, user2);
   }
 
+  public void testLoad() {
+	  User user = new User();
+	  user.setName("TestLoad");
+	  userService.save(user);
+	  
+	  logger.info("Saved new User");
+	  assertThat(user.getOid()).isGreaterThan(0);
+
+	  User found = DB.find(User.class, user.getOid());
+
+	  user.setName("TestLoad2");
+	  
+	  userService.save(user);
+	  
+	  found = DB.find(User.class, user.getOid());
+	  
+	  assertThat(user.getName()).equals(found.getName());
+  }
   /**
    * Return the user service.
    */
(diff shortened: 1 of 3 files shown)


rob-bygrave:
Maybe:

  Verdict: not an ebean bug — usage error. Reproduced the failure (LazyInitialisationException: Property not loaded: name). Root cause: the new test fetches   fetchQuery("attributeValues", "intValue") but then accesses AttributeValue.name (a newly added @ManyToOne Label) which is never fetched; combined with   setUnmodifiable(true) (read-only beans → lazy loading disabled by design), accessing the unfetched property correctly throws.

  Confirmed fix: changing to fetchQuery("attributeValues", "intValue, name") makes the test pass including assertThat(sql).isEmpty() — the Label graph resolves   entirely from the ImmutableBeanCache with zero lazy-load SQL, which is exactly the author's goal.

nedge:
I changed the code in the test case as follows (line 678)
` AttributeValueOwner owner = DB.find(AttributeValueOwner.class)
      .setId(resource.getAttributeValueOwner().id())
      .fetchLazy("attributeValues", "intValue, name")
      .fetch("attributeValues.attributeDescriptor", "name,description")
      .using(cache)
      .findOne();`

but still getting
` ResourceEntityTest.a_find_fetchQuerySecondary_inheritsImmutableCache_expect_noLabelLazyLoadSql:662->lambda$a_find_fetchQuerySecondary_inheritsImmutableCache_expect_noLabelLazyLoadSql$19:663 » LazyInitialisation Property not loaded: name`

rob-bygrave:
Agent says:

└ Re-reviewed PR #3786. Key new finding: the contributor (nedge) misapplied my suggested fix. They added , name to the find_fetchLazySecondary_… test (line 678 —   which wasn't failing and doesn't use setUnmodifiable), but the actually-failing test a_find_fetchQuerySecondary_… (~line 646) is unchanged and still has   fetchQuery("attributeValues", "intValue") with setUnmodifiable(true) — hence the same LazyInitialisationException: Property not loaded: name.

nedge:
I still have issues in Timerbee with the FetchGroup - One part may have to do with using inheritance so I will try to remove the inheritance first.

rbygrave:
I'll close this PR for now

PR #3829, opened 2026-07-03
ebean-redisson - initial commit (#3711)

### LINKED ISSUE #3711: Redisson L2 cache
Hey!
I've created a [plugin](https://github.com/dragkes/ebean-redisson/tree/master) that provides L2 caching support using Redis, similar to  [ebean-redis](https://github.com/ebean-orm/ebean/tree/master/ebean-redis), but built with Redisson instead of Jedis.
Would appreciate it if you could take a look.

dragkes:
#3711

### DIFF
--- a/ebean-redisson/pom.xml
+++ b/ebean-redisson/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <artifactId>ebean-parent</artifactId>
     <groupId>io.ebean</groupId>
-    <version>18.1.0</version>
+    <version>18.2.0</version>
   </parent>
 
   <artifactId>ebean-redisson</artifactId>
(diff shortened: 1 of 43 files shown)


rbygrave:
A quick AI review suggests 2 things to have a look at:

1. tenant-aware getAll() is broken in ebean-redisson/src/main/java/io/ebean/redisson/encode/CacheCodec.java:29-40. substring(pos + 1)  decodes 123:tenantA to tenantA, not the original cache key. That makes RedissonCache.getAll() miss every tenant-aware entry, and the DuelCache warm-up path (DuelCache.java:49-63) can’t populate near cache correctly.

2. RedissonCache.getAll() returns the wrong key type in ebean-redisson/src/main/java/io/ebean/redisson/RedissonCache.java:309-320. It 
builds a Map<String,Object> and stores stringified keys, so callers no longer get back the original ids. That breaks the ServerCache contract 
and makes the near-cache path store entries under strings instead of the real key objects, so later near.get(id) misses.

I am not sure if those comments are accurate, need to have a look.

dragkes:
Yeah, confirmed and fixed. Also added more tests (similar to `ebean-redis`). Thanks @rbygrave!

PR #3874, opened 2026-08-07
#3407 Honor isolation level on read-only transactions

### LINKED ISSUE #3407: ReadOnly transactions does not honor isolation-level
## Expected behavior
If a method is annotated with `@Transactional(readOnly = true, isolation = TxIsolation.READ_UNCOMMITTED)` the isolation level should be set

## Actual behavior
You get a readonly-transaction with READ_COMMITTED
See: https://github.com/ebean-orm/ebean/blob/master/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java#L275

## Use case
We have update locks on certain tables, but we do not want to block our user interface too much, so the idea is, to execute certain queries with `READ_UNCOMMITTED` - but I noticed, that it won't work when using the readonly connection

## Question
Is this intended or a bug

## Workaround
do not use readonly connection

arimu1:
## Summary

Fixes #3407 — when a method (or `TxScope`) is read-only **and** specifies an isolation level, Ebean dropped the isolation and always used the connection default (typically `READ_COMMITTED`).

### Root cause

`TransactionManager.createTransaction(TxScope)` used the read-only path without applying isolation:

```java
if (txScope.isReadonly()) {
  return createReadOnlyTransaction(null, false); // isolation ignored
} else {
  return createTransaction(true, txScope.getIsolationLevel());
}
```

### Fix

After creating the read-only transaction, apply the same `setIsolationLevel` helper used by normal transactions:

```java
if (txScope.isReadonly()) {
  SpiTransaction transaction = createReadOnlyTransaction(null, false);
  return transactionFactory.setIsolationLevel(transaction, txScope.getIsolationLevel());
}
```

When isolation is unset (`-1`), behavior is unchanged.

## Tests

- `TestTransactionalReadOnly#test_readonly_honors_isolation` — programmatic `TxScope.required().setReadOnly(true).setIsolation(SERIALIZABLE)`
- `TestTransactionalReadOnly#test_readonly_annotation_honors_isolation` — `@Transactional(readOnly = true, isolation = SERIALIZABLE)`
- Existing read-only metric tests still pass

```
mvn -pl ebean-test -Dtest=org.tests.transaction.TestTransactionalReadOnly test
# Tests run: 4, Failures: 0 (Temurin 21, H2)
```

## Notes

Isolation is still only as useful as the target database allows (e.g. some platforms map or reject `READ_UNCOMMITTED`). This change ensures the requested level is applied to the JDBC connection for read-only scopes when the platform supports it.

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/transaction/TransactionManager.java
@@ -285,7 +285,9 @@ public class TransactionManager implements SpiTransactionManager {
 
   private SpiTransaction createTransaction(TxScope txScope) {
     if (txScope.isReadonly()) {
-      return createReadOnlyTransaction(null, false);
+      // Honor isolation on read-only scopes (e.g. @Transactional(readOnly=true, isolation=...))
+      SpiTransaction transaction = createReadOnlyTransaction(null, false);
+      return transactionFactory.setIsolationLevel(transaction, txScope.getIsolationLevel());
     } else {
       return createTransaction(true, txScope.getIsolationLevel());
     }
--- a/ebean-test/src/test/java/org/tests/transaction/TestTransactionalReadOnly.java
+++ b/ebean-test/src/test/java/org/tests/transaction/TestTransactionalReadOnly.java
@@ -1,12 +1,17 @@
 package org.tests.transaction;
 
-import io.ebean.xtest.BaseTestCase;
 import io.ebean.DB;
+import io.ebean.Transaction;
+import io.ebean.TxScope;
 import io.ebean.annotation.Transactional;
+import io.ebean.annotation.TxIsolation;
 import io.ebean.meta.MetaTimedMetric;
+import io.ebean.xtest.BaseTestCase;
 import org.junit.jupiter.api.Test;
 import org.tests.model.basic.Customer;
 
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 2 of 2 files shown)


rbygrave:
Do you have an application hitting this issue or is it more a bug fix for a known issue?

If you have an application hitting this, what database and isolation level is being used? Can you explain the use case? Do you have a workaround?

arimu1:
@rbygrave Honest answer: I don't have a production app hitting this myself — the fix comes from the known issue [#3407](https://github.com/ebean-orm/ebean/issues/3407) reported by @rPraml.

**Reporter's use case (from the issue):** they hold update locks on certain tables and wanted some UI queries on a **read-only** connection at `READ_UNCOMMITTED` so the UI is less blocked. With `@Transactional(readOnly = true, isolation = READ_UNCOMMITTED)` they still got a read-only txn at the default isolation (READ_COMMITTED) because `createReadOnlyTransaction` never applied the requested isolation.

**Workaround they used:** skip the read-only connection (use a normal transaction) when a non-default isolation is required.

**This PR:** when `TxScope` requests both read-only and an isolation level, apply `setIsolationLevel` after creating the read-only transaction (same path as non-read-only scopes). Covered by `TestTransactionalReadOnly` for `TxScope` + `@Transactional` SERIALIZABLE on H2.

If you'd rather keep isolation unsupported on the read-only pool for product reasons (e.g. only useful on DB2), happy to close or narrow — just say the word.

rbygrave (approved):

rbygrave:
Hi @rPraml  - I'm good with this PR and looking to merge. Note sure if you've had a look at it yet but if you can that would be great. For me this change is exactly what I expect.

Thanks, Rob.

rPraml:
LGTM

PR #3877, opened 2026-08-13
Bugfix/3876

### LINKED ISSUE #3876: Batch statements issued out of order when a BeanPersistController persists from a callback (follow-up to #3148)
Hi everyone, we notice a bug when persisting an entity (DcoParent) containing a list of entities (DcoAsset) linked throuw a table of association which is Java Managed (DcoLink).
When we have an adapter doing insert into database, the batch is flushed but not respecting the correct entity order which result in DB rejecting due to FK constraints.

### Version

18.4.0 : reproduced both against the published jar and against a build of `master`. Platform: H2 (the test below runs in `ebean-test`).

### What happens

#3148 fixed the case where a **query** run from a `BeanPersistController` callback flushes the batch that is currently being executed: `BatchControl.executeNow(list)` turns `flushOnQuery` off for the duration.

A **persist** run from the same callback is not covered. It reaches `BatchControl.executeOrQueue`, which flushes, so the statements queued behind the one currently executing are issued early.

Saving a parent/child graph in batch while a row is written from `preInsert`:

```
insert into dco_link (parent_id, asset_id) values (?,?)
  NULL not allowed for column "PARENT_ID"
```

The children are issued before their parent has an id, the same symptom #3148 describes: *"the later three will fail, because customer 2 is not yet saved and has no ID"*.

Persisting from a callback is documented behaviour, the `BeanPersistController` javadoc gives it as an example:

```java
Object extaBeanToSave = ...;
Transaction t = request.getTransaction();
Database server = request.getEbeanServer();
database.save(extraBeanToSave, t);
```

### Reproducer

Three entities, an adapter and a test.

```java
@Entity
public class DcoParent {
  @Id @GeneratedValue Long id;
  String name;
  @OneToMany(cascade = CascadeType.ALL, mappedBy = "parent", orphanRemoval = true)
  List<DcoLink> links = new ArrayList<>();
}

@Entity
public class DcoLink {
  @Id @GeneratedValue Long id;
  @ManyToOne(optional = false) DcoParent parent;
  @OneToOne(optional = false, cascade = CascadeType.ALL, orphanRemoval = true) DcoAsset asset;
}

@Entity
public class DcoAsset {
  @Id @GeneratedValue Long id;
  @Version Long version;
  String name;
}

@Entity
public class DcoAudit {   // the row written from the callback
  @Id @GeneratedValue Long id;
  String message;
}
```

```java
public class DcoParentAdapter extends BeanPersistAdapter {

  @Override
  public boolean isRegisterFor(Class<?> cls) {
    return DcoParent.class.equals(cls);
  }

  @Override
  public boolean preInsert(BeanPersistRequest<?> request) {
    DcoParent parent = (DcoParent) request.bean();
    request.database().save(new DcoAudit("inserting " + parent.getName()), request.transaction());
    return true;
  }
}
```

```java
@Test
void insertParentBeforeItsLinks_whenCallbackWritesDuringFlush() {
  List<DcoParent> parents = new ArrayList<>();
  for (int i = 0; i < 3; i++) {
    DcoParent parent = new DcoParent("batch-parent-" + i);
    parent.addAsset(new DcoAsset("batch-asset-" + i));   // creates the DcoLink
    parents.add(parent);
  }

    txn.setBatchMode(true);
    txn.setBatchSize(50);
    DB.saveAll(parents);
    txn.commit();     // DataIntegrityException here
  }
}
```

### Why

- `PersistRequestBean.executeInsert()` and `executeUpdate()` call `controller.preInsert` / `preUpdate` from inside `executeNow()`, that is, during the batch execution.
- The persist done there reaches `BatchControl.executeOrQueue`, whose first branch calls `flush()`.
- `flushInternal` runs `executeAll()` re-entrantly. The bean holder being executed has already had its list taken out of it, so the remaining holders are executed first : ahead of the statements that were queued before them.

The same thing is visible on the delete side with `logSummary` on:

```
BatchControl flush [DcoLink:0 d:2, DcoAsset:1 d:2]                  <- outer flush
BatchControl flush [DcoLink:0 d:0, DcoAsset:1 d:2, DcoAudit:2 i:1]  <- from the callback
```

### Suggested fix

Guard `executeOrQueue` the way `executeNow` already guards `flushOnQuery`: while the batch is executing, queue instead of flushing. Statements added in the meantime are picked up by the `do/while` loop in `executeAll()`.

```java
  private boolean executing;

  void executeNow(ArrayList<PersistRequest> list) throws BatchedSqlException {
    boolean old = transaction.isFlushOnQuery();
    transaction.setFlushOnQuery(false);
    boolean oldExecuting = executing;
    executing = true;
    try {
      ...
    } finally {
      executing = oldExecuting;
      transaction.setFlushOnQuery(old);
    }
  }

  public int executeOrQueue(PersistRequestBean<?> request, boolean batch) throws BatchedSqlException {
    if (!executing && (!batch || (batchFlushOnMixed && !pstmtHolder.isEmpty()))) {
      flush();
    }
    ...
  }
```

With that change the test above passes, `TestBatchInsertFlush` (the test added by #3148) still passes, and the full `ebean-test` suite runs 2561 tests with no new failure. The only failing test, `TestNatKeyCacheWithForeignKey#test_findOne`, fails the same way without the patch and passes when run on its own, so it looks order-dependent and unrelated.

Will open a PR with the fix proposal and the test.

AntoineDuComptoirDesPharmacies:
Aim to fix the following issue : https://github.com/ebean-orm/ebean/issues/3876

Would like to know if this is the good direction or if there is some downside of disabling flush at this step of the execution.

Thanks !

### DIFF
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
@@ -81,6 +81,13 @@ public final class BatchControl {
    */
   private int bufferMax;
 
+  /**
+   * True while the batched requests are being executed. A persist performed from a
+   * BeanPersistController callback must not flush the batch that is executing it, the same way
+   * executeNow() stops a query from doing so.
+   */
+  private boolean executing;
+
   private final Queue[] queues = new Queue[2];
 
   static final int DELETE_QUEUE = 0;
@@ -163,8 +170,9 @@ public final class BatchControl {
    * according to the depth (object graph depth).
    */
   public int executeOrQueue(PersistRequestBean<?> request, boolean batch) throws BatchedSqlException {
-    if (!batch || (batchFlushOnMixed && !pstmtHolder.isEmpty())) {
-      // flush when mixing beans and updateSql
+    if (!executing && (!batch || (batchFlushOnMixed && !pstmtHolder.isEmpty()))) {
+      // flush when mixing beans and updateSql, unless we are inside the execution of the batch
+      // itself : flushing then would issue the statements queued behind the current one early
       flush();
     }
     if (!batch) {
@@ -226,7 +234,9 @@ public final class BatchControl {
   void executeNow(ArrayList<PersistRequest> list) throws BatchedSqlException {
     boolean old = transaction.isFlushOnQuery();
     transaction.setFlushOnQuery(false);
-    // disable flush on query due transaction callbacks
+    boolean oldExecuting = executing;
+    executing = true;
+    // disable flush on query and on persist due transaction callbacks
     try {
       for (int i = 0; i < list.size(); i++) {
         if (i % batchSize == 0) {
@@ -237,6 +247,7 @@ public final class BatchControl {
       }
       flushPstmtHolder();
     } finally {
+      executing = oldExecuting;
       transaction.setFlushOnQuery(old);
     }
   }
--- a/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
+++ b/ebean-core/src/main/java/io/ebeaninternal/server/persist/BatchControl.java
@@ -146,8 +146,9 @@ public final class BatchControl {
    * to the depth.
    */
   public int executeStatementOrBatch(PersistRequest request, boolean batch, boolean addBatch) throws BatchedSqlException {
-    if (!batch || (batchFlushOnMixed && !isBeansEmpty())) {
-      // flush when mixing beans and updateSql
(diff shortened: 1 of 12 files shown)


rbygrave:
Looks good to me. I am happy with it.

Only pondering if `executeStatementOrBatch()` has a similar need or not?

AntoineDuComptoirDesPharmacies:
Good catch ! Yes indeed it will produce the same problem.

I added a test to cover the use case and corrected the code. 👍

rbygrave (approved):

